PR #2, opened 2022-08-10
fix: SpecReader yaml handling

disq:


specs/source.go
```diff
@@ -21,13 +21,29 @@ type SourceSpec struct {
 }
 
 func (s *SourceSpec) UnmarshalYAML(n *yaml.Node) error {
-	if err := n.Decode(s); err != nil {
+	var sourceSpecWithoutUnmarshal struct {
```
disq, line 24: This can be some other way, such as adding an inner struct inside SourceSpec etc. but handling it in other places gets tedious, and we shouldn't leak implementation details into public APIs.

yevgenypats: Why we need to duplicate the spec definition here?

disq: because `yaml.Decode` calls `UnmarshalYAML` (if there is one) which then calls `Decode` and so on.

disq: Fixed by adding a new method, `Validate() error` which is called after unmarshal.

PR #12, opened 2022-08-24
feat: Support timestamp columns

hermanschaaf:
Add support for timestamp columns to v2 branch (previously they were converted to JSON)

### DIFF
--- a/codegen/golang.go
+++ b/codegen/golang.go
@@ -3,18 +3,18 @@ package codegen
 import (
 	"embed"
 	"fmt"
+	"github.com/cloudquery/plugin-sdk/schema"
+	"github.com/iancoleman/strcase"
 	"io"
 	"reflect"
 	"text/template"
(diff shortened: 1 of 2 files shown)


codegen/golang.go
```diff
@@ -26,10 +26,17 @@ func valueToSchemaType(v reflect.Type) (schema.ValueType, error) {
 		return schema.TypeInt, nil
 	case reflect.Float32, reflect.Float64:
 		return schema.TypeFloat, nil
-	case reflect.Struct, reflect.Map:
+	case reflect.Map:
+		return schema.TypeJSON, nil
+	case reflect.Struct:
+		switch f.(type) {
```
disq, line 32: Can't you check for `time.Time` from `v.PkgPath()` or `v.Name()`?

yevgenypats (approved):

PR #13, opened 2022-08-24
Add column meta struct

hermanschaaf:
I'm not sure whether this is necessarily the cleanest way to handle it, but having the original name of a field allows us to do the following inside the generate code:

```
	for u, c := range r.Table.Columns {
		if c.Meta.OriginalName != strcase.ToCamel(c.Name) {
			r.Table.Columns[u].Resolver = fmt.Sprintf(`schema.PathResolver(%q)`, c.Meta.OriginalName)
		}
	}
```

This then handles cases like `ID`, `DocURL`, etc automatically.

### DIFF
--- a/codegen/golang.go
+++ b/codegen/golang.go
@@ -109,6 +109,9 @@ func NewTableFromStruct(name string, obj interface{}, opts ...TableOptions) (*Ta
 		column := ColumnDefinition{
 			Name: t.nameTransformer(field.Name),
 			Type: columnType,
+			Meta: ColumnMeta{
+				OriginalName: field.Name,
+			},
 		}
 		t.Columns = append(t.Columns, column)
 	}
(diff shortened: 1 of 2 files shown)


erezrokah (approved):

codegen/golang.go
```diff
@@ -109,6 +109,9 @@ func NewTableFromStruct(name string, obj interface{}, opts ...TableOptions) (*Ta
 		column := ColumnDefinition{
 			Name: t.nameTransformer(field.Name),
 			Type: columnType,
+			Meta: ColumnMeta{
+				OriginalName: field.Name,
```
yevgenypats, line 113: Nit: Let's just put it as a field `OriginalName` ? this will eliminate another meta  `meta` definition ?

yevgenypats (changes requested):
One nit

hermanschaaf:
Closing in favor of https://github.com/cloudquery/plugin-sdk/pull/14

PR #16, opened 2022-08-29
feat: Speed up description parsing, support use of comments and other…

hermanschaaf:
- Cache package parsing for faster description-reading for ~30x speed-up in the Heroku case. Currently uses a global cache, which I figure is okay for a codegen binary that's not supposed to be long-running. If we move `NewTableFromStruct` to be a struct method, we can move the cache onto the same struct. Let me know how you feel about this
- if docstring is unavailable, use comment string
- add DescriptionTransformer option to optionally clean descriptions

Before:

```
✗ time go run main.go
go run main.go  40.00s user 27.64s system 417% cpu 16.212 total
```

After:

```
✗ time go run main.go
go run main.go  1.11s user 0.79s system 112% cpu 1.699 total
```

### DIFF
--- a/codegen/golang.go
+++ b/codegen/golang.go
@@ -17,11 +17,7 @@ import (
 //go:embed templates/*.go.tpl
 var TemplatesFS embed.FS
 
-var pkgCache map[string][]*packages.Package
-
-func init() {
-	pkgCache = map[string][]*packages.Package{}
-}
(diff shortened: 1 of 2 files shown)


hermanschaaf:
Closing this for now, as we might not support description parsing going forward.

PR #29, opened 2022-08-31
feat: Add Name() to source and destination clients

hermanschaaf:
Currently it uses the example config GRPC call to get the name (previously not exposed via the client). In the future we could change it for a `Name` GRPC call, if we wanted.


clients/destination.go
```diff
@@ -30,6 +30,14 @@ func NewLocalDestinationClient(p plugins.DestinationPlugin) *DestinationClient {
 	}
 }
 
+func (c *DestinationClient) Name(ctx context.Context) (string, error) {
+	res, err := c.pbClient.GetExampleConfig(ctx, &pb.GetExampleConfig_Request{})
```
yevgenypats, line 34: Should we create a protobuf call just for that?

hermanschaaf: Ok, will update 👍

hermanschaaf: Done; the RPC calls have been split into three separate ones now (name, version, config)

yevgenypats (approved):

PR #47, opened 2022-09-02
feat: Generate source plugin docs

hermanschaaf:
Adds the ability to generate markdown docs for source plugins. Mostly migrated from `cq-provider-sdk` with minor tweaks and new tests.


docs/source.go
```diff
@@ -0,0 +1,106 @@
...
+	"text/template"
+)
+
+const (
+	tablesDir = "tables"
+)
+
+// GenerateSourcePluginDocs creates table documentation for the source plugin based on its list of tables
+func GenerateSourcePluginDocs(p *plugins.SourcePlugin, outputPath string, deleteOld bool) error {
+	if deleteOld {
```
yevgenypats, line 24: I think we should never delete files and it should be part of someone script if he wants to delete everything and regenerated

docs/source.go
```diff
@@ -0,0 +1,106 @@
...
+		},
+	})
+	t, err := t.New("").Parse(tableTmpl)
+	if err != nil {
+		return err
+	}
+
+	var buf bytes.Buffer
+
+	if err := t.Execute(&buf, table); err != nil {
```
yevgenypats, line 87: Just a nit: but I think we can make it more Goish/efficient and make a writer from a file and this way execute will write to a file in chunks so it wont have it all in memory and then output to disk

yevgenypats (changes requested):
Looks good. few minor comments.

docs/source.go
hermanschaaf: @yevgenypats I agree, but this will lead to a lot of duplication since all the plugins will need this functionality, and we don't want them to share common libraries in the monorepo. 

How about we remove the option here, but we add a convenience function for deleting directories to the SDK, maybe in another package?

docs/source.go
hermanschaaf: I added a helper function in this commit: https://github.com/cloudquery/plugin-sdk/pull/47/commits/3ac29627e2bd1b6b28e04fcf0c7ba06ea77fb379

docs/source.go
hermanschaaf: I added a helper function in this commit: https://github.com/cloudquery/plugin-sdk/commit/3ac29627e2bd1b6b28e04fcf0c7ba06ea77fb379

docs/source.go
hermanschaaf: Good suggestion, thanks 👍

docs/source.go
```diff
@@ -0,0 +1,71 @@
...
+			return strings.ReplaceAll(text, "\n", " ")
+		},
+	})
+	t, err := t.New("").Parse(tableTmpl)
+	if err != nil {
+		return err
+	}
+
+	outputPath := filepath.Join(dir, fmt.Sprintf("%s.md", table.Name))
+	f, err := os.Create(outputPath)
```
yevgenypats, line 62: I think this should be truncated as a default behaviour.

docs/source.go
yevgenypats: I actually meant for us to something like `rm -rf *.md` && `go run main.go doc` so a helper function wont be needed because it highly depends on what is in the directory and specifaclly in our case what is in our website so it can also be something like `rm -rf cloudquery*.md` . I think in cobra there is also no logic around deleting - https://github.com/spf13/cobra/blob/main/doc/md_docs.go

hermanschaaf: Ah I see! Okay yeah, makes sense. I'll remove the delete functionality from this PR and update the scripts to run something like `rm -rf *.md` in a follow-up where we migrate to this implementation.

docs/source.go
hermanschaaf: No change needed here I believe, `os.Create` creates or truncates the named file. If the file already exists, it is truncated.

yevgenypats (approved):

PR #54, opened 2022-09-05
feat: Add sentry for serve.Serve function

yevgenypats:
plugin developers will be able to direct their plugin errors
to their own projects on sentry as we dont want to monitor any
community errors anyway.

Also each plugin should go into it's own project on sentry
because we are using server side grouping which are mostly
project based.

Tested it on gcp - grouping works smoothly with nice title (no more diags.Error) for every single error as we use standard `sentry.Message` as well as using sentry mechanism to group and change title based on tags.

Every error in the ETL process which is not ignore is sent with the error as string + `table_name` as tag.

https://sentry.io/organizations/cloudquery-v2/issues/?project=6720365

hopefully this brings sane error monitoring one step closer :) 🚀 

---

- [ ] Update or add tests 🧪

### DIFF
--- a/schema/table.go
+++ b/schema/table.go
@@ -3,7 +3,6 @@ package schema
 import (
 	"context"
 	"fmt"
-
 	"runtime/debug"
 	"sync"
 	"time"
--- a/schema/table.go
+++ b/schema/table.go
@@ -3,11 +3,13 @@ package schema
 import (
 	"context"
 	"fmt"
+
 	"runtime/debug"
 	"sync"
 	"time"
 
 	"github.com/cloudquery/plugin-sdk/helpers"
+	"github.com/getsentry/sentry-go"
 	"github.com/iancoleman/strcase"
 	"github.com/thoas/go-funk"
 )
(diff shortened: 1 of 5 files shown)


hermanschaaf (approved):

schema/table.go
```diff
@@ -211,16 +217,16 @@ func (t Table) Resolve(ctx context.Context, meta ClientMeta, syncTime time.Time,
...
-			if pks[resource.PrimaryKeyValue()] {
-				meta.Logger().Error().Str("table_name", t.Name).Str("primary_key", resource.PrimaryKeyValue()).Msg("duplicate primary key found")
-			} else {
-				pks[resource.PrimaryKeyValue()] = true
-			}
+			// if pks[resource.PrimaryKeyValue()] {
+			// 	meta.Logger().Error().Str("table_name", t.Name).Str("primary_key", resource.PrimaryKeyValue()).Msg("duplicate primary key found")
+			// } else {
+			// 	pks[resource.PrimaryKeyValue()] = true
+			// }
```
erezrokah, line 229:
```suggestion
			// if pks[resource.PrimaryKeyValue()] {
			// 	meta.Logger().Error().Str("table_name", t.Name).Str("primary_key", resource.PrimaryKeyValue()).Msg("duplicate primary key found")
			// } else {
			// 	pks[resource.PrimaryKeyValue()] = true
			// }
```

schema/table.go
```diff
@@ -190,7 +196,7 @@ func (t Table) Resolve(ctx context.Context, meta ClientMeta, syncTime time.Time,
 	totalResources := 0
 	// we want to check for data integrity
 	// in the future we can do that as an optinoal feature via a flag
-	pks := map[string]bool{}
+	// pks := map[string]bool{}
```
erezrokah, line 199:
```suggestion
```

serve/serve.go
```diff
@@ -99,6 +126,8 @@ func newCmdServe(opts Options) *cobra.Command {
 	cmd.Flags().StringVar(&network, "network", "tcp", `the network must be "tcp", "tcp4", "tcp6", "unix" or "unixpacket"`)
 	cmd.Flags().Var(logLevel, "log-level", fmt.Sprintf("log level. one of: %s", strings.Join(logLevel.Allowed, ",")))
 	cmd.Flags().Var(logFormat, "log-format", fmt.Sprintf("log format. one of: %s", strings.Join(logFormat.Allowed, ",")))
+	cmd.Flags().BoolVar(&noSentry, "no-sentry", false, "disable sentry")
```
erezrokah, line 129: Maybe use the same wording as the CLI e.g. `disable-telemetry`?

yevgenypats: I think for flags it's better to use `no-` and I used it everywhere else (I don't think we have `enable` or at least shouldn't ) - this should make flags short. I can change the help to `no sentry` if this is more consistent.

erezrokah: Ah sure I don't mind `no-` vs `disable-` just not use an internal implementation detail in the name of the flag

PR #64, opened 2022-09-07
chore: Golangci: remove deprecated plugins (and decorder)

hermanschaaf:
Some plugins are no longer supported and can be removed. I am also proposing we remove decorder: until it has the ability to automatically fix the issues it reports, we are wasting our time for (imo) little gain.

As for the deprecated plugins, here is the message from latest golanci-lint:

```
WARN [runner] The linter 'structcheck' is deprecated (since v1.49.0) due to: The owner seems to have abandoned the linter.  Replaced by unused.
WARN [runner] The linter 'deadcode' is deprecated (since v1.49.0) due to: The owner seems to have abandoned the linter.  Replaced by unused.
WARN [runner] The linter 'varcheck' is deprecated (since v1.49.0) due to: The owner seems to have abandoned the linter.  Replaced by unused.
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -68,8 +68,6 @@ linters:
   enable:
     - asciicheck
     - bodyclose
-    - deadcode
-    - decorder
     - depguard
     - dupl
     - errcheck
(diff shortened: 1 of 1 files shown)


erezrokah (approved):

erezrokah:
Nice 🚀  For `decorder` if we don't enforce it via a linter I suggest we don't enforce it in code reviews either (that was the reason I added it)

disq (approved):

PR #74, opened 2022-09-07
chore: Upgrade to Go 1.19

hermanschaaf:
https://tip.golang.org/doc/go1.19


erezrokah:
I think we should have released this as a breaking change. This is already a part of https://github.com/cloudquery/plugin-sdk/releases/tag/v0.2.5 anyways now :)

hermanschaaf:
Oh, yeah 🤦 I was thinking the user-facing behavior and interface shouldn't change, but you're right, probably should have been considered breaking.

PR #75, opened 2022-09-07
feat: Add option to unwrap structs 1 level down

erezrokah:
#### Summary

Adds an option to unwrap fields from embedded structs (1 level down only).
This is common in Azure to have a top level type with another inner struct for the properties.
Currently the properties will get generated as a single JSON column which doesn't make much sense.
Added an option so this is not a breaking change.

See example commit with this change:
https://github.com/cloudquery/cloudquery/pull/1754/commits/3388162b00f99dcc98d613356a589a847d3a520b

In the old code gen/manually written versions we use several inconsistent approaches:
1. Get the inner struct as a relation https://github.com/cloudquery/cloudquery/blob/fb8cb86c624719cb096483b52012c3cc51f64684/plugins/source/azure/resources/services/authorization/role_definitions.go#L117
2. Unwrap all the way down https://github.com/cloudquery/cloudquery/blob/fb8cb86c624719cb096483b52012c3cc51f64684/plugins/source/azure/resources/services/cosmosdb/sql_databases.go#L61 via built in path resolver
3. Resolve it manually https://github.com/cloudquery/cloudquery/blob/fb8cb86c624719cb096483b52012c3cc51f64684/plugins/source/azure/resources/services/monitor/activity_log_alerts.go#L154

Looking at the [changes](https://github.com/cloudquery/cloudquery/pull/1754/commits/3388162b00f99dcc98d613356a589a847d3a520b) generated by this feature unwrapping 1 level down seems like a decent approach, at least for Azure's case

Also looking at the resources we have using the new Azure SDK, it seems it's not needed as they don't have a "Properties" struct embedded into them https://github.com/erezrokah/cloudquery/blob/3388162b00f99dcc98d613356a589a847d3a520b/plugins/source/azure/resources/services/subscriptions/subscriptions.go#L18 

Finally even with this change we have some misses due to inconsistencies with the Azure SDK, for example:
https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/services/authorization/mgmt/2015-07-01/authorization#RoleAssignment has `Properties` as a regular struct, while https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/services/authorization/mgmt/2015-07-01/authorization#RoleDefinition has it as an embedded struct
image
image

---

- [ ] Update or add tests 🧪

### DIFF
--- a/codegen/golang.go
+++ b/codegen/golang.go
@@ -14,6 +14,11 @@ import (
 
 type TableOptions func(*TableDefinition)
 
+type structFieldWithPath struct {
+	*reflect.StructField
+	path string
+}
+
 //go:embed templates/*.go.tpl
 var TemplatesFS embed.FS
 
@@ -72,6 +77,20 @@ func WithExtraColumns(columns []ColumnDefinition) TableOptions {
 	}
 }
 
+// Unwrap specific struct fields (1 level deep only)
+func WithUnwrapFieldsStructs(fields []string) TableOptions {
+	return func(t *TableDefinition) {
+		t.structFieldsToUnwrap = fields
+	}
+}
+
+// Unwrap all fields that are embedded structs (1 level deep only)
+func WithUnwrapAllEmbeddedStructs() TableOptions {
+	return func(t *TableDefinition) {
+		t.unwrapAllEmbeddedStructFields = true
+	}
+}
+
 func defaultTransformer(name string) string {
 	return strcase.ToSnake(name)
 }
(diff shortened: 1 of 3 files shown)


yevgenypats (changes requested):
I like that option but I would implement it in such a way that when you specify `withEmbeddedStruct` it will get `column string` so it will unnest only specific field and it will be up to the user which one.

erezrokah:
> I like that option but I would implement it in such a way that when you specify `withEmbeddedStruct` it will get `column string` so it will unnest only specific field and it will be up to the user which one.

Ah cool that should solve the inconsistency I mentioned but with more manual configuration on the codegen side.
How about a pointer to a column string and if it's `nil` unwrap all? Then we can do both?

yevgenypats:
> > I like that option but I would implement it in such a way that when you specify `withEmbeddedStruct` it will get `column string` so it will unnest only specific field and it will be up to the user which one.
> 
> Ah cool that should solve the inconsistency I mentioned but with more manual configuration on the codegen side. How about a pointer to a column string and if it's `nil` unwrap all? Then we can do both?

Not sure I got it? maybe an example of we can do a short sync?

codegen/golang.go
```diff
@@ -87,6 +87,20 @@ func WithDescriptionsEnabled() TableOptions {
 	}
 }
 
+// Unwrap specific struct fields (1 level deep only)
+func WithUnwrapFieldsStructs(fields []string) TableOptions {
```
erezrokah, line 91: Handles the use case of:
```
type testStructWithNonEmbeddedStruct struct {
	Properties  *Properties
}
```

codegen/golang.go
```diff
@@ -87,6 +87,20 @@ func WithDescriptionsEnabled() TableOptions {
...
+	return func(t *TableDefinition) {
+		t.structFieldsToUnwrap = fields
+	}
+}
+
+// Unwrap all fields that are embedded structs (1 level deep only)
+func WithUnwrapAllEmbeddedStructs() TableOptions {
```
erezrokah, line 98: Handles the use case of:
```
type testStructWithNonEmbeddedStruct struct {
  *Properties
}
```

Without the need to configure each struct we'd like to unwrap

erezrokah:
@yevgenypats as discussed via Zoom I separated the cases of unwrapping all embedded structs and unwrapping specific struct fields. See https://github.com/cloudquery/cloudquery/pull/1754/commits/38338c52fc47d5c3a978048c7dffcb78a1cdb632 for example usage

erezrokah:
@yevgenypats added another commit to fix an issue with non embedded fields. When we unwrap those we need to pre-pend the parent field name to the path resolver https://github.com/cloudquery/plugin-sdk/pull/75/commits/cf964302bf657f6b3c85331b13395c8923c17e79

yevgenypats (changes requested):
Can we deprecate the overridefield here? it will make it easier for me to review this one.

erezrokah:
> Can we deprecate the overridefield here? it will make it easier for me to review this one.

Done in https://github.com/cloudquery/plugin-sdk/pull/86 and I'll rebase this branch after https://github.com/cloudquery/plugin-sdk/pull/86 is merged

erezrokah:
Rebased the PR and it should be good for another review

codegen/golang.go
```diff
@@ -120,31 +195,15 @@ func NewTableFromStruct(name string, obj interface{}, opts ...TableOptions) (*Ta
 
 	for i := 0; i < e.NumField(); i++ {
 		field := e.Type().Field(i)
-		if len(field.Name) == 0 {
-			continue
-		}
-		if unicode.IsLower(rune(field.Name[0])) {
-			continue
-		}
-		if sliceContains(t.skipFields, field.Name) {
```
yevgenypats, line 129: @erezrokah why move all those three checks into the `isUnwrap` ? as those are not really related?

erezrokah: I'm not sure I understand, those are extracted to `ignoreField` which is called by `addColumnFromField`.
`shouldUnwrap` is `return isFieldStruct(field.Type) && (t.unwrapAllEmbeddedStructFields && field.Anonymous || sliceContains(t.structFieldsToUnwrap, field.Name))`

erezrokah:
@yevgenypats rebased the PR with latest from `main`

yevgenypats:
Add a few commits on top - https://github.com/cloudquery/plugin-sdk/pull/111 and commented on slack

erezrokah:
Closing in favor of https://github.com/cloudquery/plugin-sdk/pull/111

PR #78, opened 2022-09-11
fix: Add missing comma when generating relations

erezrokah:
#### Summary

The `TableDefinition.Relations` accepts a string array but it's a bit confusing to use when you have multiple relations as it doesn't add a comma between relations.
As a result if someone has multiple relations they need to work around the issue by doing either:
- `relations:          []string{"siteAuthSettings(),publishingProfiles(),vnetConnections(),publishingProfiles()"}` (a single string) ([code](https://github.com/erezrokah/cloudquery/blob/1bac8634602c4db2e2cb5d8206f8a92aea6052c4/plugins/source/azure/codegen/recipes/web.go#L29))
or 
- `relations:          []string{"siteAuthSettings(),",  "publishingProfiles(),"}`

This PR fixes the issue (this makes the code consistent with [how we template columns](https://github.com/cloudquery/plugin-sdk/blob/e40466ba450499540a02737ace6655ea184fe42e/codegen/templates/column.go.tpl#L17)).

Another solution is to make relations a string type (instead of an array) to make it clear that it's printed as is.

For clarity the [first commit](https://github.com/cloudquery/plugin-sdk/commit/3bcc2a39d892b110c9d95ef9c92a55277c8c3afd) of this PR adds a test to generate a table template and the second has the fix with the change in the snapshot https://github.com/cloudquery/plugin-sdk/commit/c9d9db296856776f476ae9d4c0a2688a9fcded99 

---

- [x] Update or add tests 🧪

### DIFF
--- a/codegen/.snapshots/TestGenerateTemplate-should_add_comma_between_relations
+++ b/codegen/.snapshots/TestGenerateTemplate-should_add_comma_between_relations
@@ -1,0 +1,11 @@
+{
+		Name:         "with relations",
+		Columns: []schema.Column{
+
+		},
+
+		Relations: []*schema.Table{
+relation1relation2,
+		},
+
+}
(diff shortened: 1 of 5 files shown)


yevgenypats (approved):
I like the new snapshot idea!

PR #85, opened 2022-09-11
fix(testing): Validate all tables and relations

erezrokah:
#### Summary

I was expecting my Azure mock tests that have relations to fail as I don't handle mocking the related APIs.
I was surprised to see them pass 🙃 

The reason is that we need to validate that we got the related resources too.

---

- [x] Update or add tests 🧪

### DIFF
--- a/plugins/source_testing.go
+++ b/plugins/source_testing.go
@@ -58,11 +58,11 @@ func getTableResource(t *testing.T, table *schema.Table, resources []*schema.Res
 func validateTable(t *testing.T, table *schema.Table, resources []*schema.Resource) {
 	t.Helper()
 	resource := getTableResource(t, table, resources)
-	if resource != nil {
-		validateResource(t, resource)
+	if resource == nil {
+		t.Errorf("Expected table %s to be synced but it was not found", table.Name)
 		return
 	}
-	t.Errorf("Expected table %s to be synced but it was not found", table.Name)
+	validateResource(t, resource)
 }
 
 func validateTables(t *testing.T, tables schema.Tables, resources []*schema.Resource) {
--- a/plugins/source_testing.go
+++ b/plugins/source_testing.go
@@ -7,6 +7,7 @@ import (
 	"github.com/cloudquery/faker/v3"
 	"github.com/cloudquery/plugin-sdk/schema"
 	"github.com/cloudquery/plugin-sdk/specs"
+	"github.com/stretchr/testify/require"
 )
 
 type ResourceTestCase struct {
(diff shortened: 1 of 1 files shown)


yevgenypats (changes requested):
Good catch! Can we change it to instead of exact number of resources to just at least one resource in every table. as otherwise you can have more resources in one but no resources in others.

erezrokah:
> Good catch! Can we change it to instead of exact number of resources to just at least one resource in every table. as otherwise you can have more resources in one but no resources in others.

Good point, done in https://github.com/cloudquery/plugin-sdk/pull/85/commits/253c134e2ca00e37c34c35625ae32d3c82e0e96d

plugins/source_testing.go
```diff
@@ -23,34 +24,52 @@ func init() {
...
+}
+
+func validateTable(t *testing.T, table *schema.Table, resources []*schema.Resource) {
+	t.Helper()
+	resource := getTableResource(t, table, resources)
+	if resource != nil {
+		validateResource(t, resource)
+	} else {
+		t.Errorf("Expected table %s to be synced but it was not found", table.Name)
+	}
```
hermanschaaf, line 65: Bit of a nit, but typically linters will tell you to remove the `else` here and change it to something like:

```
if resource == nil {
   t.Errorf("Expected table %s to be synced but it was not found", table.Name)
   return
}
validateResource(t, resource)
```

This way it's a bit easier to follow the happy path logic by reading the top-level indent, with error handling happening on deeper indents.

hermanschaaf (approved):

plugins/source_testing.go
erezrokah: Good catch, fixed in https://github.com/cloudquery/plugin-sdk/pull/85/commits/c3f09268fa3065e0e1ea36736bd9e74d045084a2

hermanschaaf: Thanks! It's really not that important in this specific case, but to align on a general style guideline, in Go I believe it's preferable to have the error condition inside the `if` condition, with the happy path outside (with least indent). So I would write:

```
if resource == nil {
   t.Errorf("Expected table %s to be synced but it was not found", table.Name)
   return
}
validateResource(t, resource)
```

rather than

```
if resource != nil {
   validateResource(t, resource)
   return
}
t.Errorf("Expected table %s to be synced but it was not found", table.Name)
```

This way, when reading a function, the reader's eye can stay on the left-hand side and quickly see what a function does without needing to read about how error conditions are handled, unless they are interested in that aspect. Bonus points in this case for have a simple equality operator (`==`) over a negated one (`!=`) in the `if` condition.

The authors and main contributors of Go have discussed these points in various documents, but I'd be hard-pressed to find them now 😄

erezrokah: Cool, thanks for the additional context, I missed it in your first comment. Please let me know if https://github.com/cloudquery/plugin-sdk/pull/85/commits/78be2b3e71bfc4b2439a6b0f024227ce7075f545 works

hermanschaaf: Thanks, looks good! Didn't mean to nitpick, I've just found this a good rule of thumb for writing clean Go code, so I hope it's at least useful advice :)

erezrokah: 💯 Agree. I looked for a linter to enforce this but couldn't find one

hermanschaaf: I've been looking for an excuse to write my own linter! 😄

hermanschaaf: I'm pretty sure I'm not hallucinating that I've been warned about similar kinds of things by tooling in the past (especially a function that ends in if/else where the if branch has a return), but I also cannot find it now.

erezrokah: I think we have https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#early-return. Maybe it doesn't strictly fall under that rule

yevgenypats (approved):

PR #92, opened 2022-09-12
fix: Remove description from docs

erezrokah:
#### Summary

We don't generate the descriptions anymore (I think), so we should not generate them with the docs.
Also switched to snapshot testing so the output files are separated from the code.

Another possible implementation is to add an option to enable/disable the description column

---

- [ ] Update or add tests 🧪


hermanschaaf:
FWIW, some columns do still have descriptions, but agree, let's remove it from the docs for now since they're mostly empty.

hermanschaaf (approved):

PR #103, opened 2022-09-14
fix!: Rename MaxGoRoutines and enforce it

erezrokah:
#### Summary

When testing the Azure plugin I noticed we print a message on the number of Go routines used. From the other logs it seemed that the number is not enforced (All table fetch start logs were printed at the same time).
Since we can't really enforce a number of Go routines as sometimes we'd like to fetch related resources in parallel and we don't know those in advance, I renamed the configuration to `Concurrency` as I think we should still have a setting to control the load CloudQuery puts on the APIs (right now it creates a Go routine per top level table).
Also I switched to using channels instead of semaphore as seems more suitable for this use case, and we use channels in other places in the code.

---

- [x] Update or add tests 🧪

### DIFF
--- a/plugins/source.go
+++ b/plugins/source.go
@@ -162,7 +162,7 @@ func (p *SourcePlugin) Sync(ctx context.Context, spec specs.Source, res chan<- *
 	jobs := make(chan workerJob, jobsCount)
 	results := make(chan workerResult, jobsCount)
 
-	for w := uint64(1); w <= concurrency; w++ {
+	for i := uint64(0); i < concurrency; i++ {
 		go worker(jobs, results)
 	}
 
--- a/plugins/source.go
+++ b/plugins/source.go
@@ -3,7 +3,6 @@ package plugins
 import (
 	"context"
 	"fmt"
-	"sync"
 	"time"
 
 	"github.com/cloudquery/plugin-sdk/schema"
@@ -33,7 +32,7 @@ type SourcePlugin struct {
 
 type SourceOption func(*SourcePlugin)
 
-const minGoRoutines = 5
+const defaultConcurrency = 1
 
 func WithSourceExampleConfig(exampleConfig string) SourceOption {
 	return func(p *SourcePlugin) {
@@ -119,16 +118,12 @@ func (p *SourcePlugin) Sync(ctx context.Context, spec specs.Source, res chan<- *
 	}
 
 	// limiter used to limit the amount of resources fetched concurrently
-	maxGoroutines := spec.MaxGoRoutines
-	if maxGoroutines < minGoRoutines {
-		maxGoroutines = minGoRoutines
+	concurrency := spec.Concurrency
+	if concurrency == 0 {
+		concurrency = defaultConcurrency
 	}
-	p.logger.Info().Uint64("max_goroutines", maxGoroutines).Msg("starting fetch")
+	p.logger.Info().Uint64("concurrency", concurrency).Msg("starting fetch")
 
-	// goroutinesSem := semaphore.NewWeighted(helpers.Uint64ToInt64(maxGoroutines))
-
-	w := sync.WaitGroup{}
-	totalResources := 0
 	startTime := time.Now()
 	tableNames, err := p.interpolateAllResources(spec.Tables)
 	if err != nil {
@@ -136,8 +131,9 @@ func (p *SourcePlugin) Sync(ctx context.Context, spec specs.Source, res chan<- *
 	}
 
 	// this is the same fetchtime for all resources
-	fetchTime := time.Now()
+	syncTime := time.Now()
 
+	tableJobs := make([]workerJob, 0)
 	for _, table := range p.tables {
 		table := table
 		if funk.ContainsString(spec.SkipTables, table.Name) || !funk.ContainsString(tableNames, table.Name) {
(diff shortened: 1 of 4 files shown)


schema/table.go
```diff
@@ -198,6 +197,7 @@ func (t Table) Resolve(ctx context.Context, meta ClientMeta, syncTime time.Time,
 				}
 			}
 
+			totalResources++
```
erezrokah, line 200: It seems more clear to me to count each resource at a time when we send it to the channel instead of the current approach. Happy to revert as this is a personal opinion

yevgenypats:
I'd like to give it a go as well as I was dealing with a lot of issues around our concurrency and locks. I commented some of the stuff out because of lack of time and because I encountered some race conditions which I didn't have time to fix at the time.

erezrokah:
> I'd like to give it a go as well as I was dealing with a lot of issues around our concurrency and locks. I commented some of the stuff out because of lack of time and because I encountered some race conditions which I didn't have time to fix at the time.

💯 I understand as this is a core change. I think the challenge is that we can't predict how many "table.resolve` we have as when we have a relation we resolve it for each parent resource that was fetched. Hence we can only enforce the number of Go routines at the top level.
The only way I could think of of enforcing the max Go routines is instead of resolving the relations in the same Go routine as the parent, send it as a new job back to the worker queue. The problem with that approach is that we need to keep in memory all the resolved parent resources until the relation worker job is queued.

 
 > I deleted the comments as code in comments usually gets stale and leads to future confusion.

yevgenypats:
> > I'd like to give it a go as well as I was dealing with a lot of issues around our concurrency and locks. I commented some of the stuff out because of lack of time and because I encountered some race conditions which I didn't have time to fix at the time.
> 
> 💯 I understand as this is a core change. I think the challenge is that we can't predict how many `table.resolve` jobs we have as when we have a relation we resolve it for each parent resource that was fetched. Hence we can only enforce the number of Go routines at the top level. The only way I could think of of enforcing the max Go routines is instead of resolving the relations in the same Go routine as the parent, send it as a new job back to the worker channel/queue. The problem with that approach is that we need to keep in memory all the relation jobs until those are scheduled. For example if a top level table fetches 10,000 items and it has 4 relations, we'll queue 40,000 new jobs.
> 
> > I deleted the comments as code in comments usually gets stale and leads to future confusion.

I think this still wont solve it and will always cause deadlock. The only way I see is possible to solve it is with X parameters:

`concurrency_relations_1`, `concurrency_relations_2`, and so on. a limit on number of jobs wont solve as child relations can always cause deadlock for parent, doesn't matter if it's queue, worker channel, etc...

erezrokah:
> `concurrency_relations_1`, `concurrency_relations_2`

Yeah I thought about that too, but that's a dynamic number based on the depth of relations and also very hard to expose as a config.

> a limit on number of jobs wont solve as child relations can always cause deadlock for parent, doesn't matter if it's queue, worker channel, etc...

I think if you resolve only 1 level each time and each relation is a new job it won't deadlock as the parent already finished running (if we do a non blocking send or use a very big channel). But again that leads to memory issues as the number of jobs grows very high.

Hence this PR only implements a limit at the top level tables and avoids the term Go routines. 
If we don't have a limit at all the only way to control concurrency is to use multiple configuration files with a different set of resources per configuration.

On another note Azure has 55 top level tables, and if someone has 4 subscriptions the current code will create 220 Go routines. I'm actually not worried about the number of Go routines, but on the number of API calls and rate limiting.
⬆️ Might not be a problem yet, but it would be nice to be able to control it.

plugins/source.go
```diff
@@ -119,25 +118,22 @@ func (p *SourcePlugin) Sync(ctx context.Context, spec specs.Source, res chan<- *
...
-	totalResources := 0
 	startTime := time.Now()
 	tableNames, err := p.interpolateAllResources(spec.Tables)
 	if err != nil {
 		return err
 	}
 
-	// this is the same fetchtime for all resources
-	fetchTime := time.Now()
+	// syncTime is the time the sync opertaion started and the same for all resources
```
disq, line 133:
```suggestion
	// syncTime is the time the sync operation started and the same for all resources
```

plugins/source.go
```diff
@@ -119,25 +118,22 @@ func (p *SourcePlugin) Sync(ctx context.Context, spec specs.Source, res chan<- *
...
+	syncTime := time.Now()
 
+	tableJobs := make([]workerJob, 0)
```
disq, line 136:
```suggestion
	var tableJobs []workerJob
```

plugins/source.go
```diff
@@ -148,39 +144,40 @@ func (p *SourcePlugin) Sync(ctx context.Context, spec specs.Source, res chan<- *
...
+				res:      res,
+			})
+		}
+	}
+
+	jobsCount := len(tableJobs)
+	jobs := make(chan workerJob, jobsCount)
+	results := make(chan workerResult, jobsCount)
+
+	for w := uint64(1); w <= concurrency; w++ {
```
disq, line 165: Why not start from `0` and use `w < concurrency`? We don't use `w` anywhere else, it doesn't have to (weirdly 😂 ) start from 1.

plugins/source.go
```diff
@@ -148,39 +144,40 @@ func (p *SourcePlugin) Sync(ctx context.Context, spec specs.Source, res chan<- *
...
+	}
+
+	for i := 0; i < jobsCount; i++ {
+		jobs <- tableJobs[i]
+	}
+	close(jobs)
+
+	totalResources := 0
+	for i := 0; i < jobsCount; i++ {
+		result := <-results
```
disq, line 176: I would make the channels unbuffered and collect in a separate goroutine, to save on memory.

specs/source.go
```diff
@@ -14,12 +14,12 @@ type Source struct {
...
 	Path string `json:"path,omitempty"`
 	// Registry can be github,local,grpc. Might support things like https in the future.
-	Registry      Registry    `json:"registry,omitempty"`
-	MaxGoRoutines uint64      `json:"max_goroutines,omitempty"`
-	Tables        []string    `json:"tables,omitempty"`
-	SkipTables    []string    `json:"skip_tables,omitempty"`
-	Destinations  []string    `json:"destinations,omitempty"`
-	Spec          interface{} `json:"spec,omitempty"`
+	Registry     Registry    `json:"registry,omitempty"`
+	Concurrency  uint64      `json:"concurrency,omitempty"`
```
disq, line 18: I welcome this change as the `MaxGoRoutines` name is a bit too implementation-detaily.

plugins/source.go
erezrokah: Good point I'll change that

plugins/source.go
erezrokah: I'll make this change once @yevgenypats approves of the new approach

erezrokah:
Closing in favor of https://github.com/cloudquery/plugin-sdk/pull/129

PR #104, opened 2022-09-14
chore(test): Delete unused files

erezrokah:
#### Summary

Couldn't find a place these files are used

---

- [ ] Update or add tests 🧪


yevgenypats:
Are there tests to SpecReader ? If not I think worth adding those (I might've commented those out just to lack of time while I developed that / I wasn't sure what will be the final interface)

erezrokah:
> Are there tests to SpecReader ? If not I think worth adding those (I might've commented those out just to lack of time while I developed that / I wasn't sure what will be the final interface)

I initially thought `spec_test` cover those, but I was wrong. Re-introduced the tests in https://github.com/cloudquery/plugin-sdk/pull/110

PR #115, opened 2022-09-16
fix: Improve error message on codegen field error

hermanschaaf:
This makes it a bit easier to find which table is being referenced when running codegen.


yevgenypats (approved):
Sounds good. we also need to move it to zero log as well, I'll do that in a follow-up PR but not urgent

Issue #121, opened 2022-09-18
feature: Add version command


erezrokah:
Should be useful for debugging user issues

PR #123, opened 2022-09-18
feat!: Remove source plugins GetVersion method

erezrokah:
#### Summary

With https://github.com/cloudquery/cloudquery/pull/1856 this shouldn't be required. We should also consider removing it from destination plugins as those have hardcoded versions at the moment

---

- [ ] Update or add tests 🧪


yevgenypats:
@erezrokah So I can see that this function is not mandatory but I thought it might be quiet useful for example to have sanity tests between the verison downloaded and the version reported by the plugin itself (which is compiled in). wdyt ?

erezrokah:
This PR only removes the gRPC method as the SDK method is used in other places https://github.com/cloudquery/plugin-sdk/blob/5129165bdab225715cb3f279ce21100842de81ed/serve/serve.go#L89

erezrokah:
> version downloaded and the version reported by the plugin itself (which is compiled in)

How about using https://github.com/cloudquery/plugin-sdk/issues/121 for that?

yevgenypats:
> > version downloaded and the version reported by the plugin itself (which is compiled in)
> 
> How about using #121 for that?

It can be an addition but when the CLI is executing a plugin it will be complex to parse something from stdout rather from a gRPC which is already in place.

But not sure I understand why we need to remove this additional function which returns relatively important info which I can expect to be needed just like name.

The sdk version is used on the server side (for sentry) - it cannot be used on the client side because it is in a different process.

erezrokah:
> It can be an addition but when the CLI is executing a plugin it will be complex to parse something from stdout rather from a gRPC which is already in place.

Ah you meant a sanity check in the CLI itself? I was thinking testing it in integration tests.
I think embedding the version in a plugin should not be a requirement as we've seen that often plugin authors (including ourselves) forget to embed the version during release.
We could however print a warning version if the versions don't match if we find it useful.

I don't mind keeping it, it doesn't make a difference at the moment, just didn't want to keep an API that's not used.
Lets close the PR and re-evaluate it

PR #126, opened 2022-09-18
feat!: Idiomatic serve interface

yevgenypats:
This is a breaking change for the serve interface so we can easily add more options without breaking the interface and move errors to compile time instead of runtime.

This should be the last breaking change before v2 release this week as I want to make sure our SDK API is extensible because we wont be breaking it for a while as SDK breaking changes are very expensive.

#### Summary

---

- [ ] Update or add tests 🧪


erezrokah (approved):

serve/destination.go
```diff
@@ -0,0 +1,146 @@
...
+				return fmt.Errorf("failed to serve: %w", err)
+			}
+			return nil
+		},
+	}
+	cmd.Flags().StringVar(&address, "address", "localhost:7777", "address to serve on. can be tcp: `localhost:7777` or unix socket: `/tmp/plugin.rpc.sock`")
+	cmd.Flags().StringVar(&network, "network", "tcp", `the network must be "tcp", "tcp4", "tcp6", "unix" or "unixpacket"`)
+	cmd.Flags().Var(logLevel, "log-level", fmt.Sprintf("log level. one of: %s", strings.Join(logLevel.Allowed, ",")))
+	cmd.Flags().Var(logFormat, "log-format", fmt.Sprintf("log format. one of: %s", strings.Join(logFormat.Allowed, ",")))
+	cmd.Flags().BoolVar(&noSentry, "no-sentry", false, "disable sentry")
```
hermanschaaf, line 134:
```suggestion
	cmd.Flags().BoolVar(&noSentry, "no-sentry", false, "disable Sentry telemetry")
```

hermanschaaf: Should we maybe use `no-telemetry` rather than `no-sentry`? We may want to change the underlying provider at some point, and developers shouldn't need to care too much about what we are using

serve/destination.go
```diff
@@ -0,0 +1,146 @@
...
+
+type DestinationOption func(*destinationServe)
+
+func WithDestinationSentryDSN(dsn string) DestinationOption {
+	return func(s *destinationServe) {
+		s.sentryDSN = dsn
+	}
+}
+
+func Destination(plugin plugins.DestinationPlugin, opts ...DestinationOption) {
```
hermanschaaf, line 36: Are we happy with the inconsistency that destination plugins use an interface while source plugins use a struct pointer? Should we perhaps make both the same? Asking because now would be the best time to change it 😅

hermanschaaf (changes requested):

serve/destination.go
erezrokah: 👍 https://github.com/cloudquery/plugin-sdk/pull/54#discussion_r963330061

serve/destination.go
yevgenypats: I think yes but destinations interface we still have time to change (I remember you had that code somewhere).

serve/destination.go
yevgenypats: I think it should be two different stuff as it's error-reporting and telemetry. another option `telemtry=error-reporting,all,none`

yevgenypats:
@erezrokah can you remind me how you ignore a linter warning - It complains rightfully about duplicate code but I think in this case it is better to have the duplicate rather then functions that abstract that given that it is two different entry points.

serve/destination.go
hermanschaaf: That's fine, as long as the change we plan on making is to make the destination into a struct pointer as well (and yeah, I had a branch doing that a while back, can dig it up again if we need it)

erezrokah:
> @erezrokah can you remind me how you ignore a linter warning - It complains rightfully about duplicate code but I think in this case it is better to have the duplicate rather then functions that abstract that given that it is two different entry points.

This should fix it https://github.com/cloudquery/plugin-sdk/pull/126/commits/d6560bb69d7f7ed735f2d51c5ee59877bdca076b

serve/destination.go
hermanschaaf: I think `telemetry` usually means "anything that phones home", including metrics, logs and error reports. So `no-telemetry` would not send any data to us at all. That's probably granular enough for now, and we can introduce further options in the future if we feel it becomes necessary?

yevgenypats: Opened this instead https://github.com/cloudquery/plugin-sdk/issues/128

hermanschaaf (approved):

PR #129, opened 2022-09-19
fix: Bring concurrency back

yevgenypats:
We had concurrency limits disabled. Fixed it and also fixed the issue that we had with timings (now the time really shows how long it took for a goroutine to fetch without wait time).

This should work instead of https://github.com/cloudquery/plugin-sdk/pull/103 

---

- [ ] Update or add tests 🧪


plugins/source.go
```diff
@@ -155,39 +158,26 @@ func (p *SourcePlugin) Sync(ctx context.Context, spec specs.Source, res chan<- *
...
-				go func() {
-					defer wg.Done()
-					// defer goroutinesSem.Release(1)
-					totalTableResources += table.Resolve(ctx, client, fetchTime, nil, res)
-				}()
+		for _, client := range clients {
+			client := client
+			wg.Add(1)
+			if err := goroutinesSem.Acquire(ctx, 1); err != nil {
+				// This mean context was cancelled
```
hermanschaaf, line 165:
```suggestion
				// This means context was cancelled
```

plugins/source.go
```diff
@@ -155,39 +158,26 @@ func (p *SourcePlugin) Sync(ctx context.Context, spec specs.Source, res chan<- *
...
-		}()
+			go func() {
+				defer wg.Done()
+				defer goroutinesSem.Release(1)
+				// TODO: prob introduce client.Identify() to be used in logs
+				tableStartTime := time.Now()
+				p.logger.Info().Str("table", table.Name).Msg("fetch start")
+				totalTableResources := table.Resolve(ctx, client, startTime, nil, res)
+				totalResources += totalTableResources
+				p.logger.Info().Str("table", table.Name).Int("total_resources", totalTableResources).TimeDiff("duration", time.Now(), tableStartTime).Msg("fetch table finished")
```
hermanschaaf, line 176: `table` should be passed in to this function as a parameter, so we're not using the variable from the scope outside this goroutine

hermanschaaf (changes requested):

plugins/source.go
```diff
@@ -33,7 +35,10 @@ type SourcePlugin struct {
 
 type SourceOption func(*SourcePlugin)
 
-const minGoRoutines = 5
+const (
+	minGoRoutines        = 5
+	defaultMaxGoRoutines = 500000
```
erezrokah, line 40:
```suggestion
	defaultMaxGoRoutines = 500000
```

Does this mean that by default we spin up `500000` go routines? Also what's the reason for having a hard `min`? Shouldn't we have only `GoRoutinesCount` and ensure it's not < 0 (0 meaning default value)

plugins/source.go
yevgenypats: I think this should be fine because I do `table := table` and `client := client` so this workarounds the shadow variable thing.

erezrokah (changes requested):
This looks much better 🚀 Added a few comments

plugins/source.go
hermanschaaf: Ah thanks, I missed those shadow declarations 👍

hermanschaaf (approved):

plugins/source.go
yevgenypats: Yeah I guess you are right - it was from older piece of code where we had an issue with 1-4 go routines.

plugins/source.go
```diff
@@ -127,24 +132,22 @@ func (p *SourcePlugin) Sync(ctx context.Context, spec specs.Source, res chan<- *
 
 	// limiter used to limit the amount of resources fetched concurrently
 	maxGoroutines := spec.MaxGoRoutines
```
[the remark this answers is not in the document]
yevgenypats: done

plugins/source.go
```diff
@@ -126,25 +130,20 @@ func (p *SourcePlugin) Sync(ctx context.Context, spec specs.Source, res chan<- *
 	}
 
 	// limiter used to limit the amount of resources fetched concurrently
-	maxGoroutines := spec.MaxGoRoutines
-	if maxGoroutines < minGoRoutines {
-		maxGoroutines = minGoRoutines
+	concurrency := spec.Concurrency
+	if concurrency == 0 {
+		concurrency = defaultMaxGoRoutines
```
erezrokah, line 135:
```suggestion
		concurrency = defaultConcurrency
```

plugins/source.go
```diff
@@ -33,7 +35,9 @@ type SourcePlugin struct {
 
 type SourceOption func(*SourcePlugin)
 
-const minGoRoutines = 5
+const (
+	defaultMaxGoRoutines = 500000
```
erezrokah, line 39:
```suggestion
	defaultConcurrency = 500000
```

erezrokah (approved):
Approved with a comment on the default var name.

Also this is also a breaking change so better to merge as `fix!:`.

Finally I think the current default means "unlimited" at the moment. Not sure that's the best default as we might hit rate limits quite fast

yevgenypats:
> Approved with a comment on the default var name.
> 
> Also this is also a breaking change so better to merge as `fix!:`.
> 
> Finally I think the current default means "unlimited" at the moment. Not sure that's the best default as we might hit rate limits quite fast

Thanks! The default means 500K which is I think reasonable for 1 CPU 4GB RAM which should be pretty much anyone

erezrokah:
> The default means 500K which is I think reasonable for 1 CPU 4GB RAM which should be pretty much anyone

I should have explained it better. I'm not worried about the machine specs, but being throttled/rate limited by the service API

yevgenypats:
> > The default means 500K which is I think reasonable for 1 CPU 4GB RAM which should be pretty much anyone
> 
> I should have explained it better. I'm not worried about the machine specs, but being throttled/rate limited by the service API

I think concurrency is the right parameter are mainly to deal with machine spec and too many open files and so on. and max_back_off, max_retries are usually the right params to deal with throttle on the server side

PR #135, opened 2022-09-19
feat: Resolve proto timestamp to `schema.TypeTimestamp`

candiduslynx:
#### Summary

---

- [x] Update or add tests 🧪

### DIFF
--- a/codegen/golang.go
+++ b/codegen/golang.go
@@ -33,8 +33,8 @@ func valueToSchemaType(v reflect.Type) (schema.ValueType, error) {
 	case reflect.Map:
 		return schema.TypeJSON, nil
 	case reflect.Struct:
-		t := v.PkgPath() + "." + v.Name()
-		if t == "time.Time" {
+		switch t := v.PkgPath() + "." + v.Name(); t {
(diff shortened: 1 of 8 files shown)


codegen/golang.go
```diff
@@ -33,8 +33,8 @@ func valueToSchemaType(v reflect.Type) (schema.ValueType, error) {
 	case reflect.Map:
 		return schema.TypeJSON, nil
 	case reflect.Struct:
-		t := v.PkgPath() + "." + v.Name()
-		if t == "time.Time" {
+		switch t := v.PkgPath() + "." + v.Name(); t {
+		case "time.Time", "google.golang.org/protobuf/types/known/timestamppb.Timestamp":
```
erezrokah, line 37: I'm not sure we want the SDK to be aware of specific non built-in Go types

yevgenypats: I agree with @erezrokah . Here it's not so bad but in `column.go` it makes an import which will cause all plugins to import unrelated libraries and grow in size which we should be careful with.

candiduslynx: Well, at the very least we can do a custom mapping for the type, but we need a custom mapper for the value as well (`timestamppb.Timestamp` -> `time.Time`)

PR #137, opened 2022-09-19
fix: Skip fields that have "-" json tag

amanenk:
#### Summary

If field has a `json: "-"` tag that means that it is not the data parsed from api. so such a field should be skipped.

---

- [ ] Update or add tests 🧪

### DIFF
--- a/codegen/golang.go
+++ b/codegen/golang.go
@@ -90,6 +90,10 @@ func WithUnwrapAllEmbeddedStructs() TableOptions {
 func defaultTransformer(field reflect.StructField) string {
 	name := field.Name
 	if jsonTag := strings.Split(field.Tag.Get("json"), ",")[0]; len(jsonTag) > 0 {
+		// return empty string if the field is not related api response
+		if jsonTag == "-" {
+			return ""
+		}
 		name = jsonTag
 	}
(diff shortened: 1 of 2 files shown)


erezrokah:
Thanks @amanenk 👍 Can you share an example struct with this use case? It will make it easier to understand the reasoning behind the PR

candiduslynx:
> Thanks @amanenk 👍 Can you share an example struct with this use case? It will make it easier to understand the reasoning behind the PR

from https://pkg.go.dev/encoding/json
> As a special case, if the field tag is "-", the field is always omitted. Note that a field with name "-" can still be generated using the tag "-,".
I suppose, we don't need this special case unless someone specifically asks for it, so the default behavior of masking/skipping such fields is better IMO.

See https://pkg.go.dev/google.golang.org/api@v0.96.0/dns/v1#ManagedZone for example (notice the last fields that are of use only to SDK).

erezrokah:
Thanks @candiduslynx that makes sense 🚀

erezrokah (approved):

hermanschaaf (approved):

hermanschaaf:
@amanenk Could we please add a test case for this before merging?

PR #139, opened 2022-09-20
feat(codegen): Allow consumers to override field types

erezrokah:
#### Summary

I guess this is a difference way to do https://github.com/cloudquery/plugin-sdk/pull/135 in order to fix write errors in the Azure plugin. See corresponding PR https://github.com/cloudquery/cloudquery/pull/1885

---

- [ ] Update or add tests 🧪

### DIFF
--- a/codegen/golang.go
+++ b/codegen/golang.go
@@ -87,7 +87,7 @@ func WithUnwrapAllEmbeddedStructs() TableOptions {
 	}
 }
 
-// Allows overriding the schema type for a specific field. Return `nil` to fallback to default behavior
+// WithValueTypeOverride sets a function that can override the schema type for specific fields. Return `nil` to fall back to default behavior.
 func WithValueTypeOverride(resolver func(reflect.StructField) *schema.ValueType) TableOptions {
 	return func(t *TableDefinition) {
 		t.valueTypeOverride = resolver
--- a/codegen/golang.go
+++ b/codegen/golang.go
@@ -33,8 +33,8 @@ func valueToSchemaType(v reflect.Type) (schema.ValueType, error) {
 	case reflect.Map:
 		return schema.TypeJSON, nil
 	case reflect.Struct:
-		typeString := v.PkgPath() + "." + v.Name()
-		if typeString == "time.Time" {
+		t := v.PkgPath() + "." + v.Name()
+		if t == "time.Time" {
 			return schema.TypeTimestamp, nil
 		}
 		return schema.TypeJSON, nil
--- a/codegen/golang.go
+++ b/codegen/golang.go
@@ -19,8 +19,8 @@ type TableOptions func(*TableDefinition)
 var TemplatesFS embed.FS
 
 func (t TableDefinition) valueToSchemaType(v reflect.Type) (schema.ValueType, error) {
-	if t.valueToSchemaTypeOverride != nil {
-		if vt := t.valueToSchemaTypeOverride(v); vt != nil {
+	if t.valueTypeOverride != nil {
+		if vt := t.valueTypeOverride(v); vt != nil {
 			return *vt, nil
 		}
 	}
@@ -92,9 +92,10 @@ func WithUnwrapAllEmbeddedStructs() TableOptions {
 	}
 }
 
-func WithValueToSchemaType(resolver func(reflect.Type) *schema.ValueType) TableOptions {
+// Allows overriding the schema type for a specific field. Return `nil` to fallback to default behavior
+func WithValueTypeOverride(resolver func(reflect.Type) *schema.ValueType) TableOptions {
 	return func(t *TableDefinition) {
-		t.valueToSchemaTypeOverride = resolver
+		t.valueTypeOverride = resolver
 	}
 }
 
--- a/codegen/golang.go
+++ b/codegen/golang.go
@@ -7,7 +7,6 @@ import (
 	"reflect"
 	"strings"
 	"text/template"
-	"time"
 	"unicode"
 
 	"github.com/cloudquery/plugin-sdk/schema"
@@ -39,8 +38,8 @@ func (t TableDefinition) valueToSchemaType(v reflect.Type) (schema.ValueType, er
 	case reflect.Map:
 		return schema.TypeJSON, nil
 	case reflect.Struct:
-		timeValue := time.Time{}
-		if v == reflect.TypeOf(timeValue) {
+		typeString := v.PkgPath() + "." + v.Name()
+		if typeString == "time.Time" {
 			return schema.TypeTimestamp, nil
 		}
 		return schema.TypeJSON, nil
--- a/codegen/golang.go
+++ b/codegen/golang.go
@@ -19,7 +19,12 @@ type TableOptions func(*TableDefinition)
 //go:embed templates/*.go.tpl
 var TemplatesFS embed.FS
 
-func valueToSchemaType(v reflect.Type) (schema.ValueType, error) {
+func (t TableDefinition) valueToSchemaType(v reflect.Type) (schema.ValueType, error) {
+	if t.valueToSchemaTypeOverride != nil {
+		if vt := t.valueToSchemaTypeOverride(v); vt != nil {
+			return *vt, nil
+		}
+	}
 	k := v.Kind()
 	switch k {
 	case reflect.String:
@@ -40,7 +45,7 @@ func valueToSchemaType(v reflect.Type) (schema.ValueType, error) {
 		}
 		return schema.TypeJSON, nil
 	case reflect.Pointer:
-		return valueToSchemaType(v.Elem())
+		return t.valueToSchemaType(v.Elem())
 	case reflect.Slice:
 		switch v.Elem().Kind() {
 		case reflect.String:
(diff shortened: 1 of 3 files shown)


codegen/golang.go
```diff
@@ -139,7 +145,14 @@ func (t *TableDefinition) addColumnFromField(field reflect.StructField, parent *
 		return
 	}
 
-	columnType, err := valueToSchemaType(field.Type)
+	var columnType schema.ValueType
+	var err error
+	if t.valueToSchemaType != nil {
+		columnType, err = t.valueToSchemaType(field.Type)
+	}
+	if err != nil {
```
erezrokah, line 153: Maybe better to return a pointer instead of an error to do the `nil` check

codegen/golang.go
```diff
@@ -87,6 +93,12 @@ func WithUnwrapAllEmbeddedStructs() TableOptions {
 	}
 }
 
+func WithValueToSchemaType(resolver func(reflect.Type) *schema.ValueType) TableOptions {
```
erezrokah, line 96: Happy to use a better name here

codegen/golang.go
```diff
@@ -33,13 +39,13 @@ func valueToSchemaType(v reflect.Type) (schema.ValueType, error) {
 	case reflect.Map:
 		return schema.TypeJSON, nil
 	case reflect.Struct:
-		t := v.PkgPath() + "." + v.Name()
-		if t == "time.Time" {
+		timeValue := time.Time{}
+		if v == reflect.TypeOf(timeValue) {
```
erezrokah, line 43: This seems like a safer way to check for types instead of string comparison

yevgenypats: Im not sure this will work as `TypeOf` will return Struct

yevgenypats: cc @hermanschaaf I think battled with that recently

yevgenypats (changes requested):
This needs some more testing and I think we need to add the string comparison back, I recall there wasn't another way, maybe @hermanschaaf can weigh in as I dont remember on top of my head.

codegen/golang.go
hermanschaaf: There's a test case for this and looks like it passes 👍  I spent a few minutes a while back and couldn't figure out how to make it work, but this looks good!

erezrokah: Ah, let me check it doesn't return `true` for all structs

hermanschaaf: Ah yeah, that could be 😅

erezrokah: >Im not sure this will work as TypeOf will return Struct

`reflect.TypeOf(timeValue)` returns an instance of `reflect.Type` so probably has all the information about the type to do the comparison

erezrokah: But this change is unrelated to this PR, so I'll open another one for it

hermanschaaf: Good idea, let's discuss in another PR. But I think it must work, since the test case includes a struct that gets converted correctly, as well as the `*time.Time` and `time.Time` cases.

hermanschaaf (changes requested):
I think this makes sense, and will enable us to handle a wide range of cases that we will eventually need to handle. It also seems better than trying to handle every possible type in the SDK. 

I wonder if the `reflect.Type` passed in will be enough to handle all the cases though? What if the value type decision depended on the name of the field in a struct? Like maybe you'd want to convert things with `Date` in the name to timestamp, even though their Go type is string. For that reason, we may want to provide the full `field reflect.StructField`, rather than just the `reflect.Type` of the field.

Also, could you add a test case or two for this functionality?

erezrokah:
> I wonder if the `reflect.Type` passed in will be enough to handle all the cases though? What if the value type decision depended on the name of the field in a struct? Like maybe you'd want to convert things with `Date` in the name to timestamp, even though their Go type is string. For that reason, we may want to provide the full `field reflect.StructField`, rather than just the `reflect.Type` of the field.

Good point I was wondering that too. Lets go with passing the `field`. It makes the code a bit let pretty but I think that's fine

codegen/golang.go
```diff
@@ -87,6 +87,13 @@ func WithUnwrapAllEmbeddedStructs() TableOptions {
 	}
 }
 
+// Allows overriding the schema type for a specific field. Return `nil` to fallback to default behavior
+func WithValueTypeOverride(resolver func(reflect.StructField) *schema.ValueType) TableOptions {
```
hermanschaaf, line 91: Let's change it to standard godoc style
```suggestion
// WithValueTypeOverride sets a function that can override the schema type for specific fields. Return `nil` to fall back to default behavior.
func WithValueTypeOverride(resolver func(reflect.StructField) *schema.ValueType) TableOptions {
```

hermanschaaf: (We should probably apply this consistently everywhere, but good to start somewhere...)

codegen/golang.go
```diff
@@ -139,7 +146,17 @@ func (t *TableDefinition) addColumnFromField(field reflect.StructField, parent *
 		return
 	}
 
-	columnType, err := valueToSchemaType(field.Type)
+	var overrideType *schema.ValueType
+	if t.valueTypeOverride != nil {
+		overrideType = t.valueTypeOverride(field)
```
yevgenypats, line 151: I think the type definition for `valueTypeOverride` should be the same as `valueToSchemaType` i.e also return an error so we propogate this, log it, have consistent behaviour and so on.

codegen/golang.go
```diff
@@ -87,6 +87,13 @@ func WithUnwrapAllEmbeddedStructs() TableOptions {
 	}
 }
 
+// WithValueTypeOverride sets a function that can override the schema type for specific fields. Return `nil` to fall back to default behavior.
+func WithValueTypeOverride(resolver func(reflect.StructField) *schema.ValueType) TableOptions {
```
candiduslynx, line 91: let's do the `resolver func(any) schema.ValueType`?
So that I'm able to pass `codegen.WithValueTypeOverride(time.Diration{})` for instance

codegen/golang.go
disq: Maybe remove the default value fallback. If there's an override, always use that. The override can decide to call `ValueToSchemaType` on its own if it doesn't want any changes (we need to export that fn...)

candiduslynx (changes requested):

codegen/golang.go
yevgenypats: Yeah agree, also no need for a fallback. i.e if there is an error then it just errors out. (returning nil will panic but that's fine as this is undefined behaviour if no error is returned)

erezrokah: Do we want to expose `valueToSchemaType`? It's private at the moment

disq: if we remove the fallback and expose the inner func then no need for it to return a ptr, it can have the exact same signature.

erezrokah: > Yeah agree, also no need for a fallback. i.e if there is an error then it just errors out. (returning nil will panic but that's fine as this is undefined behaviour if no error is returned)

Should we expose `valueToSchemaType` as a package function or a struct function for the table?
I'm not sure that looks nicer that handling the fallback ourselves

candiduslynx (approved):

codegen/golang.go
```diff
@@ -88,6 +88,13 @@ func WithUnwrapAllEmbeddedStructs() TableOptions {
 	}
 }
 
+// WithValueTypeOverride sets a function that can override the schema type for specific fields. Return `nil` to fall back to default behavior.
+func WithValueTypeOverride(resolver func(reflect.StructField) (*schema.ValueType, error)) TableOptions {
```
yevgenypats, line 92: nit: If we return an error now then I don't think we need a pointer anymore just `schema.ValueType` should do it.

erezrokah:
I think I messed up the rebase, let me check this PR again

erezrokah:
@candiduslynx can you please check, I think I fixed the bad rebase I did and @yevgenypats please review

codegen/golang_test.go
```diff
@@ -155,13 +159,31 @@ func TestTableFromGoStruct(t *testing.T) {
...
+					switch a.(type) {
+					case time.Time, *time.Time:
+						return schema.TypeJSON
+					default:
+						return schema.TypeInvalid
+					}
+				})},
+			},
+			want: TableDefinition{Name: "test_struct",
+				// We expect the time column to be of type JSON, since we override the type of all structs to be JSON
```
erezrokah, line 183:
```suggestion
				// We expect the time column to be of type JSON, since we override the type of `time.Time` to be JSON
```

codegen/golang.go
candiduslynx: Yes, I had it in https://github.com/erezrokah/plugin-sdk/pull/1

candiduslynx (approved):

hermanschaaf (approved):

codegen/golang.go
erezrokah: The confusion is my fault as @yevgenypats reviewed my botched rebase. The PR should be up to date now

erezrokah:
Continued in https://github.com/cloudquery/plugin-sdk/pull/157

Issue #114, opened 2022-09-15
Question: Should we be counting resources from relations under the total count?


erezrokah:
We count the total number of resources including relations:
https://github.com/cloudquery/plugin-sdk/blob/b45c12e4ab381a7f5bd30e5871489aebf7498525/schema/table.go#L203

However when we print it we use the top level table name:
https://github.com/cloudquery/plugin-sdk/blob/b45c12e4ab381a7f5bd30e5871489aebf7498525/plugins/source.go#L180

Which can be confusing as if the user takes the name of the table from the logs and run a count on the database, they might experience a different number of resources.

Should we only consider top level resources in the count? Maybe find a better way to print the count?

PR #143, opened 2022-09-20
fix: Don't count relations under total resources for top level tables

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/plugin-sdk/issues/114

---

- [ ] Update or add tests 🧪


schema/table.go
```diff
@@ -200,9 +199,10 @@ func (t Table) Resolve(ctx context.Context, meta ClientMeta, syncTime time.Time,
 				}
 			}
 
+			totalResources++
```
erezrokah, line 202: Incrementing when we send the result to the channel seems more clear to me than counting all the objects before the loop

erezrokah: Maybe worth adding a separate log message for relations if we don't have one

hermanschaaf:
Counter-point: I think we _should_ count relations, or at least report them; we're doing non-trivial work and API calls to fetch them. The ideal would be to report on relational table counts separately--not sure how easy that would be to do in practice

erezrokah:
> Counter-point: I think we _should_ count relations, or at least report them; we're doing non-trivial work and API calls to fetch them. The ideal would be to report on relational table counts separately--not sure how easy that would be to do in practice

💯 Agree we should count them separately. Let me see if I can add it to this PR

yevgenypats (approved):

PR #144, opened 2022-09-20
feat: Validate undefined column in TestResource

yevgenypats:
I found quite a few issue where we use resource.Set but we dont define the column and the destination pluginfails to write because the column doesn't exist

#### Summary

---

- [ ] Update or add tests 🧪

### DIFF
--- a/plugins/source_testing.go
+++ b/plugins/source_testing.go
@@ -2,6 +2,7 @@ package plugins
 
 import (
 	"context"
+	"encoding/json"
 	"testing"
 
 	"github.com/cloudquery/faker/v3"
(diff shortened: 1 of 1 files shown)


erezrokah (approved):

disq (approved):

plugins/source_testing.go
```diff
@@ -75,9 +76,36 @@ func validateTables(t *testing.T, tables schema.Tables, resources []*schema.Reso
...
+	}
 	for _, columnName := range resource.Table.Columns.Names() {
-		if resource.Get(columnName) == nil && !resource.Table.Columns.Get(columnName).IgnoreInTests {
+		if data[columnName] == nil && !resource.Table.Columns.Get(columnName).IgnoreInTests {
 			t.Errorf("table: %s with unset column %s", resource.Table.Name, columnName)
 		}
+		val := data[columnName]
+		if val != nil {
+			switch resource.Table.Columns.Get(columnName).Type {
+			case schema.TypeJSON:
```
hermanschaaf, line 95: I think there's also going to be a new class of bug where we set a JSON column using a string or byte slice (that was already marshalled to JSON). It's very rare, if ever, that we'd want to write a string directly to a JSON column, even though it would marhsal to JSON just fine, so maybe we can check for that issue here

PR #150, opened 2022-09-20
feat(codegen): Add custom resolvers for types

candiduslynx:
This will go with #139


codegen/table.go
```diff
@@ -1,51 +1,149 @@
...
+	// skip field if there is no name
+	if name == "" {
+		return
+	}
+	if parent != nil {
+		path = parent.Name + `.` + path
+		name = t.nameTransformer(*parent) + "_" + name
+	}
+	resolver := `schema.PathResolver("` + path + `")`
+	resolverProvider, ok := t.customTypeResolvers[columnType]
```
erezrokah, line 82: Should `customTypeResolvers` return the full resolver here? I would expect that if someone overrides a resolver, they override all of it (including the path)

yevgenypats:
Any chance to do the structure changes in a different PR ? as otherwise hard to know what was changed and what was moved ?

yevgenypats (changes requested):
I would love the code + restructure to be in different PRs or at least different commits as very hard for me to review this and understand the diff

codegen/table.go
candiduslynx: No, I think that only the func that takes a path will be more robust (allowing overrides for unwrapped fields as well)

candiduslynx:
Closing this to make new one for restructure and a separate for the resolver after #139 is merged

PR #155, opened 2022-09-21
fix: Codegen: Add specific case to handle k8s's 'metav1.Time'

shimonp21:
Attempting to insert it as a json fails because it formats to "2022-01-01T00:01:02" which is not a valid json...

#### Summary

---

- [ ] Update or add tests 🧪


yevgenypats (changes requested):
Im working now on https://github.com/cloudquery/plugin-sdk/pull/139 which will remove the need for this as otherwise we will have to bundle it infinite number of libraries in the SDK

codegen/golang.go
```diff
@@ -12,6 +12,7 @@ import (
 
 	"github.com/cloudquery/plugin-sdk/schema"
 	"github.com/iancoleman/strcase"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
yevgenypats, line 15: we shouldn't import anything plugin specific in the SDK. working on a solution.

erezrokah: This is the related PR https://github.com/cloudquery/plugin-sdk/pull/139

shimonp21:
Closing in favour of https://github.com/cloudquery/plugin-sdk/pull/139

PR #157, opened 2022-09-21
feat(codegen): Add WithTypeTransformer

yevgenypats:
This is take two of https://github.com/cloudquery/plugin-sdk/pull/139

Trying to get right the `WithTypeTransformer` so we wont have to change it in the near future as it's heavilly used external API.

I've also added error handling to all functions as we were basically ignoring all of them and printing to the screen is not the right approach catching those locally and in CI.

---

- [ ] Update or add tests 🧪

### DIFF
--- a/codegen/golang.go
+++ b/codegen/golang.go
@@ -87,7 +87,7 @@ func WithUnwrapAllEmbeddedStructs() TableOptions {
 	}
 }
 
-// Allows overriding the schema type for a specific field. Return `nil` to fallback to default behavior
+// WithValueTypeOverride sets a function that can override the schema type for specific fields. Return `nil` to fall back to default behavior.
 func WithValueTypeOverride(resolver func(reflect.StructField) *schema.ValueType) TableOptions {
 	return func(t *TableDefinition) {
 		t.valueTypeOverride = resolver
--- a/codegen/golang.go
+++ b/codegen/golang.go
@@ -33,8 +33,8 @@ func valueToSchemaType(v reflect.Type) (schema.ValueType, error) {
 	case reflect.Map:
 		return schema.TypeJSON, nil
 	case reflect.Struct:
-		typeString := v.PkgPath() + "." + v.Name()
-		if typeString == "time.Time" {
+		t := v.PkgPath() + "." + v.Name()
+		if t == "time.Time" {
 			return schema.TypeTimestamp, nil
 		}
 		return schema.TypeJSON, nil
--- a/codegen/golang.go
+++ b/codegen/golang.go
@@ -19,8 +19,8 @@ type TableOptions func(*TableDefinition)
 var TemplatesFS embed.FS
 
 func (t TableDefinition) valueToSchemaType(v reflect.Type) (schema.ValueType, error) {
-	if t.valueToSchemaTypeOverride != nil {
-		if vt := t.valueToSchemaTypeOverride(v); vt != nil {
+	if t.valueTypeOverride != nil {
+		if vt := t.valueTypeOverride(v); vt != nil {
 			return *vt, nil
 		}
 	}
@@ -92,9 +92,10 @@ func WithUnwrapAllEmbeddedStructs() TableOptions {
 	}
 }
 
-func WithValueToSchemaType(resolver func(reflect.Type) *schema.ValueType) TableOptions {
+// Allows overriding the schema type for a specific field. Return `nil` to fallback to default behavior
+func WithValueTypeOverride(resolver func(reflect.Type) *schema.ValueType) TableOptions {
 	return func(t *TableDefinition) {
-		t.valueToSchemaTypeOverride = resolver
+		t.valueTypeOverride = resolver
 	}
 }
 
--- a/codegen/golang.go
+++ b/codegen/golang.go
@@ -7,7 +7,6 @@ import (
 	"reflect"
 	"strings"
 	"text/template"
-	"time"
 	"unicode"
 
 	"github.com/cloudquery/plugin-sdk/schema"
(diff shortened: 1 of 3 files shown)


codegen/golang.go
```diff
@@ -19,9 +20,11 @@ type TableOptions func(*TableDefinition)
 //go:embed templates/*.go.tpl
 var TemplatesFS embed.FS
 
-func valueToSchemaType(v reflect.Type) (schema.ValueType, error) {
+func defaultTypeTransformer(v reflect.Type) (schema.ValueType, error) {
```
erezrokah, line 23: Is the default transformer exposed to consumers? Otherwise any consumer that overrides will need to re-implement it

codegen/table.go
```diff
@@ -4,13 +4,17 @@ import (
...
+	"github.com/grpc-ecosystem/go-grpc-middleware/providers/zerolog/v2"
 )
 
 type ResourceDefinition struct {
 	Name  string
 	Table *TableDefinition
 }
 
+type NameTransformer func(reflect.StructField) (string, error)
+type TypeTransformer func(reflect.Type) (schema.ValueType, error)
```
erezrokah, line 16: I think both @hermanschaaf and @candiduslynx suggested to extend this API to allow either the field or any type.

erezrokah (changes requested):
Thanks @yevgenypats added a few comments that were discussed in the previous PR

codegen/table.go
yevgenypats: if it will be `any` then the codegen will need to potentially invoke the type and create the objects instead of the custom transformer to do `reflect.TypeOf`. I need to see the code though for the `any` but I feel it might be more prone to issues. `StructField` is fine imo I was on the fence here but I can go with StructField so the transformer will have all the context if necessary. 

cc @hermanschaaf @candiduslynx

codegen/golang.go
yevgenypats: Good catch. I'll bring the override behaviour back.

yevgenypats: done

codegen/table.go
yevgenypats: Actually second though I removed the implicit override behaviour and made `DefaultTypeTransformer` public so the user can decide for example on the order and so on. this also make the sdk code cleaner and more explicit.

codegen/golang.go
yevgenypats: Actually second thought I removed the implicit override behaviour  and made `DefaultTypeTransformer` public so the user can decide for example on the order and so on. this also make the sdk code cleaner and more explicit.

codegen/table.go
disq: using `any` enables easier and more confident implementation by the developers, as they can just do a type-check (or type switch) and handle accordingly. Otherwise they'll need to learn about the `reflect` package.

hermanschaaf: I would still vote for `StructField`, in order to provide all the necessary context, even if it means having to learn about the `reflect` package. Otherwise, we're bound to run into cases where we need that context but don't have it. (I'm thinking especially about cases where the name of the field indicates something, like an IP or a date, so you want to override it, but the type of the field is string)

yevgenypats: Im in with @hermanschaaf here. I think developers that uses codegen will anyhow need to learn about reflect either way. Will update and request re-review.

candiduslynx (approved):

codegen/golang.go
```diff
@@ -81,23 +86,37 @@ func WithUnwrapFieldsStructs(fields []string) TableOptions {
...
 	}
 }
 
-func DefaultTransformer(field reflect.StructField) string {
+// WithLogger
+func WithLogger(logger zerolog.Logger) TableOptions {
+	return func(t *TableDefinition) {
+		t.logger = logger
+	}
+}
```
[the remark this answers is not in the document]
yevgenypats: Correct. It's just for future use so we wont use fmt.Println

codegen/golang.go
```diff
@@ -81,23 +86,37 @@ func WithUnwrapFieldsStructs(fields []string) TableOptions {
...
+
+// WithValueTypeTransformer sets a function that can override the schema type for specific fields. Return `schema.TypeInvalid` to fall back to default behavior.
```
hermanschaaf, line 103: I think this comment is now incorrect: does it still fall back to default behavior if you return `schema.TypeInvalid`?

codegen/golang.go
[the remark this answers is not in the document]
candiduslynx: `WithLogger` is an option, but it's really interesting not to see the usage, I agree.
@yevgenypats

codegen/golang.go
candiduslynx: Yes, it fallbacks, so should state so

yevgenypats: I think yes. as @candiduslynx requested it. I added the fallback behaviour back.

codegen/table.go
candiduslynx: I finally agree that `refrect.StructField` allows for more interesting use-cases, so let's stick with it

codegen/golang.go
hermanschaaf: Oh sorry I see, it does yeah.

hermanschaaf (approved):

codegen/golang.go
yevgenypats: I think we can ad basically you can have either fallback or let's say you want to use `Default` but in some different order 🤷  I would keep both options in that case (even though it's additional external API).

erezrokah (approved):

PR #162, opened 2022-09-21
chore(codegen): Restructure `codegen` package

candiduslynx:


codegen/options.go
```diff
@@ -0,0 +1,54 @@
...
+
+// WithExtraColumns allows passing additional ColumnDefinitions
+func WithExtraColumns(columns ColumnDefinitions) TableOptions {
+	return func(t *TableDefinition) {
+		t.extraColumns = columns
+	}
+}
+
+// WithUnwrapFieldsStructs allows to unwrap specific struct fields (1 level deep only)
+func WithUnwrapFieldsStructs(fields []string) TableOptions {
```
hermanschaaf, line 29: I think this slipped through previous review, but shouldn't this be `WithUnwrapStructFields`? 🤔

yevgenypats (approved):
git didn't do a great job showing the diff but given it's just moving stuff around seems ok :)

codegen/options.go
candiduslynx: Will fix in the next PR

PR #163, opened 2022-09-21
feat: Ignore specific field types

erezrokah:
#### Summary

https://github.com/cloudquery/plugin-sdk/pull/157 changed the behavior of printing a log message on unsupported types (like `interface`) to failing.
~~I'm not sure that's what we want as now the consumer has to skip all the interface fields to make code gen work.~~
~~I kept the failure on custom type transformer as I believe that should be indeed a failure to generate.~~

Changed the code to explicitly specific types

---

- [x] Update or add tests 🧪

### DIFF
--- a/codegen/table_test.go
+++ b/codegen/table_test.go
@@ -33,6 +33,7 @@ type testStruct struct {
 	JSONTAG        *string    `json:"json_tag"`
 	SKIPJSONTAG    *string    `json:"-"`
 	NOJSONTAG      *string
+	InterfaceCol   interface{}
 
 	*embeddedStruct
 }
--- a/codegen/table.go
+++ b/codegen/table.go
@@ -73,11 +73,21 @@ func (t *TableDefinition) getUnwrappedFields(field reflect.StructField) []reflec
 	return fields
 }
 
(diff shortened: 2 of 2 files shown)


yevgenypats:
why not fail here? seems like it the last fallback and it still generates an error. maybe you can skip this explicitly? otherwise if we skip errors it will become really hard to about any issues in codegen as it will turn into log parsing race

erezrokah:
> why not fail here? seems like it the last fallback and it still generates an error. maybe you can skip this explicitly? otherwise if we skip errors it will become really hard to about any issues in codegen as it will turn into log parsing race

We're discussing this on Slack. It's hard to know in advance which fields to skip without implementing the reflection logic in consumers. Not sure we should be creating columns for fields that are interfaces.

codegen/table_test.go
```diff
@@ -33,6 +33,7 @@ type testStruct struct {
 	JSONTAG        *string    `json:"json_tag"`
 	SKIPJSONTAG    *string    `json:"-"`
 	NOJSONTAG      *string
+	InterfaceCol   interface{}
```
erezrokah, line 36: Without the change in this PR this cause the code gen to error out

PR #164, opened 2022-09-21
feat(codegen): Add `WithResolverTransformer` option

candiduslynx:
Add `WithResolverTransformer` option (based on #150)

### DIFF
--- a/codegen/options.go
+++ b/codegen/options.go
@@ -59,6 +59,8 @@ func WithTypeTransformer(transformer TypeTransformer) TableOption {
 
 // WithResolverTransformer sets a function that can override the resolver for a field.
 // DefaultResolverTransformer is used as the default.
+// If the transformer provided returns error, gen will fail.
+// To fallback onto the default resolver return "", nil.
 func WithResolverTransformer(transformer ResolverTransformer) TableOption {
 	return func(t *TableDefinition) {
 		t.resolverTransformer = transformer
--- a/codegen/options.go
+++ b/codegen/options.go
@@ -6,13 +6,6 @@ import (
 
 type TableOption func(*TableDefinition)
 
-// WithNameTransformer overrides how column name will be determined.
-func WithNameTransformer(transformer NameTransformer) TableOption {
-	return func(t *TableDefinition) {
-		t.nameTransformer = transformer
-	}
-}
-
 // WithSkipFields allows to specify what struct fields should be skipped.
 func WithSkipFields(fields []string) TableOption {
 	return func(t *TableDefinition) {
@@ -48,9 +41,26 @@ func WithLogger(logger zerolog.Logger) TableOption {
 	}
 }
 
+// WithNameTransformer overrides how column name will be determined.
+// DefaultNameTransformer is used as the default.
+func WithNameTransformer(transformer NameTransformer) TableOption {
+	return func(t *TableDefinition) {
+		t.nameTransformer = transformer
+	}
+}
+
 // WithTypeTransformer sets a function that can override the schema type for specific fields. Return `schema.TypeInvalid` to fall back to default behavior.
+// DefaultTypeTransformer is used as the default.
 func WithTypeTransformer(transformer TypeTransformer) TableOption {
 	return func(t *TableDefinition) {
 		t.typeTransformer = transformer
+	}
+}
+
+// WithResolverTransformer sets a function that can override the resolver for a field.
+// DefaultResolverTransformer is used as the default.
+func WithResolverTransformer(transformer ResolverTransformer) TableOption {
+	return func(t *TableDefinition) {
+		t.resolverTransformer = transformer
 	}
 }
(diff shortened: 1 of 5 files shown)


codegen/options.go
```diff
@@ -1,54 +1,66 @@
...
+func WithTypeTransformer(transformer TypeTransformer) TableOption {
 	return func(t *TableDefinition) {
 		t.typeTransformer = transformer
 	}
 }
+
+// WithTypeResolver allows specifying a custom resolver to generate for a specific schema.ValueType.
+// Value should name a schema.ColumnResolverProvider func.
+// Can be specified several times.
+func WithTypeResolver(vt schema.ValueType, resolver string) TableOption {
```
hermanschaaf, line 62: I think, when used in conjunction with `WithTypeTransformer`, this will not be flexible enough for some cases, and we need to carefully consider this function signature.

As an example (that I've really seen, but worked around), imagine the struct has a field using a custom timestamp type, called `MyTime`. We can use `WithTypeTransformer` to transform the column type to a normal `time.Time` type. However, we can't directly inject `MyTime` into the column: we need a resolver to convert it from `MyTime` to `time.Time` first.  Now, with the current way `WithTypeResolver` is defined, we would have to generate resolvers for *all* timestamp-ValueType columns, not just `MyTime` ones.

Maybe this is okay, since you could do some type assertions in the resolver to distinguish `time.Time` from `MyTime`? But it seems like there should be a way to define this function so that you can generate a resolver for specific struct fields, rather than specific ValueTypes.

I have no idea if this would be possible to implement, but maybe a signature like this could work:

```
type FieldResolver func(reflect.StructField) (string, error)

func WithFieldResolver(resolver TypeResolver) TableOption {
...
```

WDYT?

yevgenypats (changes requested):
Same request :) 

Very hard to review major breaking features for public APIs with refactor code that moves stuff around.

candiduslynx:
I'll rebase the code once #166 is merged, so there'll be only appropriate changes present in the PR.

codegen/options.go
hermanschaaf: Right, okay, so you'd have to do a type switch / assertion in the resolver.

I have another toy example I'd like to ask a question about:

```
type MyStruct struct {
    Name string
    Start time.Time
    End time.Time
    Date string
}
```

`Date` is a kind of time, so we would like to use `WithTypeTransformer` to transform it to `schema.TypeTimestamp`. We can look at the name of the field to identify `Date` and do that transformation. Now we also need to parse the date in a resolver before insertion into the column.

With `codegen.WithTypeResolver(schema.TypeTimestamp, "TimestampPathResolver")`, will we be forced to use the same resolver for `Start` and `End` as for `Date`? What if we need a different resolver for `Start` and `End`, or only for `End`?  Can we override it in that case?

yevgenypats:
Let's discuss live today but from quick glance seems to me we can keep our current convention and add `WithPathResolverTransformer` and have the same signature (for `WithNameTransformer` and `WithTypeTransformer`) that get's struct field and returns a `string, error`

hermanschaaf:
I'm wondering whether we need to add `WithPathResolverTransformer` to the SDK, increasing its surface area. In theory, it should be possible to call `codegen.NewTableFromStruct`, then loop over the returned columns, updating the resolver to what you need it to be for special cases. 

To be fair though, we probably could have done that for all the other transformers as well, so maybe this is fine since it will make plugin code more standardized.

codegen/options.go
candiduslynx: replaced by `ResolverTransformer` approach

codegen/transformers.go
```diff
@@ -66,3 +66,13 @@ func defaultGoTypeToSchemaType(v reflect.Type) (schema.ValueType, error) {
 		return schema.TypeInvalid, fmt.Errorf("unsupported type: %s", k)
 	}
 }
+
+type ResolverTransformer func(field reflect.StructField, path string) string
+
+func DefaultResolverTransformer(_ reflect.StructField, path string) string {
```
yevgenypats, line 72: Should it also return an `error` to be consistent with all other APIs and not to break compatibility if we need to add it in the future.

hermanschaaf (approved):
LGTM 👍

codegen/transformers.go
yevgenypats: Also, for the same reason of having max context and no introducing more options and/or breaking changes it should be also `func(field reflect.StructField, parent reflect.StructField) string, error`.

yevgenypats (changes requested):
Left some comments but looks good I think. cc @hermanschaaf @candiduslynx

codegen/transformers.go
candiduslynx: I don't think it should have info about parent, as the only thing that matters is type info + path where you can get it.

yevgenypats: but that path is basically parent.name + child.name ?

yevgenypats (approved):

PR #166, opened 2022-09-22
chore(codegen): Move code for better coupling

candiduslynx:


codegen/util.go
```diff
@@ -14,3 +14,15 @@ func isFieldStruct(reflectType reflect.Type) bool {
 		return false
 	}
 }
+
+func isTypeIgnored(t reflect.Type) bool {
```
erezrokah, line 18: Since these utils are only used in a single place, how about keeping them close to where they are used?

`util` files tend to get bloated with just about everything with time, so unless we re-use these functions might be ok to keep them closer to their usage, WDYT?

erezrokah (approved):
This looks good to me, but I also found the previous code ok to read

codegen/util.go
candiduslynx: I think it'll be an issue if we have 10+ funcs in this file.

PR #169, opened 2022-09-22
feat: Addition and hopefully soon removal of gen command

yevgenypats:
While I was  working on fixing bugs in the gen command, moving it to the SDK (from the CLI after moving it from from the SDK 2 two weeks ago :) , protocol changes, tests, templates), adding flags to the gen command (testing the gen command)  I realised once again that we are slowly becoming a yaml automation company (https://xkcd.com/1319/) and now we have to make sure the "example config" that our baked cli example config should be up to date with the non-existent documentation (which is actually more important then the baked gen command). 

I believe we should drop the gen command as there is very little difference or time saving between the following:

```
cloudquery gen source aws --output example/config.yml
cloudquery gen destination postgresql --output example/config
```

vs

create those two files

```
kind: source
name: aws
```

```
kind: destination
name: postgresql
```

I believe we should stop maintaing this command before releasing v2 pre-release and maintain this public api as we can put our effort in better places by providing up-to-date documentation and consistent behaviour. 

Also, a bit more context on defaults:

if plugins have defaults this can work with the above small files as we shouldn't create defaults by saving those into a string and then outputing them back into a file and reading them back to the spec :)

### DIFF
--- a/specs/destination.go
+++ b/specs/destination.go
@@ -2,10 +2,16 @@ package specs
 
 import (
 	"bytes"
+	_ "embed"
 	"encoding/json"
 	"fmt"
+	"io"
 	"strings"
+	"text/template"
 )
+
+//go:embed templates/destination.go.tpl
+var destinationExampleTemplate string
 
 type WriteMode int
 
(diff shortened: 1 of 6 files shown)


hermanschaaf:
Agree, let's drop the `gen` command before v2 release. This will force us to make sure each plugin has good (and up-to-date) examples in the docs that you can copy-paste. And we'd only need to maintain those in one place, which is a good thing.

hermanschaaf:
So if all agree then this PR will be closed, right?

PR #174, opened 2022-09-25
fix!: Make destinations work over gRPC only

yevgenypats:
This should go with https://github.com/cloudquery/cloudquery/pull/2043


erezrokah (approved):

hermanschaaf:
Nice, makes things much simpler 👍

PR #181, opened 2022-09-27
fix: Spec unmarshalling now supports defaults and validation

yevgenypats:
Fixes: defaults, validation and unknown keys


specs/destination_test.go
```diff
@@ -54,3 +40,137 @@ func TestDestinationUnmarshalSpec(t *testing.T) {
...
+	name   string
+	spec   string
+	err    string
+	source *Source
+}{
+	{
+		"invalid_kind",
+		`kind: nice`,
+		"failed to decode spec: unknown kind nice",
+		nil,
```
hermanschaaf, line 54: nit: let's add keys to these struct instantiations

specs/source.go
```diff
@@ -54,12 +54,24 @@ func (s *Source) UnmarshalSpec(out interface{}) error {
...
+		return fmt.Errorf("name is required")
+	}
+	if s.Version == "" {
+		return fmt.Errorf("version is required")
+	}
+	if !strings.HasPrefix(s.Version, "v") {
+		return fmt.Errorf("version must start with v")
+	}
+	if len(s.Destinations) == 0 {
+		return fmt.Errorf("at leats one destination is required")
```
hermanschaaf, line 74:
```suggestion
		return fmt.Errorf("at least one destination is required")
```

specs/source.go
```diff
@@ -54,12 +54,24 @@ func (s *Source) UnmarshalSpec(out interface{}) error {
...
-	return nil, nil
+func (s *Source) Validate() error {
+	if s.Name == "" {
+		return fmt.Errorf("name is required")
+	}
+	if s.Version == "" {
+		return fmt.Errorf("version is required")
+	}
+	if !strings.HasPrefix(s.Version, "v") {
+		return fmt.Errorf("version must start with v")
```
hermanschaaf, line 71: Do we not support `latest` version at all anymore?

yevgenypats:
Closes https://github.com/cloudquery/cloudquery/issues/2054

specs/spec_reader_test.go
```diff
@@ -2,46 +2,17 @@ package specs
...
-
-func TestNoSpecs(t *testing.T) {
-	_, err := NewSpecReader("testdata")
-	require.Equal(t, err.Error(), "no valid config files found in directory testdata")
+	if len(specReader.sources) != 1 {
+		t.Fatalf("expected 1 source got %d", len(specReader.sources))
+	}
+	if len(specReader.destinations) != 1 {
+		t.Fatalf("expected 1 destination got %d", len(specReader.destinations))
+	}
```
hermanschaaf, line 17: nit: Go convention is to have the test failure order mirror the comparison above it, so it would be `got %d sources, want %d` in this case (See https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures)

specs/source_test.go
```diff
@@ -1,44 +1,154 @@
...
+		nil,
+	},
+	{
+		"destination_required",
+		`kind: source
+spec:
+  name: test
+  version: v1.1.0
+`,
+		"at leats one destination is required",
```
hermanschaaf, line 107:
```suggestion
		"at least one destination is required",
```

hermanschaaf (changes requested):

specs/destination_test.go
yevgenypats: I feel for this test driven use-case it reduces the boilerplate imo and makes it easier to add new tests. just opinion ofc.

hermanschaaf: Makes sense, though as someone reading it for the first time, I found it hard to know what all these parameters were for, and had to scroll up and down a few times. I think it improves readability if you have the keys right there

hermanschaaf (approved):

specs/destination.go
```diff
@@ -43,10 +40,23 @@ func (d *Destination) UnmarshalSpec(out interface{}) error {
...
+}
+
+func (d *Destination) Validate() error {
+	if d.Name == "" {
+		return fmt.Errorf("name is required")
+	}
+	if d.Version == "" {
+		return fmt.Errorf("version is required")
+	}
+	if !strings.HasPrefix(d.Version, "v") {
```
erezrokah, line 56: Nit, we should do full SemVer validation as the following string `vv1.0.0` passes this validation

specs/destination.go
```diff
@@ -43,10 +40,23 @@ func (d *Destination) UnmarshalSpec(out interface{}) error {
...
+		return fmt.Errorf("version must start with v")
```
erezrokah, line 57: Then change this error message to `version must be a valid SemVer string`)

PR #183, opened 2022-09-27
feat!: SpecReader to support multiple files, dirs and yaml

yevgenypats:
This was a requested feature by @bbernays to take a similar approach [kubectl appl](https://kubernetes.io/docs/reference/kubectl/cheatsheet/#kubectl-apply)  so it can accept multiple files, directories and files with multiple yamls.

This was observed as a use-case from multiple users.

Also this introduce environment substitution in configs.


specs/spec_reader.go
```diff
@@ -8,78 +8,93 @@ import (
...
-		if !file.IsDir() && !strings.HasPrefix(file.Name(), ".") && strings.HasSuffix(file.Name(), ".yml") {
-			data, err := os.ReadFile(filepath.Join(directory, file.Name()))
-			if err != nil {
-				return nil, fmt.Errorf("failed to read file %s: %w", file.Name(), err)
-			}
-			var s Spec
-			if err := SpecUnmarshalYamlStrict(data, &s); err != nil {
-				return nil, fmt.Errorf("failed to unmarshal file %s: %w", file.Name(), err)
+	// support multiple yamls in one file
+	for i, doc := range strings.Split(string(data), "---") {
```
hermanschaaf, line 23: Bit of an edge case to be sure, but if some string in the config happened to contain `---`, that would cause the file to be broken up incorrectly. Might be better to split by lines first, then look for a line that only contains `---`

yevgenypats: fixed

specs/spec_reader.go
```diff
@@ -8,78 +8,92 @@ import (
...
+		} else {
+			if err := reader.loadSpecsFromFile(path); err != nil {
+				return nil, err
+			}
 		}
 	}
-	return nil
+
+	if len(reader.Sources) == 0 {
+		return nil, fmt.Errorf("expecting at least one source in: %v ", paths)
```
hermanschaaf, line 91:
```suggestion
		return nil, fmt.Errorf("expecting at least one source in: %v", paths)
```

hermanschaaf (approved):

PR #187, opened 2022-09-28
fix: Streaming to destination plugin wasn't implemented correctly

yevgenypats:
We need to call CloseAndRecv on the client side and CloseAndSend on Server side for everything to close gracefully.

This should help with places where data wasn't written because the connection was closed before everything was written to database.

https://github.com/grpc/grpc-go/blob/master/examples/route_guide/client/client.go


hermanschaaf (approved):
Nice find 👍

PR #188, opened 2022-09-28
fix: Validate versions only for github registry

shimonp21:
#### Summary

---

- [ ] Update or add tests 🧪


hermanschaaf (approved):
Makes sense, even though there is a workaround I think this will improve developer experience a little.

erezrokah (approved):

yevgenypats:
@shimonp21 can you also add a test to that so we can catch regressions.

PR #191, opened 2022-09-28
fix: Add initialisms for k8s

shimonp21:
#### Summary

---

- [ ] Update or add tests 🧪


disq (approved):

yevgenypats (changes requested):
Seems like we shouldn't have it here but rather have an option to pass it from the plugins. otherwise the sdk versions will blow up and/or it will introduce behaviour to plugins.

yevgenypats (approved):
I think it's fine but I do think we need to remove this global api and produce a case object with standard go options, defaults so it adhres to go best practices and all our other public APIS.

PR #205, opened 2022-09-30
feat(schema): Add schema.TypeTimeInterval

candiduslynx:
#### Summary

Add duration column type for #178 

---

- [x] Update or add tests 🧪

### DIFF
--- a/codegen/table_test.go
+++ b/codegen/table_test.go
@@ -3,12 +3,12 @@ package codegen
 import (
 	"bytes"
 	"fmt"
-	"github.com/cloudquery/plugin-sdk/caser"
 	"reflect"
 	"strings"
 	"testing"
 	"time"
 
+	"github.com/cloudquery/plugin-sdk/caser"
 	"github.com/cloudquery/plugin-sdk/schema"
 	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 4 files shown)


yevgenypats (changes requested):
Looks good. One question - maybe we should call it interval? i.e stick to postgresql types? or at least be consistent with one db? as in the json when it is being sent over the wire it is string anyway.

hermanschaaf (approved):
LGTM :+1: 

Let's mark it as a breaking change though, because destination plugins will need to explicitly handle this new type.

yevgenypats:
> LGTM 👍
> 
> Let's mark it as a breaking change though, because destination plugins will need to explicitly handle this new type.

I think for destination plugin we need to always default to text as this field can contain anything and we shouldn't break destinations every-time we support new field in CQ.

So I wouldn't mark it as breaking even it will break:) maybe just follow-up PR with the right behaviour to default to text?

candiduslynx:
Now it'll be treated as `int64` -> `schema.TypeInt`, so IDK if that's a good idea to convert this to text.

As for the interval - will do

yevgenypats (approved):

PR #206, opened 2022-09-30
feat: Move NewSourceClientSpawn (and dest) to SDK

yevgenypats:
#### Summary

This removes the need of an additional client that contains another client in the CLI.  After this is merged a PR to the CLI will come shortly after. 

Also I made the http download request now get the context which should help with cancellation


clients/destination.go
```diff
@@ -21,6 +40,59 @@ func NewDestinationClient(cc grpc.ClientConnInterface) *DestinationClient {
 	}
 }
 
+// NewSourceClientSpawn starts a new source plugin process, connects to it via gRPC server
```
hermanschaaf, line 43:
```suggestion
// NewDestinationClientSpawn starts a new destination plugin process, connects to it via gRPC server
```

clients/destination.go
```diff
@@ -21,6 +40,59 @@ func NewDestinationClient(cc grpc.ClientConnInterface) *DestinationClient {
...
+// and returns a new SourceClient
```
hermanschaaf, line 44:
```suggestion
// and returns a new DestinationClient
```

clients/destination.go
```diff
@@ -1,18 +1,37 @@
...
+
+type DestinationClientOption func(*DestinationClient)
+
+func WithDestinationLogger(logger zerolog.Logger) func(*DestinationClient) {
+	return func(c *DestinationClient) {
+		c.logger = logger
+	}
 }
 
 func NewDestinationClient(cc grpc.ClientConnInterface) *DestinationClient {
```
hermanschaaf, line 37: Should this function maybe also accept the new options?

yevgenypats:
Also, removed the dependency on a UI component directly and used writer interface.

clients/destination.go
```diff
@@ -21,6 +40,59 @@ func NewDestinationClient(cc grpc.ClientConnInterface) *DestinationClient {
 	}
 }
 
+// NewSourceClientSpawn starts a new source plugin process, connects to it via gRPC server
+// and returns a new SourceClient
+func NewDestinationClientSpawn(ctx context.Context, path string, opts ...DestinationClientOption) (*DestinationClient, error) {
```
hermanschaaf, line 45:
```suggestion
func NewDestinationClient(ctx context.Context, path string, opts ...DestinationClientOption) (*DestinationClient, error) {
```

Naming is hard, but I'm not sure the current name is a great description of what the function does... Maybe `NewDestinationClientFromPath`? With further clarification in the comment, I think it should be fairly clear.

clients/destination.go
```diff
@@ -85,3 +157,27 @@ func (c *DestinationClient) Write(ctx context.Context, resources <-chan []byte)
 
 	return res.FailedWrites, nil
 }
+
+// Close is used only in conjuncetion with NewDestinationPluginSpawn
+// it closes the connection it created, kills the spawned process and removes the socket file
+func (c *DestinationClient) Close() error {
+	if c.grpcSocketName != "" {
```
hermanschaaf, line 164:
```suggestion
// Close is used only in conjunction with NewDestinationPluginSpawn.
// It closes the connection it created, kills the spawned process and removes the socket file.
func (c *DestinationClient) Close() error {
	if c.grpcSocketName != "" {
```

clients/destination.go
hermanschaaf: Or maybe `SpawnNewDestinationClient` 🤔

clients/source.go
```diff
@@ -80,3 +153,27 @@ func (c *SourceClient) Sync(ctx context.Context, spec specs.Source, res chan<- [
 		res <- r.Resource
 	}
 }
+
+// Close is used only in conjuncetion with NewSourcePluginSpawn
+// it closes the connection it created, kills the spawned process and removes the socket file
+func (c *SourceClient) Close() error {
```
hermanschaaf, line 159:
```suggestion
// Close is used only in conjunetion with NewSourcePluginSpawn.
// It closes the connection it created, kills the spawned process and removes the socket file.
func (c *SourceClient) Close() error {
```

hermanschaaf (changes requested):
Looks good, I just commented on a few minor things we may want to discuss or change before merging

clients/destination.go
yevgenypats: On the fence here, maybe in the future (in anycase Im less worried about it because it is mainly used by the sdk). specifaclly the log has not impact if the server is already running

clients/destination.go
yevgenypats: Im good with either, I thought about all three and this one seemed to me the most consistent but Im good with either

hermanschaaf: What about `NewManagedDestinationClient`, because it's managing the underlying process for you?

clients/destination.go
hermanschaaf: Ok yeah, let's leave it 👍

clients/destination.go
```diff
@@ -85,3 +157,27 @@ func (c *DestinationClient) Write(ctx context.Context, resources <-chan []byte)
 
 	return res.FailedWrites, nil
 }
+
+// Close is used only in conjunction with NewDestinationPluginSpawn.
```
hermanschaaf, line 161:
```suggestion
// Close is used only in conjunction with NewManagedDestinationClient.
```

hermanschaaf (approved):

PR #208, opened 2022-09-30
feat: Make NewSourceClient (and dest) as one interface

yevgenypats:
#### Summary

I think I finally nailed the Client interface. This will remove completely the need for "PluginManager" in the CLI and yet another abstraction layer and prop drilling with WithOptions duplicates for every abstraction layer.

Also, the SDK is aware of `registry`, `path` and `version` so I really think it should be here. it turns the CLI intro real CLI finally.

### DIFF
--- a/clients/destination.go
+++ b/clients/destination.go
@@ -7,18 +7,20 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"path/filepath"
+	"strings"
 
 	"github.com/cloudquery/plugin-sdk/internal/pb"
 	"github.com/cloudquery/plugin-sdk/schema"
 	"github.com/cloudquery/plugin-sdk/specs"
 	"github.com/rs/zerolog"
-	"github.com/rs/zerolog/log"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
 )
 
 type DestinationClient struct {
 	pbClient       pb.DestinationClient
+	directory      string
 	cmd            *exec.Cmd
 	logger         zerolog.Logger
 	conn           *grpc.ClientConn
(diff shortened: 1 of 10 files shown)


clients/destination.go
```diff
@@ -34,21 +36,48 @@ func WithDestinationLogger(logger zerolog.Logger) func(*DestinationClient) {
 	}
 }
 
-func NewDestinationClient(cc grpc.ClientConnInterface) *DestinationClient {
-	return &DestinationClient{
-		pbClient: pb.NewDestinationClient(cc),
+func WithDesintationDirectory(directory string) func(*DestinationClient) {
```
hermanschaaf, line 39:
```suggestion
func WithDestinationDirectory(directory string) func(*DestinationClient) {
```

clients/destination_test.go
```diff
@@ -31,11 +30,7 @@ func TestDestinationClient(t *testing.T) {
 	for _, tc := range newDestinationClientTestCases {
 		t.Run(tc.Path+"_"+tc.Version, func(t *testing.T) {
 			dirName := t.TempDir()
-			localPath := path.Join(dirName, "plugin")
-			if err := DownloadPluginFromGithub(ctx, localPath, tc.Path, tc.Version, PluginTypeDestination); err != nil {
-				t.Fatal(err)
-			}
-			c, err := NewManagedDestinationClient(ctx, localPath, WithDestinationLogger(l))
+			c, err := NewDestinationClient(ctx, tc.Registry, tc.Path, tc.Version, WithDestinationLogger(l), WithDesintationDirectory(dirName))
```
hermanschaaf, line 33:
```suggestion
			c, err := NewDestinationClient(ctx, tc.Registry, tc.Path, tc.Version, WithDestinationLogger(l), WithDestinationDirectory(dirName))
```

clients/source.go
```diff
@@ -40,22 +43,61 @@ func WithSourceLogger(logger zerolog.Logger) func(*SourceClient) {
...
+func WithSourceDirectory(directory string) func(*SourceClient) {
+	return func(c *SourceClient) {
+		c.directory = directory
 	}
 }
 
-// NewManagedSourceClient starts a new source plugin process, connects to it via gRPC server
-// and returns a new SourceClient
-func NewManagedSourceClient(ctx context.Context, path string, opts ...SourceClientOption) (*SourceClient, error) {
+func WithSourceGrpcConn(userConn *grpc.ClientConn) func(*SourceClient) {
```
hermanschaaf, line 52:
```suggestion
func WithSourceGRPCConnection(userConn *grpc.ClientConn) func(*SourceClient) {
```

hermanschaaf: I must be missing something: why don't destinations have this option?

serve/source_test.go
```diff
@@ -105,7 +105,10 @@ func TestServe(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Failed to dial bufnet: %v", err)
 	}
-	c := clients.NewSourceClient(conn)
+	c, err := clients.NewSourceClient(ctx, specs.RegistryGrpc, "", "", clients.WithSourceGrpcConn(conn))
```
hermanschaaf, line 108:
```suggestion
	c, err := clients.NewSourceClient(ctx, specs.RegistryGrpc, "", "", clients.WithSourceGRPCConnection(conn))
```

hermanschaaf (changes requested):

clients/source.go
yevgenypats: It wasn't really used, it is mainly used in tests when I wanted to create a special grpc conn which is in memory. prob don't have this test yet for destinations.

hermanschaaf (approved):

clients/source.go
yevgenypats: Adding a test as well?

PR #211, opened 2022-09-30
feat: Add Close() to destination interface and new writemode

hermanschaaf:
This is useful in cases where the destination plugin may need to close files or connections, or perform a final flush, when the sync is done.


yevgenypats (approved):
Nice. Also added new `writemode` `overwrite-deletestale` which will give destination plugins to delete stale data on close

PR #220, opened 2022-10-02
fix: Create doc directory if doesn't exist

erezrokah:
#### Summary

If you run the `doc` command and the output directory doesn't exist, the command will fail. This fixes it

---

- [ ] Update or add tests 🧪

### DIFF
--- a/plugins/source_docs.go
+++ b/plugins/source_docs.go
@@ -17,6 +17,10 @@ var templatesFS embed.FS
 // GenerateSourcePluginDocs creates table documentation for the source plugin based on its list of tables
 func (p *SourcePlugin) GenerateSourcePluginDocs(dir string) error {
 	for _, table := range p.Tables() {
+		if err := os.MkdirAll(dir, os.ModePerm); err != nil {
+			fmt.Printf("failed to create directory %q: %v", dir, err)
+			return err
+		}
 		if err := renderAllTables(table, dir); err != nil {
 			fmt.Printf("render table %s error: %s", table.Name, err)
 			return err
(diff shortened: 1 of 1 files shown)


plugins/source_docs.go
```diff
@@ -17,6 +17,10 @@ var templatesFS embed.FS
 // GenerateSourcePluginDocs creates table documentation for the source plugin based on its list of tables
 func (p *SourcePlugin) GenerateSourcePluginDocs(dir string) error {
 	for _, table := range p.Tables() {
+		if err := os.MkdirAll(dir, os.ModePerm); err != nil {
```
yevgenypats, line 20: I thinks this should be before the loop

yevgenypats: also, no need to print to stdout as it return an error anyway and it will print (otherwise it will be printed twice which will be confusing). same of the underlying line (I know not related to this PR)

yevgenypats (changes requested):

plugins/source_docs.go
erezrokah: > I thinks this should be before the loop

Good point, fixed in https://github.com/cloudquery/plugin-sdk/pull/220/commits/d6e2729021e1cf255bfa4705d5998e6bcf183e2c. I think it has the downside of creating the directory even if there are no tables, but I don't think that's a concern we should be dealing with.

erezrokah: >also, no need to print to stdout as it return an error anyway and it will print (otherwise it will be printed twice which will be confusing). same of the underlying line (I know not related to this PR)

Fixed in https://github.com/cloudquery/plugin-sdk/pull/226

yevgenypats (approved):

PR #223, opened 2022-10-02
fix: Use correct path for Windows zip

erezrokah:
#### Summary

Hard to test this as we need to release a new SDK, update it in the CLI and then test again via https://github.com/cloudquery/cloudquery/pull/2204.

Similar to the fix in https://github.com/cloudquery/cloudquery/pull/1894

---

- [ ] Update or add tests 🧪

### DIFF
--- a/clients/download.go
+++ b/clients/download.go
@@ -50,6 +50,9 @@ func DownloadPluginFromGithub(ctx context.Context, localPath string, org string,
 	if org != "cloudquery" {
 		pathInArchive = fmt.Sprintf("cq-%s-%s", typ, name)
 	}
+	if runtime.GOOS == "windows" {
+		pathInArchive += ".exe"
+	}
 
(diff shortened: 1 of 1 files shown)


yevgenypats (changes requested):
does `.exe` really needed on windows? Im not sure that the case especially for something that is not user faceing and it will make the code more complex because path will change depending on OS

erezrokah:
Either we fix this in the SDK, or we change our build configuration (if possible).
I think having binaries with `exe` extension is quite common, expected by users and is the default for Windows.
I'm not sure actually if the `exe` is added by `go build` or Go Releaser.
From a quick look at this guide https://www.digitalocean.com/community/tutorials/how-to-build-go-executables-for-multiple-platforms-on-ubuntu-16-04 it seems `exe` is added by `go` and required?
image

We can do some tests to see if we can drop it, but maybe we can merge the fix as currently the CLI is broken on Windows?

PR #224, opened 2022-10-02
feat!: Add overwrite-delete-stale mode for destination plugins

yevgenypats:
This introduce number of things:
- Major: Adds `overwrite-deletestale` mode
- Major: make destination plugin API consistent with source plugin api
- Minor: removes dep on snapshot testing library as we want to avoid this and use standard comparison and/or google diff library (see one of many articles - https://medium.com/@sapegin/whats-wrong-with-snapshot-tests-37fbe20dfe8e). Also had extensive experience with snapshot testing in React so def agree with the article/s

Related postgresql and CLI support - https://github.com/cloudquery/cloudquery/pull/2220


plugins/testdata/relation_table.md
```diff
@@ -9,6 +10,8 @@ This table depends on [`test_table`](test_table.md).
 ## Columns
 | Name          | Type          |
 | ------------- | ------------- |
-|string_col|String|
 |_cq_id (PK)|UUID|
-|_cq_fetch_time|Timestamp|
+|_cq_parent_id|UUID|
+|_cq_source_name|String|
```
hermanschaaf, line 15: I guess I'll get to it as I read on, but I'm wondering what `_cq_source_name` contains and why we need it

hermanschaaf (commented):
I'm not quite done with the review, but leaving some questions in the meantime

specs/destination.go
```diff
@@ -19,9 +19,9 @@ type Destination struct {
 }
 
 const (
-	WriteModeAppend WriteMode = iota
+	WriteModeOverwriteDeletestale WriteMode = iota
 	WriteModeOverwrite
```
erezrokah, line 23: Does `WriteModeOverwrite` translates to `upsert` (which is the current default)?

specs/destination.go
```diff
@@ -19,9 +19,9 @@ type Destination struct {
 }
 
 const (
-	WriteModeAppend WriteMode = iota
+	WriteModeOverwriteDeletestale WriteMode = iota
```
hermanschaaf, line 22:
```suggestion
	WriteModeOverwriteDeleteStale WriteMode = iota
```

erezrokah (changes requested):
Added a few comments, but I think a live overview of this feature will make the PR much easier to understand.

If I got this correctly we use `sync_time` and `source_name` to determine which resources to delete which makes sense.

If so and also [based on your comment](https://github.com/cloudquery/cloudquery/issues/2168#issuecomment-1264748875), `source_name` has to be unique across configuration files, correct?

Also, if someone renames a source the logic fails correct?

plugins/testdata/test_table.md
```diff
@@ -10,8 +11,10 @@ The following tables depend on `test_table`:
 ## Columns
 | Name          | Type          |
 | ------------- | ------------- |
+|_cq_id|UUID|
+|_cq_parent_id|UUID|
+|_cq_source_name|String|
+|_cq_sync_time|Timestamp|
```
hermanschaaf, line 17: Instead of `_cq_source_name` and `_cq_sync_time`, which will be the same for all inserted rows in a sync, I would like to suggest that we introduce a CloudQuery syncs table, for which an ID gets referenced here.

So for example, here we'd add one new column, `_cq_sync_id`. In another table, maybe called `cq_syncs`, we can list all past syncs with the following columns to start with:

```
id text (pk)
source_name text
start_time timestamp -- indexed

--- other ideas, not strictly required now:
end_time timestamp -- update at end of sync
status text -- updated when we transition from one state to another
errors uint
warnings uint
metadata jsonb -- plugins can use this to store cursors, for example
```

For delete-stale functionality, we wouldn't need to join or reference this table: we would only need to delete all rows with sync IDs that are not the current sync's ID.

Users who wish to delete stale records on their own can query this table to get the latest sync ID and then delete all rows with a different sync id. If we expand it with some more data (some ideas listed above), this table could also give users a great place to start monitoring their syncs without needing to parse CLI output or logs.

In any case, even if we keep it minimal for now, such a table will allow us to expand our metadata in the future without polluting all tables with new columns.

This table could also be useful for future functionality, such as cursors, where source plugins need to store some information for retrieval on the next sync. But that can also be designed differently.

internal/pb/source.proto
```diff
@@ -11,20 +11,33 @@ service Source {
   rpc GetVersion(GetVersion.Request) returns (GetVersion.Response);
   // Get all tables the source plugin supports
   rpc GetTables(GetTables.Request) returns (GetTables.Response);
+  // GetSyncSummary returns the latest sync summary of the source plugin. we don't want to send the summary on
+  // every sync request.
```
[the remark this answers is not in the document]
yevgenypats: that can work

specs/destination.go
```diff
@@ -19,9 +19,9 @@ type Destination struct {
...
 	WriteModeOverwrite
-	WriteModeOverwriteDeletestale
+	WriteModeAppend
```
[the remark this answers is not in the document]
yevgenypats: it is already implemented in postgresql and I gave it a manual test as well

plugins/testdata/test_table.md
yevgenypats: I would avoid for now the need for additional table and foreign keys. I think it is quite common in datalakes to have duplicates in favour of simplicity.

So basically simple implementation first -> operational experience -> more complex one later. 

we can still add later on a table with all sync times but I would still avoid the foreign keys as they are also not supported in all databases. but I don't think this a blocker and we can always add this table later.

plugins/source_docs_test.go
```diff
@@ -58,13 +73,16 @@ func TestGenerateSourcePluginDocs(t *testing.T) {
 		t.Fatalf("unexpected error calling GenerateSourcePluginDocs: %v", err)
 	}
 
-	expectFiles := []string{"test_table.md", "relation_table.md", "README.md"}
-	for _, exp := range expectFiles {
-		t.Run(exp, func(t *testing.T) {
-			output := path.Join(tmpdir, exp)
+	for filename, content := range expectedDocFiles {
```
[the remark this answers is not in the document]
yevgenypats: sounds good, I'll move it out of this PR.

specs/destination.go
```diff
@@ -19,9 +19,9 @@ type Destination struct {
...
 )
 
 func (d *Destination) SetDefaults() {
```
hermanschaaf, line 27: Github doesn't allow me to comment on that line, but on [line 68](https://github.com/cloudquery/plugin-sdk/pull/224/files#diff-7c146e9fc20b6cf14b74be16d0cece21f950b510aa75b4d92c9527cfd1f7a8f1R68), let's change it to `overwrite-delete-stale` (with a dash)

plugins/testdata/test_table.md
hermanschaaf: Agree, makes sense to keep things simple at first, and data lakes often do denormalize for the sake of simplicity. I don't think a database needs to strictly support foreign keys for this secondary table to be useful though; you can always look up the latest ID and then run further queries using it (accepting the risk in this case that operations won't necessarily be atomic, again depending on the database).

For me the important thing is having a handle to a unique sync id with which you can reference further information, should you need it. If we think that's something we'll add later, to maintain backwards-compatibility we will then have 5 `_cq` columns instead of 3:

```
|_cq_id|UUID|
|_cq_parent_id|UUID|
|_cq_source_name|String|
|_cq_sync_time|Timestamp|
|_cq_sync_id|UUID|
```

I think that's okay, and if you think it's okay we can do it, but worth noting that we have the chance now to keep the number of internal columns to their theoretical minimum.

Or the other option, down the line, would be to use `sync_time` and `source_name` as a combined PK, if we want to keep it to 4 internal columns but make joins a bit harder.

specs/destination.go
yevgenypats: done

plugins/testdata/test_table.md
yevgenypats: I think `_cq_sync_time` (_source_name) should be good enough for now for uniqueness and I wouldn't optimize yet for small amount of columns it doesn't really cost us anything atm. so would suggest to look into that down the line when/if it will become an option.

internal/pb/source.proto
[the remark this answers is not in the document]
yevgenypats: let's do this in a follow-up PR though as it's already pretty big

erezrokah (approved):
I think we can merge the initial PR and iterate on it with follow up PRs

PR #231, opened 2022-10-03
fix: Use correct binary path on Windows

erezrokah:
#### Summary

Based on `go` source code binaries on Windows must end in `exe`, `cmd`, etc. See https://github.com/golang/go/blob/f6d844510d5f1e3b3098eba255d9b633d45eac3b/src/os/exec/lp_windows.go#L39
⬆️ Is the code that looks up the binary when you use the `exec` package

---

- [ ] Update or add tests 🧪

### DIFF
--- a/clients/destination.go
+++ b/clients/destination.go
@@ -88,6 +88,7 @@ func NewDestinationClient(ctx context.Context, registry specs.Registry, path str
 		}
 		org, name := pathSplit[0], pathSplit[1]
 		localPath := filepath.Join(c.directory, "plugins", string(PluginTypeDestination), org, name, version, "plugin")
+		localPath = withBinarySuffix(localPath)
 		if err := DownloadPluginFromGithub(ctx, localPath, org, name, version, PluginTypeDestination, c.writers...); err != nil {
 			return nil, err
 		}
--- a/clients/download.go
+++ b/clients/download.go
@@ -50,9 +50,7 @@ func DownloadPluginFromGithub(ctx context.Context, localPath string, org string,
 	if org != "cloudquery" {
 		pathInArchive = fmt.Sprintf("cq-%s-%s", typ, name)
 	}
-	if runtime.GOOS == "windows" {
-		pathInArchive += ".exe"
-	}
+	pathInArchive = withBinarySuffix(pathInArchive)
 
 	fileInArchive, err := archive.Open(pathInArchive)
 	if err != nil {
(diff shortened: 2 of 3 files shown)


hermanschaaf (approved):

clients/destination.go
```diff
@@ -88,6 +88,7 @@ func NewDestinationClient(ctx context.Context, registry specs.Registry, path str
 		}
 		org, name := pathSplit[0], pathSplit[1]
 		localPath := filepath.Join(c.directory, "plugins", string(PluginTypeDestination), org, name, version, "plugin")
+		localPath = withBinarySuffix(localPath)
```
yevgenypats, line 91: I don't think we should save it with `.exe` (I get the goreleaser part but no need to do it on the save part as it will work without the exe)

clients/source.go
```diff
@@ -94,6 +94,7 @@ func NewSourceClient(ctx context.Context, registry specs.Registry, path string,
 		}
 		org, name := pathSplit[0], pathSplit[1]
 		localPath := filepath.Join(c.directory, "plugins", string(PluginTypeSource), org, name, version, "plugin")
+		localPath = withBinarySuffix(localPath)
```
yevgenypats, line 97: this wont be needed.

clients/download.go
```diff
@@ -50,9 +50,7 @@ func DownloadPluginFromGithub(ctx context.Context, localPath string, org string,
 	if org != "cloudquery" {
 		pathInArchive = fmt.Sprintf("cq-%s-%s", typ, name)
 	}
-	if runtime.GOOS == "windows" {
-		pathInArchive += ".exe"
-	}
+	pathInArchive = withBinarySuffix(pathInArchive)
```
yevgenypats, line 53: I think we just needed it here so no need to do `withBinarySuffix`

yevgenypats (changes requested):
I think on the save side let's make it consistent without suffix

clients/destination.go
erezrokah: See my comment, it doesn't work without it as the `go` code looks for the extension:
https://github.com/golang/go/blob/f6d844510d5f1e3b3098eba255d9b633d45eac3b/src/os/exec/lp_windows.go#L39

See failure https://github.com/cloudquery/cloudquery/actions/runs/3173117211/jobs/5168369062#step:7:9
image

clients/download.go
erezrokah: We already have it here and it still fails (see failure in other comment)

yevgenypats (approved):

PR #234, opened 2022-10-03
fix: SDK compile error, and add workflow

yevgenypats:
#### Summary

Seems we need to add a build step as otherwise we ship passing test but broken version 🤔


yevgenypats:
Actually Im not even sure what happened now as seems that it does compile. my IDE went crazy apparently

erezrokah:
I think linting should fail if it doesn't compile

PR #237, opened 2022-10-04
fix: Error on incorrect table configuration

hermanschaaf:
This adds checks for a whole range of misconfigurations that could happen in the tables config.

Some of the main cases handled:
 - invalid table name
 - wildcard in skip_tables section
 - child table specified without its parent table
 - list of tables is empty

Closes https://github.com/cloudquery/cloudquery/issues/2312


plugins/source.go
```diff
@@ -114,19 +116,21 @@ func (p *SourcePlugin) Sync(ctx context.Context, logger zerolog.Logger, spec spe
 	if concurrency == 0 {
 		concurrency = defaultConcurrency
 	}
-	logger.Info().Uint64("concurrency", concurrency).Msg("starting fetch")
+	logger.Info().Uint64("concurrency", concurrency).Msg("starting sync")
```
yevgenypats, line 119: nit: maybe starting `source sync` just to distinguish between CLI sync (which is source->destination) or should we keep it fetch...naming is hard....

plugins/source.go
```diff
@@ -114,19 +116,21 @@ func (p *SourcePlugin) Sync(ctx context.Context, logger zerolog.Logger, spec spe
...
 	startTime := time.Now()
-	tableNames, err := p.interpolateAllResources(spec.Tables)
+	tableNames, err := p.listAndValidateTables(spec.Tables, spec.SkipTables)
 	if err != nil {
 		return nil, err
 	}
 
+	logger.Debug().Interface("tables", tableNames).Msg("got table names")
+
 	for _, table := range p.tables {
```
yevgenypats, line 131: If we already validates tables, let's just use `tableName` and do `range _, table := range tableName` so the validation logic is in one place

plugins/source.go
```diff
@@ -154,25 +158,74 @@ func (p *SourcePlugin) Sync(ctx context.Context, logger zerolog.Logger, spec spe
...
-		return make([]string, 0), nil
+func (p *SourcePlugin) listAndValidateTables(tables, skipTables []string) ([]string, error) {
+	if len(tables) == 0 {
+		return nil, fmt.Errorf("list of tables is empty")
+	}
+
+	// return an error if skip tables contains a wildcard or glob pattern
+	for _, t := range skipTables {
+		if strings.Contains(t, "*") {
+			return nil, fmt.Errorf("glob matching in skipped table name %q is not currently supported", t)
```
yevgenypats, line 173: nit: I would put `is not supported` as we don't know if we want to support it in the future yet.

plugins/source.go
```diff
@@ -154,25 +158,74 @@ func (p *SourcePlugin) Sync(ctx context.Context, logger zerolog.Logger, spec spe
...
+	// wildcard should not be combined with other tables
 	if funk.ContainsString(tables, "*") {
-		return nil, fmt.Errorf("invalid \"*\" resource, with explicit resources")
+		return nil, fmt.Errorf("wildcard \"*\" table not allowed with explicit tables")
+	}
+
+	// return an error if other kinds of glob-matching is detected
+	for _, t := range tables {
+		if strings.Contains(t, "*") {
+			return nil, fmt.Errorf("glob matching in table name %q is not currently supported", t)
```
yevgenypats, line 197: same here: `....not supported...`

yevgenypats (changes requested):
Looks good. few nits.

plugins/source.go
hermanschaaf: `tableNames` is of type `[]string`, while `p.tables` is of type `schema.Tables`. We can't range over `tableNames`, since need the `schema.Table` here. We could add a lookup for the `schema.Table` object? I played around with it, but it doesn't seem better

plugins/source.go
hermanschaaf: updated 👍

plugins/source.go
hermanschaaf: updated 👍

plugins/source.go
```diff
@@ -154,25 +158,74 @@ func (p *SourcePlugin) Sync(ctx context.Context, logger zerolog.Logger, spec spe
...
+
+	// return an error if a given table name doesn't match any known tables
+	for _, t := range tables {
+		if !funk.ContainsString(p.tables.TableNames(), t) {
+			return nil, fmt.Errorf("name %s does not match any known table names", t)
+		}
+	}
+
+	// return an error if child table is included, but not its parent table
+	selectedTables := map[string]bool{}
```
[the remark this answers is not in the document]
yevgenypats: In that case, if child table is specified I'll automatically fetch parent tables if we are at it already.

yevgenypats:
@hermanschaaf can you solve the conflict and we can merge?

PR #242, opened 2022-10-04
fix: Race condition in log streaming

hermanschaaf:
This adds a sync.WaitGroup to ensure that we wait for all logs to finish streaming before returning from Terminate. Otherwise, the log file might get closed before we get to write the final lines.


clients/source.go
```diff
@@ -138,6 +141,7 @@ func (c *SourceClient) newManagedClient(ctx context.Context, path string) (*Sour
 				jsonToLog(c.logger, structuredLogLine)
 			}
 		}
+		c.wg.Done()
```
disq, line 144: Use `defer c.wg.Done()` (start of func) in case of panics.

disq (changes requested):

clients/source.go
```diff
@@ -30,6 +31,7 @@ type SourceClient struct {
 	conn           *grpc.ClientConn
 	grpcSocketName string
 	cmdWaitErr     error
+	wg             sync.WaitGroup
```
disq, line 34: This should always be a pointer (and initialized somewhere) to defend against any future variable copy issues.

disq (changes requested):

clients/source.go
hermanschaaf: Good point, fixed 👍

clients/source.go
```diff
@@ -222,6 +226,10 @@ func (c *SourceClient) Terminate() error {
 			return err
 		}
 	}
+
+	// wait for log streaming to complete before returning
+	c.wg.Wait()
```
yevgenypats, line 231: this wont be called if `c.cmd.Process.Kill` fails. I think we need maybe to log + wrap all errors ?

yevgenypats (changes requested):
Really good catch. found one bug

clients/source.go
hermanschaaf: Oh good point; maybe we can do `defer c.wg.Wait()` at the top of the function?

disq: Interesting approach, should work

disq: but if `Kill` fails (and the plugin is still alive) we can't be sure it's going to close the channel/logs anyway? Also maybe log the `Kill` `err` before returning it, so that if that happens and there's a hang, the log is printed so we know what the cause may be?

disq (changes requested):
We need the same logic in the destinationclient as well.

clients/source.go
hermanschaaf: Good point... it seems fairly unlikely that killing the plugin would fail but that the plugin can still be alive and streaming logs at that point, so let's log it so we can identify if that happens, but otherwise leave it like this for now?

hermanschaaf: Ok, updated to the way it was initially with an additional comment 👍

clients/source.go
```diff
@@ -119,7 +122,9 @@ func (c *SourceClient) newManagedClient(ctx context.Context, path string) (*Sour
 	}()
 	c.cmd = cmd
 
+	c.wg.Add(1)
```
yevgenypats, line 125: you prob also want to to do `cq.wg.Add` before the previous `go func()` . This is why we didn't see the `plugin exited` in the tests

clients/source.go
yevgenypats: Kill can't fail. it can only fail if the process already killed. I think we still need to do the

```
defer func() {
if err:= c.wg.wait(); err != nil {
   c.log().....
 }
}
```

yevgenypats (changes requested):
I think it needs a few more tweaks. still race conditions here.

clients/source.go
hermanschaaf: But I will add it to the other one too 👍

clients/source.go
```diff
@@ -211,9 +216,16 @@ func (c *SourceClient) Terminate() error {
 	}
 	if c.cmd != nil && c.cmd.Process != nil {
 		if err := c.cmd.Process.Kill(); err != nil {
+			// if we fail to kill the process, we also won't wait for logs to finish streaming
```
yevgenypats, line 219: I don't think this comment is correct or at least we need to see what are the error message from `kill`.

This is the snippet from signal, it really can only fail if it already closed. you can always kill a child process. so we do want to wait for the logs if it is already killed.
```
func (p *Process) signal(sig Signal) error {
	if p.Pid == -1 {
		return errors.New("os: process already released")
	}
	if p.Pid == 0 {
		return errors.New("os: process not initialized")
	}
	p.sigMu.RLock()
	defer p.sigMu.RUnlock()
	if p.done() {
		return ErrProcessDone
	}
	s, ok := sig.(syscall.Signal)
	if !ok {
		return errors.New("os: unsupported signal type")
	}
	if e := syscall.Kill(p.Pid, s); e != nil {
		if e == syscall.ESRCH {
			return ErrProcessDone
		}
		return e
	}
	return nil
}
```

yevgenypats (changes requested):
Commented on the kill side (prob I should just use the comments - sorry for the re-review)

clients/source.go
hermanschaaf: @yevgenypats Updated

yevgenypats (approved):

PR #252, opened 2022-10-05
fix: Exit gracefully on context cancelled

yevgenypats:
#### Summary

Found a nasty hang when context was cancelled. Still need to add some tests for that to catch regression.

### DIFF
--- a/internal/servers/destinations.go
+++ b/internal/servers/destinations.go
@@ -90,11 +90,11 @@ func (s *DestinationServer) Write(msg pb.Destination_WriteServer) error {
 			return status.Errorf(codes.InvalidArgument, "failed to unmarshal spec: %v", err)
 		}
 		select {
-			case resources <- resource:
-			case <-msg.Context().Done():
-				close(resources)
-				wg.Wait()
-				return msg.Context().Err()
+		case resources <- resource:
+		case <-msg.Context().Done():
+			close(resources)
+			wg.Wait()
+			return msg.Context().Err()
 		}
 	}
 }
--- a/internal/servers/destinations.go
+++ b/internal/servers/destinations.go
@@ -89,7 +89,13 @@ func (s *DestinationServer) Write(msg pb.Destination_WriteServer) error {
 			wg.Wait()
 			return status.Errorf(codes.InvalidArgument, "failed to unmarshal spec: %v", err)
 		}
-		resources <- resource
+		select {
+			case resources <- resource:
+			case <-msg.Context().Done():
+				close(resources)
+				wg.Wait()
+				return msg.Context().Err()
+		}
 	}
 }
 
--- a/serve/destination.go
+++ b/serve/destination.go
@@ -134,7 +134,7 @@ func newCmdDestinationServe(destination *destinationServe) *cobra.Command {
 					logger.Info().Str("address", listener.Addr().String()).Msg("Got interrupt. Destination plugin server shuting down")
 					s.Stop()
 				case <-ctx.Done():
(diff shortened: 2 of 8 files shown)


internal/servers/destinations.go
```diff
@@ -89,7 +89,13 @@ func (s *DestinationServer) Write(msg pb.Destination_WriteServer) error {
 			wg.Wait()
 			return status.Errorf(codes.InvalidArgument, "failed to unmarshal spec: %v", err)
 		}
-		resources <- resource
+		select {
+		case resources <- resource:
+		case <-msg.Context().Done():
+			close(resources)
```
disq, line 95: I would add these two lines as a defer statement and remove multiple copies of it here and above.

disq (changes requested):

internal/servers/destinations.go
yevgenypats: That won't work because when `err == io.EOF` I want wait for `write` so we can finish writing and send the writeSummary

serve/destination.go
```diff
@@ -120,8 +121,25 @@ func newCmdDestinationServe(destination *destinationServe) *cobra.Command {
...
+			}()
 
-			logger.Info().Str("address", listener.Addr().String()).Msg("server listening")
+			go func() {
+				select {
+				case <-c:
+					logger.Info().Str("address", listener.Addr().String()).Msg("Got interrupt. Destination plugin server shuting down")
+					s.Stop()
+				case <-ctx.Done():
+					logger.Info().Str("address", listener.Addr().String()).Msg("Context cancelled. Destination plugin server shuting down")
```
hermanschaaf, line 137:
```suggestion
					logger.Info().Str("address", listener.Addr().String()).Msg("Context cancelled. Destination plugin server shutting down")
```

serve/destination.go
```diff
@@ -120,8 +121,25 @@ func newCmdDestinationServe(destination *destinationServe) *cobra.Command {
...
+			signal.Notify(c, os.Interrupt)
+			defer func() {
+				signal.Stop(c)
+			}()
 
-			logger.Info().Str("address", listener.Addr().String()).Msg("server listening")
+			go func() {
+				select {
+				case <-c:
+					logger.Info().Str("address", listener.Addr().String()).Msg("Got interrupt. Destination plugin server shuting down")
```
hermanschaaf, line 134:
```suggestion
					logger.Info().Str("address", listener.Addr().String()).Msg("Got interrupt. Destination plugin server shutting down")
```

serve/source.go
```diff
@@ -123,7 +124,25 @@ func newCmdSourceServe(source *sourceServe) *cobra.Command {
...
+			}()
+
+			go func() {
+				select {
+				case <-c:
+					logger.Info().Str("address", listener.Addr().String()).Msg("Got interrupt. Source plugin server shuting down")
+					s.Stop()
+				case <-ctx.Done():
+					logger.Info().Str("address", listener.Addr().String()).Msg("Context cancelled. Source plugin server shuting down")
+					s.Stop()
```
hermanschaaf, line 141:
```suggestion
				case <-c:
					logger.Info().Str("address", listener.Addr().String()).Msg("Got interrupt. Source plugin server shutting down")
					s.Stop()
				case <-ctx.Done():
					logger.Info().Str("address", listener.Addr().String()).Msg("Context cancelled. Source plugin server shutting down")
					s.Stop()
```

hermanschaaf (approved):
Nice catch :+1:

disq (approved):

PR #257, opened 2022-10-07
fix: Recover panic in table resolver and object resolver flows

yevgenypats:
This also adds quite needed tests for our scheduler

### DIFF
--- a/schema/table.go
+++ b/schema/table.go
@@ -180,6 +180,7 @@ func (t Table) Resolve(ctx context.Context, meta ClientMeta, parent *Resource, r
 					sentry.CurrentHub().CaptureMessage(stack)
 				})
 				meta.Logger().Error().Interface("error", err).Str("table", t.Name).TimeDiff("duration", time.Now(), startTime).Str("stack", stack).Msg("table resolver finished with panic")
+				//nolint:all as we use atomic. (false positive)
 				atomic.AddUint64(&summary.Panics, 1)
 			}
 			close(res)
@@ -187,6 +188,7 @@ func (t Table) Resolve(ctx context.Context, meta ClientMeta, parent *Resource, r
 		meta.Logger().Debug().Str("table", t.Name).Msg("table resolver started")
 		if err := t.Resolver(ctx, meta, parent, res); err != nil {
 			meta.Logger().Error().Str("table", t.Name).TimeDiff("duration", time.Now(), startTime).Err(err).Msg("table resolver finished with error")
+			//nolint:all as we use atomic. (false positive)
 			atomic.AddUint64(&summary.Errors, 1)
 			return
 		}
--- a/schema/table.go
+++ b/schema/table.go
@@ -70,6 +70,12 @@ type Table struct {
 	columnsMap map[string]int
 }
 
+func (s *SyncSummary) Merge(other SyncSummary) {
+	atomic.AddUint64(&s.Resources, other.Resources)
+	atomic.AddUint64(&s.Errors, other.Errors)
+	atomic.AddUint64(&s.Panics, other.Panics)
+}
+
 func (tt Tables) TableNames() []string {
 	ret := []string{}
 	for _, t := range tt {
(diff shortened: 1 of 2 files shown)


schema/table_test.go
```diff
@@ -56,17 +214,32 @@ func TestTableExecution(t *testing.T) {
...
 					t.Errorf("expected %v, got %v", tc.Resources[i].Data, resource)
 				}
 				i++
 			}
+			if tc.Summary.Resources != i {
+				t.Errorf("expected %d resources, got %d", tc.Summary.Resources, i)
+			}
+			if tc.Summary.Resources != summary.Resources {
+				t.Errorf("expected %d summary resources, got %d", tc.Summary.Resources, summary.Resources)
+			}
```
hermanschaaf, line 234: nit: Looks like the second comparison here can never fail except if either: the case above it already failed, or the test is set up incorrectly. I'd probably remove the `!= i` check

hermanschaaf (approved):
Nice, I like the tests. LGTM 👍

schema/table_test.go
yevgenypats: it can (It did and this is why I've added it :) ) because the first check checks how many resources were returned on the channel and the second one just checks the statistics struct

schema/table.go
```diff
@@ -159,21 +165,20 @@ func (t Table) TableNames() []string {
...
 	res := make(chan interface{})
 	startTime := time.Now()
 	go func() {
 		defer func() {
 			if err := recover(); err != nil {
+				stack := fmt.Sprintf("%s\n%s", err, string(debug.Stack()))
 				sentry.WithScope(func(scope *sentry.Scope) {
 					scope.SetTag("table", t.Name)
-					sentry.CurrentHub().Recover(err)
+					sentry.CurrentHub().CaptureMessage(stack)
```
disq, line 180: Sentry should already include the stack trace?

schema/table_test.go
hermanschaaf: Oh I see now, sorry yeah. It's because it's written in `want != got` instead of `got != want` order that I misread it. My nit would then be to change the ordering :)

schema/table.go
yevgenypats: their logic is incorrect it doesn't print it in the right way and try to do some "smart things" - just a string of the stack trace should work for us I think.

schema/table.go
```diff
@@ -159,21 +165,20 @@ func (t Table) TableNames() []string {
 }
 
 // Call the table resolver with with all of it's relation for every reolved resource
-func (t Table) Resolve(ctx context.Context, meta ClientMeta, parent *Resource, resolvedResources chan<- *Resource) SyncSummary {
+func (t Table) Resolve(ctx context.Context, meta ClientMeta, parent *Resource, resolvedResources chan<- *Resource) (summary *SyncSummary) {
```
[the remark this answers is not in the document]
yevgenypats: I've already changed it back to struct (was experimenting with both options.)

yevgenypats:
@disq @hermanschaaf can you please take a look at the lint error - https://github.com/cloudquery/plugin-sdk/actions/runs/3203350691/jobs/5233341574#step:4:26

I think it's a false positive but just want to be sure before I ignore it.

schema/table.go
```diff
@@ -194,47 +199,40 @@ func (t Table) Resolve(ctx context.Context, meta ClientMeta, parent *Resource, r
...
-
-			atomic.AddUint64(&summary.Resources, 1)
-			resolvedResources <- resource
-			for _, rel := range t.Relations {
-				relationSummary := rel.Resolve(ctx, meta, resource, resolvedResources)
-				atomic.AddUint64(&summary.Errors, relationSummary.Errors)
-				atomic.AddUint64(&summary.Panics, relationSummary.Panics)
-				atomic.AddUint64(&summary.Resources, relationSummary.Resources)
-			}
+			summary.Merge(t.resolveObject(ctx, meta, parent, objects[i], resolvedResources))
```
disq, line 202: Does this work with summary being non-ptr? Maybe return the new value from `Merge` and set it to summary?
`summary = summary.Merge(t.resolveObject(...))`

yevgenypats: I think that work because `Merge` has a pointer `receiver`

disq:
> I think it's a false positive but just want to be sure before I ignore it.

@yevgenypats I don't see how it can be a datarace, as we close `res` in the goroutine itself, so `range res` can't stop before resolvering finishes.

yevgenypats:
That's right. I think it is a false positive because the linter doesn't detect that Im using `atomic` ?

disq:
> That's right. I think it is a false positive because the linter doesn't detect that Im using `atomic` ?

`atomic` is unrelated in this error, it can still be a datarace if the goroutine doesn't finish before the function returns. But it does.

hermanschaaf:
@yevgenypats I'm not 100% sure it is a data race either, but rather than ignore the warning, let's pass in a pointer to the goroutine so that it's definitely safe?

```go func(syncSummary *SyncSummary) {
...
atomic.AddUint64(&syncSummary.Panics, 1)
...
atomic.AddUint64(&syncSummary.Errors, 1)
...
}(&summary)
```

yevgenypats:
> 

good idea. done.

schema/table.go
```diff
@@ -258,5 +256,22 @@ func (t Table) resolveColumns(ctx context.Context, meta ClientMeta, resource *Re
...
+	if t.PostResourceResolver != nil {
+		meta.Logger().Trace().Str("table", t.Name).Msg("post resource resolver started")
+		if err := t.PostResourceResolver(ctx, meta, resource); err != nil {
+			meta.Logger().Error().Str("table", t.Name).Stack().Err(err).Msg("post resource resolver finished with error")
+			summary.Errors++
+		} else {
+			meta.Logger().Trace().Str("table", t.Name).Msg("post resource resolver finished successfully")
+		}
+	}
+	summary.Resources = 1
```
disq, line 269:
```suggestion
	summary.Resources++
```

schema/table_test.go
yevgenypats: I think it is usually `expected != got` but also sometimes it's hard to keep the order consistent because for example you can't do `1 != some_var` so you still have constraints.

hermanschaaf: In Go it's usually `got != expected`, so that it's consistent when you have a constant `got != 1` as well. https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures

PR #260, opened 2022-10-07
fix: Add missing defer

hermanschaaf:
The waitgroup here should defer before being marked as done. Spotted this while looking through the code, but this wasn't causing any serious issues as far as I'm aware though.


yevgenypats (approved):

yevgenypats:
Nice catch!

PR #263, opened 2022-10-07
fix: Add custom log reader implementation to fix hang on long log lines

hermanschaaf:
This fixes an issue where the CLI would hang indefinitely when it encountered a log line from a plugin that was longer than the buffer size allowed for by `bufio.Scanner`. The fix uses a wrapper around `bufio.NewReader` to skip lines that are too long to be parsed, printing an error in the log with the first 1000 characters to help with identification.

### DIFF
--- a/clients/destination.go
+++ b/clients/destination.go
@@ -1,10 +1,11 @@
 package clients
 
 import (
-	"bufio"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"io"
 	"net"
 	"os"
 	"os/exec"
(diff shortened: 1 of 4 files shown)


clients/destination.go
```diff
@@ -124,12 +125,18 @@ func (c *DestinationClient) newManagedClient(ctx context.Context, path string) (
...
 	go func() {
 		defer c.wg.Done()
-		scanner := bufio.NewScanner(reader)
-		for scanner.Scan() {
+		lr := newLogReader(reader)
+		for {
+			line, err := lr.NextLine()
+			if errors.Is(err, io.EOF) {
+				break
+			} else if err != nil {
```
yevgenypats, line 133: can we somehow distinguish between line-to-long to some other errors where we actually want to exit.

yevgenypats:
Is it possible to do some benchmarks vs the regular scanner just to see that we are not eating cpu/memory in comparision to the built-in scanner?

hermanschaaf:
@yevgenypats I added some benchmarks:

```
Benchmark_BufferedScanner-8   	261038530	         4.488 ns/op	       0 B/op	       0 allocs/op
Benchmark_LogReader-8         	79254348	        14.98 ns/op	       0 B/op	       0 allocs/op
```

A tiny bit slower than BufferedScanner in this test (but we're talking nanoseconds here) and no additional allocs.

disq (approved):

yevgenypats (approved):

Issue #264, opened 2022-10-07
Parallelization of Detail Calls 


bbernays:
Using `PreResourceResolver` should enable all of the resolvers to be invoked in parallel (with concurrency guardrails) this is impacting user experience that are fetching resources that require the list/detail pattern: https://github.com/cloudquery/cloudquery/issues/2392

This functionality is also blocking the full removal of the ListDetail pattern in the AWS Plugin: https://github.com/cloudquery/cloudquery/pull/2460

yevgenypats:
Solved with https://github.com/cloudquery/plugin-sdk/pull/268

PR #268, opened 2022-10-09
feat!: Support table_concurrency and resource_concurrency

yevgenypats:
#### Summary

Closes https://github.com/cloudquery/cloudquery/issues/2313 https://github.com/cloudquery/cloudquery/issues/159 https://github.com/cloudquery/plugin-sdk/issues/264

This renames `concurrency` to `table_concurrency` (this has the same behaviour as previously but the name was too generic).

Also this introduce a new option `resource_concurrency` - this limits the number of go routines that resolve a specific resource (useful when a lot of column resolvers that have api calls and `PreResourceResolver`).

Importe Note: both options are only for top-level concurrency control i.e we only spawn go-routines for parent tables and for resources of parent table. The reason for that is that there is no concurrency model that can work with one variable for recursive calls otherwise it will get deadlock.

So right now the SDK will support only for one level - in the future if we want to support additional level we can do `table_concurrency_1` and `resource_concurrency_1` but I think for now we can skip this.


schema/table.go
```diff
@@ -199,9 +200,26 @@ func (t Table) Resolve(ctx context.Context, meta ClientMeta, parent *Resource, r
 			continue
 		}
 		for i := range objects {
-			summary.Merge(t.resolveObject(ctx, meta, parent, objects[i], resolvedResources))
+			i := i
+
+			// write now we support only concurrency
```
disq, line 205:
```suggestion
			// right now we support only concurrency
```

schema/table.go
```diff
@@ -199,9 +200,26 @@ func (t Table) Resolve(ctx context.Context, meta ClientMeta, parent *Resource, r
 			continue
 		}
 		for i := range objects {
-			summary.Merge(t.resolveObject(ctx, meta, parent, objects[i], resolvedResources))
+			i := i
+
+			// right now we support only concurrency
```
hermanschaaf, line 205: I don't quite understand what this comment is trying to convey; it looks like we're supporting both?

plugins/source.go
```diff
@@ -106,25 +102,25 @@ func (p *SourcePlugin) Version() string {
...
-	if concurrency == 0 {
-		concurrency = defaultConcurrency
+	c, err := p.newExecutionClient(ctx, logger, spec)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create execution client for source plugin %s: %w", p.name, err)
 	}
-	logger.Info().Uint64("concurrency", concurrency).Msg("starting source plugin sync")
-	goroutinesSem := semaphore.NewWeighted(helpers.Uint64ToInt64(concurrency))
+	logger.Info().Interface("spec", spec).Msg("starting sync")
+	goroutinesSem := semaphore.NewWeighted(helpers.Uint64ToInt64(spec.TableConcurrency))
```
hermanschaaf, line 119: Should we perhaps rename `goroutinesSem` to `tablesSem` for consistency?

specs/source.go
```diff
@@ -20,8 +25,9 @@ type Source struct {
 	// For the gRPC registry the path will be the address of the gRPC server: host:port
 	Path string `json:"path,omitempty"`
 	// Registry can be github,local,grpc.
-	Registry    Registry `json:"registry,omitempty"`
-	Concurrency uint64   `json:"concurrency,omitempty"`
+	Registry            Registry `json:"registry,omitempty"`
+	TableConcurrency    uint64   `json:"table_concurrency,omitempty"`
```
hermanschaaf, line 29: Just a thought: we could keep this change backwards-compatible by keeping `concurrency` as an option, but using it to set `table_concurrency` if it is provided.

hermanschaaf (approved):
This is great 👍 I left a few comments / thoughts.

Not for this PR, but I suppose we could add another option later to control concurrency for child tables at well. It wouldn't be a global pool (could get deadlock), but we could say, for example, that every table is only allowed to spin up 10 goroutines for its children. That way you'd have some control and children can be processed concurrently, but you can't run into a deadlock situation. Goroutines and memory usage can theoretically grow unbounded though.

There are other more complicated options too. Having parent tables relinquish control over their goroutine before waiting on child tables to complete should also help avoid deadlock. But it gets real complicated fast; I like the approach in this PR.

schema/table.go
```diff
@@ -199,9 +200,26 @@ func (t Table) Resolve(ctx context.Context, meta ClientMeta, parent *Resource, r
...
+			if resourcesSem == nil {
+				summary.Merge(t.resolveObject(ctx, meta, parent, objects[i], resolvedResources))
+			} else {
+				if err := resourcesSem.Acquire(ctx, 1); err != nil {
```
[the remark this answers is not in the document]
yevgenypats: I wouldn't go that path. also this can hang for much longer because it tries to acquire more rather then a simple fifo when acquiring one.

yevgenypats:
> This is great 👍 I left a few comments / thoughts.
> 
> Not for this PR, but I suppose we could add another option later to control concurrency for child tables at well. It wouldn't be a global pool (could get deadlock), but we could say, for example, that every table is only allowed to spin up 10 goroutines for its children. That way you'd have some control and children can be processed concurrently, but you can't run into a deadlock situation. Goroutines and memory usage can theoretically grow unbounded though.
> 
> There are other more complicated options too. Having parent tables relinquish control over their goroutine before waiting on child tables to complete should also help avoid deadlock. But it gets real complicated fast; I like the approach in this PR.

Thanks! I don't think the 10 per child will work as then we can't bound goroutines (this is what we had with the list/detail resolvers where we just spawn up as many goroutines as we wanted) which is the whole point otherwise we can just have unlimited go routines and that's it but that we know can cause trouble.

schema/table.go
disq: Yeah it adds unnecessary confusion. the extra `nil` in line 273 should be enough.

specs/source.go
yevgenypats: let's just break it. I don't think anyone is using this at the moment anyway given we just released it last week.

disq: It would open up more confusion, ie. should check and error out if both of those are set by the user.

schema/table.go
yevgenypats: I updated the comment (prob got distracted in the middle of the sentence)

plugins/source.go
yevgenypats: Fixed

specs/source.go
yevgenypats: if someone will use `concurrency` it will error out by default because we check if there are any unknown fields via built-in `json` decoder

PR #269, opened 2022-10-10
fix: DeleteStale feature

shimonp21:
#### Summary

testing matrix (two syncs, different versions for first and second sync )
image

---

- [ ] Update or add tests 🧪

### DIFF
--- a/plugins/.snapshots/TestGenerateSourcePluginDocs-relation_table.md
+++ b/plugins/.snapshots/TestGenerateSourcePluginDocs-relation_table.md
@@ -10,8 +10,8 @@ This table depends on [`test_table`](test_table.md).
 ## Columns
 | Name          | Type          |
 | ------------- | ------------- |
-|_cq_id (PK)|UUID|
-|_cq_parent_id|UUID|
 |_cq_source_name|String|
 |_cq_sync_time|Timestamp|
+|_cq_id (PK)|UUID|
+|_cq_parent_id|UUID|
 |string_col|String|
--- a/plugins/destination.go
+++ b/plugins/destination.go
@@ -73,6 +73,8 @@ func (p *DestinationPlugin) Init(ctx context.Context, logger zerolog.Logger, spe
 
 // we implement all DestinationClient functions so we can hook into pre-post behavior
 func (p *DestinationPlugin) Migrate(ctx context.Context, tables schema.Tables) error {
+	SetDestinationManagedCqColumns(tables)
+
 	if p.client == nil {
 		return fmt.Errorf("destination client not initialized")
 	}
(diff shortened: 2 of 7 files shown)


plugins/destination.go
```diff
@@ -127,3 +125,13 @@ func (p *DestinationPlugin) Close(ctx context.Context) error {
 	}
 	return p.client.Close(ctx)
 }
+
+// Overwrites or adds the CQ columns that are managed by the destination plugins (_cq_sync_time, _cq_source_name).
+func OverwriteOrAddCqColumns(tables []*schema.Table) {
```
disq, line 130:
```suggestion
func SetCqColumns(tables []*schema.Table) {
```

disq (approved):

plugins/.snapshots/TestGenerateSourcePluginDocs-relation_table.md
```diff
@@ -10,8 +10,8 @@ This table depends on [`test_table`](test_table.md).
 ## Columns
 | Name          | Type          |
 | ------------- | ------------- |
-|_cq_id (PK)|UUID|
-|_cq_parent_id|UUID|
 |_cq_source_name|String|
```
yevgenypats, line 13: Why `_cq_source_name` is still here if we removed it from the source plugin?

schema/table.go
```diff
@@ -163,7 +157,7 @@ func (t Table) TableNames() []string {
 }
 
 // Call the table resolver with with all of it's relation for every reolved resource
-func (t Table) Resolve(ctx context.Context, meta ClientMeta, parent *Resource, resourcesSem *semaphore.Weighted, resolvedResources chan<- *Resource) (summary SyncSummary) {
+func (t *Table) Resolve(ctx context.Context, meta ClientMeta, parent *Resource, resourcesSem *semaphore.Weighted, resolvedResources chan<- *Resource) (summary SyncSummary) {
```
yevgenypats, line 160: Im not sure we really want to turn it into `*` at least not in this PR or why it is needed at all?

yevgenypats (changes requested):
Looks great! Let's just not change receiver to `*table`  as I recall there were reasons for why it is not pointer receiver. let's keep it `table` and maybe open an issue to look into making it pointer receiever in the future but it will need to take special care of.

schema/table.go
shimonp21: https://go.dev/doc/faq#methods_on_values_or_pointers

The pointer receiver is required because I added a method that changes the table (if the receiver is a value, it is impossible to have methods that modify the `Table`).

From the FAQ I linked:

> Next is consistency. If some of the methods of the type must have pointer receivers, the rest should too, so the method set is consistent regardless of how the type is used. See the section on [method sets](https://go.dev/doc/faq#different_method_sets) for details.

And also 

> Second is the consideration of efficiency. If the receiver is large, a big struct for instance, it will be much cheaper to use a pointer receiver.

The table struct is very large..

Unless there is a specific reason to use value-receiver (I can't think of one here), than pointer-receivers are a good default.

yevgenypats (approved):

PR #270, opened 2022-10-10
feat: Plugin vs. Spec

disq:
For initial review. ~The `Warnings` part I don't like but it can be improved with suggestions.~

### DIFF
--- a/specs/spec.go
+++ b/specs/spec.go
@@ -11,8 +11,10 @@ import (
 type Kind int
 
 type Spec struct {
-	Kind Kind        `json:"kind"`
-	Spec interface{} `json:"spec"`
+	Kind   Kind        `json:"kind"`
+	Plugin interface{} `json:"plugin"`
+
+	Warnings []string `json:"-"`
 }
 
 const (
@@ -55,7 +57,10 @@ func KindFromString(s string) (Kind, error) {
 
 func (s *Spec) UnmarshalJSON(data []byte) error {
 	var t struct {
-		Kind Kind        `json:"kind"`
+		Kind   Kind        `json:"kind"`
+		Plugin interface{} `json:"plugin"`
+
+		// Deprecated
 		Spec interface{} `json:"spec"`
 	}
 	dec := json.NewDecoder(bytes.NewReader(data))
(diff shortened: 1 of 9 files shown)


hermanschaaf:
Seems like a pretty decent backwards-compatible solution 👍

specs/spec.go
```diff
@@ -11,8 +11,10 @@ import (
 type Kind int
 
 type Spec struct {
-	Kind Kind        `json:"kind"`
-	Spec interface{} `json:"spec"`
+	Kind   Kind        `json:"kind"`
+	Plugin interface{} `json:"plugin"`
+
+	Warnings []string `json:"-"`
```
hermanschaaf, line 17: Another option, if we don't like `Warnings` on this struct (which is a bit weird, I agree), but want to maintain backwards-compatibility in the SDK, is we can return a custom Error object with a `Warnings` attribute from `SpecUnmarshalYamlStrict`

specs/spec.go
```diff
@@ -65,29 +70,38 @@ func (s *Spec) UnmarshalJSON(data []byte) error {
...
 	}
-	b, err := json.Marshal(t.Spec)
+
+	var (
+		b   []byte
+		err error
+	)
+	if t.Spec != nil {
+		b, err = json.Marshal(t.Spec)
+		s.Warnings = append(s.Warnings, `"spec" keyword is deprecated at this level and will be removed in a future version. Use "plugin" instead`)
```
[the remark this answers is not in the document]
disq: done :)

specs/spec.go
disq: `error` but not an error? that's also equally weird (and even unclear maybe?)

disq:
> Seems like a pretty decent backwards-compatible solution 👍

I almost renamed the inner-level `spec` to be `plugin` as well... and cmd+z didn't work so had to revert by hand.

yevgenypats (changes requested):
I don't think we need any warnings at this stage. developing warnings parsing system for keyword deprecation in json is out of scope for this PR.

disq:
> I don't think we need any warnings at this stage. developing warnings parsing system for keyword deprecation in json is out of scope for this PR.

removed

PR #271, opened 2022-10-10
fix: Make concurrency change backwards-compatible

hermanschaaf:
This makes https://github.com/cloudquery/plugin-sdk/pull/268 backwards-compatible with prior versions, as support for `concurrency` is retained for now.

### DIFF
--- a/specs/source.go
+++ b/specs/source.go
@@ -26,6 +26,7 @@ type Source struct {
 	Path string `json:"path,omitempty"`
 	// Registry can be github,local,grpc.
 	Registry            Registry `json:"registry,omitempty"`
+	Concurrency         uint64   `json:"concurrency,omitempty"` // deprecated: use TableConcurrency and ResourceConcurrency instead
 	TableConcurrency    uint64   `json:"table_concurrency,omitempty"`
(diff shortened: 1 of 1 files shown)


specs/source.go
```diff
@@ -53,6 +54,12 @@ func (s *Source) SetDefaults() {
 		s.Tables = []string{"*"}
 	}
 
+	if s.Concurrency != 0 {
+		// attempt to make a sensible backwards-compatible choice, but the CLI
+		// should raise a warning about this until concurrency is fully removed.
```
disq, line 59: also need to check if both are set... and not overwrite if so, just leave the SDK to warn? oof too complex

hermanschaaf: Yeah that's probably fine...

disq: But then it will break without warning when we remove this block in a future release.

hermanschaaf: No, neither `table_concurrency` nor `resource_concurrency` are required, and if we remove this block it will be because we completely removed the `concurrency` option, so users will get an error about an unrecognized option.

disq: By "block" I meant this `if s.Concurrency != 0 {` code block. Resolving anyway 🤷

yevgenypats (approved):

PR #276, opened 2022-10-10
fix: Make plugin instead of spec

yevgenypats:
#### Summary

I think this should be simpler solution then https://github.com/cloudquery/plugin-sdk/pull/270 changing only the user layer without breaking also all our official and community plugins code-wise.

### DIFF
--- a/specs/spec.go
+++ b/specs/spec.go
@@ -55,8 +55,8 @@ func KindFromString(s string) (Kind, error) {
 
 func (s *Spec) UnmarshalJSON(data []byte) error {
 	var t struct {
-		Kind Kind        `json:"kind"`
-		Spec interface{} `json:"spec,omitempty"`
+		Kind   Kind        `json:"kind"`
+		Spec   interface{} `json:"spec,omitempty"`
 		Plugin interface{} `json:"plugin,omitempty"`
 	}
 	dec := json.NewDecoder(bytes.NewReader(data))
@@ -76,6 +76,7 @@ func (s *Spec) UnmarshalJSON(data []byte) error {
 	}
 	var b []byte
 	var err error
+	// nolint:all
 	if t.Spec != nil {
 		b, err = json.Marshal(t.Spec)
 		if err != nil {
@@ -88,7 +89,7 @@ func (s *Spec) UnmarshalJSON(data []byte) error {
 		}
 	} else {
 		return fmt.Errorf("plugin is required and shouldn't be empty")
-	} 
+	}
 	dec = json.NewDecoder(bytes.NewReader(b))
 	dec.UseNumber()
 	dec.DisallowUnknownFields()


specs/destination_test.go
```diff
@@ -109,7 +109,8 @@ var destinationUnmarshalSpecValidateTestCases = []struct {
 	{
 		"required_name",
 		`kind: destination
-spec:`,
+spec:
+  version: v1.1.0`,
```
disq, line 113: Unused field? Wouldn't this fail because of `DisallowUnknownFields`?

specs/spec.go
```diff
@@ -73,9 +74,21 @@ func (s *Spec) UnmarshalJSON(data []byte) error {
...
+		if err != nil {
+			return err
+		}
+	} else if t.Plugin != nil {
+		b, err = json.Marshal(t.Plugin)
+		if err != nil {
+			return err
+		}
+	} else {
+		return fmt.Errorf("plugin is required and shouldn't be empty")
```
disq, line 91: This check wasn't here before, new functionality?

disq (changes requested):
to reiterate: You're suggesting to keep all code complexity (and developer "unfriendliness") associated with double `spec` and not even checking for the new syntax (with `plugin`) in tests?

specs/spec.go
yevgenypats: yeah I think it was a bug prob but can go either way

specs/destination_test.go
yevgenypats: there are no unknown fields. version is known.

yevgenypats:
> to reiterate: You're suggesting to keep all code complexity (and developer "unfriendliness") associated with double `spec` and not even checking for the new syntax (with `plugin`) in tests?

Exactly. I think the value of changing `spec` to `plugin` is not that big in the first place but let's say we change and in a backward compatible way and now we need to update all documentation, I think we shouldn't change now also the codewise, let's just leave as is and avoid updating all plugins for that, I just feel we took the naming thing a bit too far.

yevgenypats:
> > to reiterate: You're suggesting to keep all code complexity (and developer "unfriendliness") associated with double `spec` and not even checking for the new syntax (with `plugin`) in tests?
> 
> Exactly. I think the value of changing `spec` to `plugin` is not that big in the first place but let's say we change and in a backward compatible way and now we need to update all documentation, I think we shouldn't change now also the codewise, let's just leave as is and avoid updating all plugins for that, I just feel we took the naming thing a bit too far.

At this point Im not even sure we really need to change it all given the hussle just for this change and then the docs just for the sake of the name.

specs/destination_test.go
disq: ah sorry I thought of something like the spec version for some reason

disq:
> At this point Im not even sure we really need to change it all given the hussle just for this change and then the docs just for the sake of the name.

You're still unsure this change is needed that explains why this PR just renames the field from the user standpoint but doesn't change anything else internally.

yevgenypats:
Closing as no-fix for now and will be gathering more experience to see if the `spec` -> `plugin` name is a real issue

PR #279, opened 2022-10-11
fix: Call Release on resource semaphore

hermanschaaf:
We were missing a `Release` call on the `resource_concurrency` semaphore.


disq (approved):

yevgenypats (changes requested):
small nit on order

yevgenypats (approved):

PR #281, opened 2022-10-11
fix: Tests

shimonp21:
Up to now, we only tested that the first resource has values in all columns. A more correct test tests that every column has at least one value in every column. This is important to avoid failures that are based just on ordering of resources (especially now that we added more concurrency to resolving). Important for tests with more than one resource.

#### Summary

---

- [ ] Update or add tests 🧪


yevgenypats (approved):

hermanschaaf (commented):
Great find 👍

PR #287, opened 2022-10-12
feat(tests): More faker options

disq:
This is so that we can remove most of the `cloudquery/faker` usage. See https://github.com/cloudquery/cloudquery/pull/2726


faker/options.go
```diff
@@ -0,0 +1,23 @@
...
+func WithSkipEFace() Option {
+	return func(f *faker) {
+		f.ignoreEFace = true
+	}
+}
+
+func WithSkipFields(fieldName ...string) Option {
+	return func(f *faker) {
+		for _, field := range fieldName {
+			f.skipFields[field] = struct{}{}
```
erezrokah, line 20: Can you give an example of fields we skip? I remember needing to skip only fields faker couldn't fake for some reason (usually a bug in faker). Maybe we should just ignore those

disq: > Can you give an example of fields we skip? I remember needing to skip fields only when faker couldn't fake them for some reason (usually a bug in faker). Maybe we should just ignore those

Just search for `faker.FakeDataSkipFields`

faker/options.go
```diff
@@ -0,0 +1,23 @@
...
+
+func WithMaxDepth(depth int) Option {
+	return func(f *faker) {
+		f.maxDepth = depth
+	}
+}
+
+func WithSkipEFace() Option {
+	return func(f *faker) {
+		f.ignoreEFace = true
```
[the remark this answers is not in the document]
disq: Sure, done in [`ca4096f` (#287)](https://github.com/cloudquery/plugin-sdk/pull/287/commits/ca4096f20b6984f0b6fc9a27281e6fe27ae294b3)

faker/options.go
erezrokah: What happens if we use our faker and don't use `FakeDataSkipFields`?
I remember needing to skip fields due to infinite recursion, but that shouldn't happen anymore.

disq: I'll give it a try

disq: @erezrokah seems to work (I wasn't able to test wafv2 and iot, but others are OK) so removed the option in [`006b015` (#287)](https://github.com/cloudquery/plugin-sdk/pull/287/commits/006b0158290d1ae019b382a3114d54a21fe05fdd)

faker/helpers.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+	MustFakeObject(&s)
+	return s.Name
+}
+
+func Word() string {
+	return Name()
+}
+
+func RandomUnixTime() int64 {
```
erezrokah, line 28: Similar to `SkipFields`. Are these needed?  Do the tests pass without them?

disq: They apparently do, all removed @erezrokah

erezrokah (approved):

erezrokah (approved):

faker/faker.go
```diff
@@ -52,7 +53,13 @@ func (f faker) getFakedValue(a interface{}) (reflect.Value, error) {
 				}
 				val, err := f.getFakedValue(v.Field(i).Interface())
 				if err != nil {
-					fmt.Println(err)
+					if err == errEFaceNotAllowed { // skip empty interfaces
+						continue
+					}
+
+					if f.verbose {
+						fmt.Println(err)
```
yevgenypats, line 61: can we use logger instead of `fmt.Println` this way we wont need the verbose flag as well as it will but just configured by the parent logger

yevgenypats (changes requested):
Looks good. Let's use logger and add `WithLogger` if it doesn't exist yet

disq:
> Looks good. Let's use logger and add `WithLogger` if it doesn't exist yet

Done

faker/faker.go
```diff
@@ -160,6 +161,7 @@ func FakeObject(obj interface{}, opts ...Option) error {
 	}
 	f := &faker{
 		maxDepth: 12,
+		logger:   zerolog.Nop(),
```
yevgenypats, line 164: I would use global one by default, but not super important for this PR.

yevgenypats (approved):

PR #292, opened 2022-10-13
fix(tests): Parallel plugin testing, remove old faker

disq:
Something like https://github.com/cloudquery/cloudquery/pull/2788 but in the sdk.

### DIFF
--- a/plugins/source_testing.go
+++ b/plugins/source_testing.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"testing"
 
-	"github.com/cloudquery/faker/v3"
 	"github.com/cloudquery/plugin-sdk/schema"
 	"github.com/cloudquery/plugin-sdk/specs"
 	"github.com/rs/zerolog"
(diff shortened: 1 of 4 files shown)


plugins/source_testing.go
```diff
@@ -21,13 +20,15 @@ type ResourceTestCase struct {
...
-	_ = faker.SetRandomMapAndSliceMinSize(1)
-	_ = faker.SetRandomMapAndSliceMaxSize(1)
+type TestSourcePluginOptions struct {
+	NoParallel bool
 }
 
-func TestSourcePluginSync(t *testing.T, plugin *SourcePlugin, logger zerolog.Logger, spec specs.Source) {
+func TestSourcePluginSync(t *testing.T, plugin *SourcePlugin, logger zerolog.Logger, spec specs.Source, opts TestSourcePluginOptions) {
 	t.Helper()
+	if !opts.NoParallel {
```
disq, line 29: This can be removed for now, but forcing plugin authors to mark their every-syncing-test to be parallel doesn't sound good.

yevgenypats: I think default sounds good but let's make it more goish with `opts ...TestSourcePluginOption` and introduce `With`

yevgenypats (changes requested):
commented on the interface

plugins/source_testing.go
yevgenypats: logger probably also can go into option

disq:
> commented on the interface

done, had to be slightly different than your normal goish opts in implementation.

plugins/source_testing.go
```diff
@@ -21,14 +20,17 @@ type ResourceTestCase struct {
...
-
-func TestSourcePluginSync(t *testing.T, plugin *SourcePlugin, logger zerolog.Logger, spec specs.Source) {
+func TestSourcePluginSync(t *testing.T, plugin *SourcePlugin, logger zerolog.Logger, spec specs.Source, opts ...TestSourcePluginOption) {
 	t.Helper()
 
+	o := &testSourcePluginOptions{}
+	for _, opt := range opts {
+		opt(o)
+	}
+	if !o.NoParallel {
```
yevgenypats, line 30: Last nit: Let's make this variable called `parallel` (also I think no need to export it so it can be lowercase).
default for `parallel: true` at line 26.

and let's do convention `WithTestSourcePluginNoParallel` - i.e `With` + `Name of the struct/option` + option

yevgenypats (changes requested):
Last nit.

yevgenypats (approved):

PR #294, opened 2022-10-14
feat: Support application level protocol message.

yevgenypats:
This will let us make breaking changes to the protocol while ensuring users get the right message on what todo. e.g update plugin or cli (client).

This is needed in general and also is a pre-requisite for the performance issue in destination plugins that will require breaking change in the destination protocol.


clients/source.go
```diff
@@ -94,26 +99,40 @@ func NewSourceClient(ctx context.Context, registry specs.Registry, path string,
...
 	default:
 		return nil, fmt.Errorf("unsupported registry %s", registry)
 	}
+
+	protocolVersion, err := c.GetProtocolVersion(ctx)
+	if err != nil {
+		return nil, err
+	}
+	if protocolVersion < versions.SourceProtocolVersion {
+		return nil, fmt.Errorf("destination plugin protocol version %d is lower than client version %d. Try updating client", protocolVersion, versions.SourceProtocolVersion)
```
disq, line 114: typo: source

clients/source.go
```diff
@@ -94,26 +99,40 @@ func NewSourceClient(ctx context.Context, registry specs.Registry, path string,
...
+	} else if protocolVersion > versions.SourceProtocolVersion {
+		return nil, fmt.Errorf("destination plugin protocol version %d is higher than client version %d. Try updating destination plugin", protocolVersion, versions.SourceProtocolVersion)
```
disq, line 116: typo: source

yevgenypats:
> Looks like a good first step. Rejecting versions and just suggesting upgrading plugins (and for example never suggesting upgrading the cli) won't cover all cases though. In 0.23.0 we had "[module version negotiation](https://github.com/cloudquery/cloudquery/pull/489)" where each module responded with a list of versions it supports, and the cli chose the maximum common version. Something like that might be in the cards for the future for here too.

agree it might be possible future improvment but also I actually don't think more is that is needed. Because in v1 with go-plugin and with protocol negotiation we had so much bugs that most of the time it even didn't print the correct version, so I would just start with a correct message that always work and see how criticial it is to support multiple verisons in the CLI.

The solution will be not in the sdk anyway because it will be on the cli version to try and use client for older version or for newer version but that's too much for now.

PR #298, opened 2022-10-16
feat: Metrics and Type system

yevgenypats:
While fixing the [PostgreSQL](https://github.com/cloudquery/cloudquery/issues/2462) performance bugs found quite a few issues and rabbit holes we need to take care of (This is the quick fix around order of columns which should be good for now - https://github.com/cloudquery/cloudquery/pull/2887).

Few change are required to write protocol:
- First message should contain tables schema (just like in migration) as this can be useful for writes to improve performance and in general it is not expensive as every message after that can contain only the values (without columns) - this should also save some bandwidth and solve the ordering bug out of the box.
- we have an initial start of stats, metrics but this started to go clunky so I already implement a Stats/Metrics API which clients can now query on their schedule and also we can push those in the future to various places like Prometheus and so on.
- While working on stats, I saw that the `Resolver` function is in the incorrect place so I had to move it to `plugins` as otherwise things like stats/metrics wont be possible to implement.

Because those are major changes in the protocol including incompatible protobuf changes, I changed the application level to 2.

Also found number of concurrency issues and graceful shutdown not working that is fixed in this PR.

Another big missing piece I found that is in this PR:

- Our typing system: Is quite broken: There is no standard interface for types, verification is all around the place. In general it should look something like this - https://github.com/jackc/pgx/tree/master/pgtype but for cqtypes. I ended up reusing pgtype for our use-cases as this library is battle tested and kept the license and the copyright.

Also, Implemented concurrent dfs scheduler to make sure memory is limited by O(h) (h is height of the tree/tables).


yevgenypats:
Closing as this will continue:

https://github.com/cloudquery/plugin-sdk/pull/320
https://github.com/cloudquery/plugin-sdk/pull/318

PR #308, opened 2022-10-21
fix: Silence error log when plugin exits

hermanschaaf:
Right now, whenever a plugin gets interrupted by our own signal, an error is logged, even though this is expected to happen under normal circumstances.

This change checks the exit code of the plugin: if it's -1, this indicates the process was interrupted. This is expected, and then we only log an info line.

I'm not sure whether this will work on Windows systems. If not, the behavior there will be the same as before, so then this would only be an improvement for Unix systems (but I haven't tested on Windows). I played around with using sigterm as a nicer way to stop plugins, but realized that likely also [won't work on Windows](https://github.com/golang/go/issues/46345)

### DIFF
--- a/clients/destination.go
+++ b/clients/destination.go
@@ -137,7 +137,8 @@ func (c *DestinationClient) newManagedClient(ctx context.Context, path string) e
 		defer c.wg.Done()
 		if err := cmd.Wait(); err != nil {
 			if cmd.ProcessState != nil && cmd.ProcessState.ExitCode() == -1 {
-				// process killed by our own signal, this is expected
+				// process interrupted by our own signal, this is expected
+				c.logger.Info().Str("plugin", path).Msg("plugin exited")
 				return
 			}
 			c.cmdWaitErr = err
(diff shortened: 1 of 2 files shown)


erezrokah (approved):
Thanks @hermanschaaf this is much better than before 🚀 

Thinking more about this, I'm wondering if using `Kill` here:
https://github.com/cloudquery/plugin-sdk/blob/b87d9ff1c71463fd28bbd62f543151a2c2a55238/clients/source.go#L279

is the right approach, or we should send an interrupt signal so the plugin can gracefully exit and set the exit code to something we control.

Similar to the approach in https://github.com/golang/go/issues/22757#issuecomment-345009730 (try to gracefully shutdown the process, then do a kill).

Then we could guarantee [any shutdown code](https://github.com/cloudquery/plugin-sdk/blob/b87d9ff1c71463fd28bbd62f543151a2c2a55238/serve/source.go#L138) actually runs.

⬆️ **This comment is out of scope of this PR and something we could consider as a follow up**

hermanschaaf:
Will be closed by https://github.com/cloudquery/plugin-sdk/pull/320

Issue #291, opened 2022-10-13
feat: Add retry logic to plugins download


erezrokah:
Sometimes downloading of plugins fails due to a glitch with GitHub:
https://github.com/cloudquery/cloudquery/actions/runs/3243530449/jobs/5318376371#step:7:9
![image](https://user-images.githubusercontent.com/26760571/195637028-41410d7c-639a-459b-bd99-33057cfe1979.png)

and a simple retry works.

We should add retry logic as to not fail in these cases

cychiang:
Hi @erezrokah I would like to discuss this task with you. I found it a bit difficult to write test cases for downloading an artifact from GitHub. In order to mock a http request, I need to create an interface for `http.Client`, then I can mock a http client for it and start to write test cases for it - it becomes a big task. So I go around the code and what is inside the `download.go` and is basically:

- format url with inputs 
- setup output location and file name
- save file 
- show progressbar

Because create an interface and re-write some code with test cases take times so maybe can go for another way. I'm thinking to improve this part with [go-resty](https://github.com/go-resty/resty) and [httpmock](https://github.com/jarcoal/httpmock) so we can have better handling for downloading files and proper test cases. But not sure is it worth to do it or not. 🤔

erezrokah:
Thanks for the PR @cychiang and pushing this issue forward. I commented on the PR and testing part in https://github.com/cloudquery/plugin-sdk/pull/310#pullrequestreview-1157745326

PR #310, opened 2022-10-25
feat: Add retry logic when downloading plugins from GitHub

cychiang:
#### Summary
This PR is trying to solve #291 

---

- [ ] Update or add tests 🧪

### DIFF
--- a/clients/download.go
+++ b/clients/download.go
@@ -20,6 +20,8 @@ const (
 	PluginTypeSource      PluginType = "source"
 	PluginTypeDestination PluginType = "destination"
 	DefaultDownloadDir               = ".cq"
+	RetryAttempts                    = 5
+	SleepTime                        = 1 * time.Second
 )
 
 func DownloadPluginFromGithub(ctx context.Context, localPath string, org string, name string, version string, typ PluginType) error {
@@ -75,6 +77,21 @@ func DownloadPluginFromGithub(ctx context.Context, localPath string, org string,
 	return nil
 }
 
+func retryDo(req *http.Request) (resp *http.Response, err error) {
+	for i := 0; i < RetryAttempts; i++ {
+		resp, err = http.DefaultClient.Do(req)
+		if resp.StatusCode != http.StatusOK {
+			// Close body before retry
+			resp.Body.Close()
+			time.Sleep(SleepTime)
+			continue
+		} else {
+			return resp, nil
+		}
+	}
+	return resp, fmt.Errorf("failed to download attempt %d times", RetryAttempts)
+}
+
 func downloadFile(ctx context.Context, localPath string, url string) (err error) {
 	// Create the file
 	out, err := os.Create(localPath)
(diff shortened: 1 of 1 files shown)


erezrokah (changes requested):
Thanks for the PR @cychiang! WDYT about using https://github.com/avast/retry-go for the retry logic?

I think it could simplify the code and also avoid the need to write test cases as we rely on an already tested library.

Please let me know that you think

cychiang:
> Thanks for the PR @cychiang! WDYT about using https://github.com/avast/retry-go for the retry logic?
> 
> I think it could simplify the code and also avoid the need to write test cases as we rely on an already tested library.
> 
> Please let me know that you think

Yes, I do also thinking about the library you mentioned, focusing on the retry itself. But [go-resty](https://github.com/go-resty/resty) supports [save response into file](https://github.com/go-resty/resty#save-http-response-into-file) in the other hand might be helpful to simplify the codebase? But if the project would like to support downloading plugins other than GitHub in the future then I think would be good to consider which one would be good in the next step. Maybe I just make this issue more complicate and confuse 😶‍🌫️

erezrokah:
> Yes, I do also thinking about the library you mentioned, focusing on the retry itself. But [go-resty](https://github.com/go-resty/resty) supports [save response into file](https://github.com/go-resty/resty#save-http-response-into-file) in the other hand might be helpful to simplify the codebase? But if the project would like to support downloading plugins other than GitHub in the future then I think would be good to consider which one would be good in the next step. Maybe I just make this issue more complicate and confuse 😶‍🌫️

That's a very good point about `go-resty`. How about we try with [retry-go](https://github.com/avast/retry-go) as that would be a smaller change and will resolve https://github.com/cloudquery/plugin-sdk/issues/291, and open a new issue to consider using `go-resty` so we can discuss the benefits of that approach?

I can open the issue to use `go-resty` or you can do that too and explain the benefits. Please let me know what you prefer

cychiang:
> > Yes, I do also thinking about the library you mentioned, focusing on the retry itself. But [go-resty](https://github.com/go-resty/resty) supports [save response into file](https://github.com/go-resty/resty#save-http-response-into-file) in the other hand might be helpful to simplify the codebase? But if the project would like to support downloading plugins other than GitHub in the future then I think would be good to consider which one would be good in the next step. Maybe I just make this issue more complicate and confuse 😶‍🌫️
> 
> That's a very good point about `go-resty`. How about we try with [retry-go](https://github.com/avast/retry-go) as that would be a smaller change and will resolve #291, and open a new issue to consider using `go-resty` so we can discuss the benefits of that approach?
> 
> I can open the issue to use `go-resty` or you can do that too and explain the benefits. Please let me know what you prefer

Sounds good, let's put `go-resty` to another issue and focus on `retry-go` for the #291 😄

erezrokah (approved):
This looks great, thanks for the follow up @cychiang 👍 

I'll let @yevgenypats comment too

erezrokah:
I'll go ahead and merge (and release this) as it fixes the bug. We can do any follow in future PRs.

One thing I noticed is that the error is printed at the end of all the retry attempts. It would be better to print the attempt on each iteration, which I fixed in https://github.com/cloudquery/plugin-sdk/pull/310/commits/a4a4f5f10be1204c3aaeffb5e7c113315a93df7a, see:

```bash
Loading spec(s) from examples/config-azure.yml
Failed downloading https://github.com/cloudquery/cloudquery/releases/download/plugins-source-azure-v1.2.0/azure_darwin_arm64.zip with status code <code>. Retrying
Failed downloading https://github.com/cloudquery/cloudquery/releases/download/plugins-source-azure-v1.2.0/azure_darwin_arm64.zip with status code <code>. Retrying
Failed downloading https://github.com/cloudquery/cloudquery/releases/download/plugins-source-azure-v1.2.0/azure_darwin_arm64.zip with status code <code>. Retrying
Failed downloading https://github.com/cloudquery/cloudquery/releases/download/plugins-source-azure-v1.2.0/azure_darwin_arm64.zip with status code <code>. Retrying
Failed downloading https://github.com/cloudquery/cloudquery/releases/download/plugins-source-azure-v1.2.0/azure_darwin_arm64.zip with status code <code>. Retrying
Error: failed to get source plugin client for azure: failed to download plugin: failed downloading: https://github.com/cloudquery/cloudquery/releases/download/plugins-source-azure-v1.2.0/azure_darwin_arm64.zip
```

PR #311, opened 2022-10-27
chore: Release 1.0.0

erezrokah:
Release-As: 1.0.0

feat: [Official v1 release](https://www.cloudquery.io/blog/cloudquery-v1-release)

#### Summary

We should move to a major version for the SDK to be consistent with our main repository. See https://github.com/googleapis/release-please#how-do-i-change-the-version-number for the `Release-As` docs

---

- [ ] Update or add tests 🧪


yevgenypats:
Im not sure this should follow CLI versioning but rather Go semantic versioning (though we can start from v1 but very quickly it can become v5 while cli will be v1 as it's mostly affects plugins API and not necessarily the communication between plugin and cli). and every breaking change will go into `/v2`, `/v3` and so on

erezrokah:
> Im not sure this should follow CLI versioning but rather Go semantic versioning (though we can start from v1 but very quickly it can become v5 while cli will be v1 as it's mostly affects plugins API and not necessarily the communication between plugin and cli). and every breaking change will go into `/v2`, `/v3` and so on

Yes I agree we should use go versioning (packages into `v2` `v3` etc). This means that we won't have breaking changes commits (e.g. `fix!` or `feat!`) which will work with this PR.

This PR sets the tag numbering to start from `v1.0.0` so it doesn't prevent us from doing versioning via packages (i.e. if we do a `feat` with `/v2` package it will bump the version to `v1.1.0` as a non breaking feature).

TLDR: The way we version our releases/tags does not depend on how we version the Go packages, and vice versa. This PR just makes it so our tags start with `v1` instead of `v0`.

yevgenypats:
Closing as taken care of [here](https://github.com/cloudquery/plugin-sdk/pull/348/files) and starting from `v2` as not possible in go to start from `v1` as it's `v1` by default.

PR #316, opened 2022-10-27
fix: Increment tableResources

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/3201.

The increment was removed in https://github.com/cloudquery/plugin-sdk/pull/224/files#diff-abecd8d935ca8952a3554e15fa6989972dd0cea6de3410b6541931883b0d6d84R219.

I tried to think of a way to use the `summary` to get each table resources, but it will require some refactoring, as each table's summary is merged with its children so we can't use `summary.Resources` to get a table resources (it contains all nested tables too).

---

- [ ] Update or add tests 🧪

### DIFF
--- a/schema/table.go
+++ b/schema/table.go
@@ -234,6 +234,7 @@ func (t *Table) Resolve(ctx context.Context, meta ClientMeta, parent *Resource, 
 			continue
 		}
 		for i := range objects {
+			tableResources++
 			i := i
 
 			// right now we support concurrency only for objects/resources of parent tables


yevgenypats:
Maybe we should hold off with that as I removed summary and turned it into proper stateful metrics here - https://github.com/cloudquery/plugin-sdk/pull/298

erezrokah:
When do we expect https://github.com/cloudquery/plugin-sdk/pull/298 to be released? The issue this PR fixes was reported by 2 users, so it might be better to fix it soon to avoid confusion and do any follow up improvements later, WDYT?

yevgenypats:
> When do we expect #298 to be released? The issue this PR fixes was reported by 2 users, so it might be better to fix it soon to avoid confusion and do any follow up improvements later, WDYT?

hopefully soon but sure we can merge this pr as long as it wont involve a lot more follow-ups as otherwise it will be more work fixing all of this.

erezrokah:
> hopefully soon but sure we can merge this pr as long as it wont involve a lot more follow-ups as otherwise it will be more work fixing all of this.

I don't think it needs any follow up PRs. This used to work and the increment was remove not so long ago

yevgenypats (changes requested):
I think let's wait a bit as my PR is mostly ready and otherwise I'll have more rebase hell - https://github.com/cloudquery/plugin-sdk/pull/298

erezrokah:
Replaced by https://github.com/cloudquery/plugin-sdk/pull/298

PR #318, opened 2022-10-30
feat: Add Metrics and improve scheduler with DFS

yevgenypats:
Trying to split this PR: https://github.com/cloudquery/plugin-sdk/pull/298 into smaller bits.

Metrics:

- This introduce Metrics as stateful struct as it should be (instead of prop drilling) + GetMetrics Protobuf API

Scheduler:

- I believed I've hit a bug/deadlock in our current scheduler and also for metrics to work I needed to move the scheduler from `schema` to `plugins` so I added that in the same PR.

Currently the user will specify only one variable `concurrency` and the scheduler will decide on how to split it between levels. For simplicity I kept it the same way as before with concurrency for only the first level.

Concurrent DFS will make sure there are no deadlocks and memory is always kept at O(goroutines) and o(h) (where h is height).


plugins/source_scheduler_dfs.go
```diff
@@ -0,0 +1,192 @@
...
+	// This is very similar to the concurrent web crawler problem with some minor changes.
+	// We are using DFS to make sure memory usage is capped at O(h) where h is the height of the tree/subchilds.
+	tableConcurrency := spec.Concurrency / minResourceConcurrency
+	if tableConcurrency < minTableConcurrency {
+		tableConcurrency = minTableConcurrency
+	}
+	resourceConcurrency := tableConcurrency * minResourceConcurrency
+
+	p.tableSem = semaphore.NewWeighted(int64(tableConcurrency))
+	p.resourceSem = semaphore.NewWeighted(int64(resourceConcurrency))
```
erezrokah, line 38: We might want to create a resource semaphore per table instead of sharing it across tables, as this means one branch of the DFS tree can block other branches (e.g. table 1 can block table 2 if it has many resources to handle in the current batch).

plugins/source_scheduler_dfs.go
```diff
@@ -0,0 +1,192 @@
...
+		}
+		wg.Wait()
+	}()
+
+	for resource := range resourcesChan {
+		resolvedResources <- resource
+		if resource.Table.Relations == nil {
+			continue
+		}
+		for _, relation := range resource.Table.Relations {
```
erezrokah, line 139: We should consider parallelizing this part as well (future PR), as I think it would benefit us more at the moment.
I don't think many plugins use `PreResourceResolver` for list/detail pattern yet

plugins/source_test.go
```diff
@@ -35,148 +61,199 @@ func testTable() *schema.Table {
...
-			t.Fatalf("unexpected resource column value: %v", val)
-		}
-	}
-	if err := g.Wait(); err != nil {
-		t.Fatal(err)
+	var i uint64
+	for range resources {
+		// if !tc.data[i].Equal(resource.GetValues()) {
+		// 	t.Fatalf("expected in item %d %v. got %v", i, tc.data[i], resource.GetValues())
+		// }
```
erezrokah, line 244:
```suggestion
		// if !tc.data[i].Equal(resource.GetValues()) {
		// 	t.Fatalf("expected in item %d %v. got %v", i, tc.data[i], resource.GetValues())
		// }
```

serve/destination.go
```diff
@@ -103,7 +103,7 @@ func newCmdDestinationServe(destination *destinationServe) *cobra.Command {
 				err = sentry.Init(sentry.ClientOptions{
 					Dsn:              destination.sentryDSN,
 					Debug:            false,
-					AttachStacktrace: true,
+					AttachStacktrace: false,
```
erezrokah, line 106: This makes sense after reading the docs:
https://docs.sentry.io/platforms/unity/configuration/options/#attach-stacktrace

image

specs/source.go
```diff
@@ -54,17 +53,13 @@ func (s *Source) SetDefaults() {
 		s.Tables = []string{"*"}
 	}
 
-	if s.Concurrency != 0 && s.TableConcurrency == 0 && s.ResourceConcurrency == 0 {
+	if s.TableConcurrency != 0 || s.ResourceConcurrency != 0 {
 		// attempt to make a sensible backwards-compatible choice, but the CLI
-		// should raise a warning about this until the `concurrency` option is fully removed.
-		s.TableConcurrency = s.Concurrency
-		s.ResourceConcurrency = s.Concurrency
+		// should raise a warning about this until the `table_concurrency` and `resource_concurrency` option is fully removed.
```
erezrokah, line 58:
```suggestion
		// should raise a warning about this until the `table_concurrency` and `resource_concurrency` options are fully removed.
```

schema/meta.go
```diff
@@ -5,11 +5,10 @@ import (
 	"time"
 
 	"github.com/google/uuid"
-	"github.com/rs/zerolog"
 )
 
 type ClientMeta interface {
-	Logger() *zerolog.Logger
+	Name() string
```
erezrokah, line 11:
```suggestion
	ID() string
```

Not sure but this should be unique otherwise the log message in `resolveTableDfs` won't make sense correct?
https://github.com/cloudquery/plugin-sdk/pull/318/files#diff-9ba9bcdf8d30be3df89fb7ad1cffbaa5cdd72b28097d7eec3cd1eb5f5dea5e7bR70

Also what's the reason for the removal of `Logger`?

> Changes to `ClientMeta` are technically breaking changes (which is OK).
If we want we can make this a non breaking change by using properties (e.g. `Name string`) and then plugins need to add them via the constructor otherwise the log field will be empty

erezrokah (approved):
This looks good and nothing blocking.

Added a few comments but I don't think any are blocking

plugins/source.go
```diff
@@ -29,8 +26,22 @@ type SourcePlugin struct {
...
+	// resourceSem is a semaphore that limits the number of concurrent resources being fetched
+	resourceSem *semaphore.Weighted
+	// tableSem is a semaphore that limits the number of concurrent tables being fetched
+	tableSem *semaphore.Weighted
+	// maxDepth is the max depth of tables
+	maxDepth uint64
 }
 
+const (
+	maxAllowedDepth = 4
```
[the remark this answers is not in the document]
yevgenypats: I would start with enforcing to make sure users/authors are following best practices of flattening/not-nesting tables. If someone will hit a blocker we can change that.

The issue is that DFS scheduler might become inefficient with big `h` which we want to avoid. and in general using nested tables if possible to use APIs directly makes plugin inefficient unrelated to any scheduler.

plugins/source_metrics.go
```diff
@@ -0,0 +1,104 @@
...
+
+type TableClientMetrics struct {
+	Resources uint64
+	Errors    uint64
+	Panics    uint64
+	StartTime time.Time
+	EndTime   time.Time
+}
+
+func (s *TableClientMetrics) Equal(other *TableClientMetrics) bool {
```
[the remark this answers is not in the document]
yevgenypats: `DeepEqual` is super slow and can have unpredictable results depending on what you are trying to achieve so we want to define equal for every type. DeepEqual is usually used for testing purposes when you don't control or have access to some external struct.

plugins/source_scheduler_dfs.go
yevgenypats: Yeah I think there is much benchmarking to do but for now I'd keep it simple (it is basically the same scheduling as we had before just without the deadlock)

serve/destination.go
yevgenypats: Stacktrace in Go with sentry don't work well at all as it's trying to do some smart things like compiling in the source code which is unnecessary and makes also the even on sentry confusing.

schema/meta.go
yevgenypats: We don't use the logger and we don't need to enforce it on the interface level. the only thing we need is `ID` or `Name` so logs will be easily understood.

plugins/source_metrics.go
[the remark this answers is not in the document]
erezrokah: My concern is that someone adds a field to the struct and then forgets to add it to the `Equal`.
i.e. We'll have to maintain this implementation. Not sure the performance hit is something that can really slow us down.

Where is the `Equal` function used?

plugins/source.go
```diff
@@ -108,62 +143,45 @@ func (p *SourcePlugin) Version() string {
...
-				atomic.AddUint64(&summary.Errors, tableSummary.Errors)
-				atomic.AddUint64(&summary.Panics, tableSummary.Panics)
-			}()
-		}
-	}
-	wg.Wait()
-	logger.Info().Uint64("total_resources", summary.Resources).TimeDiff("duration", time.Now(), startTime).Msg("sync finished")
-	return &summary, nil
+	p.logger.Info().Uint64("resources", p.metrics.TotalResources()).Uint64("errors", p.metrics.TotalErrors()).Uint64("panics", p.metrics.TotalPanics()).TimeDiff("duration", time.Now(), startTime).Msg("sync finished")
+	// this for backward compatibility and will be removed in syncv2 so the way to get the metrics would be separate
```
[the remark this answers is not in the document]
yevgenypats: removed comment. it's old one.

plugins/source_metrics.go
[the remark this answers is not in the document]
yevgenypats: we use it in tests but what it the issue of maintaining this and using a library that is used to compare structs which are not in control of the author?

If you look at `SourceMetrics` the `Equal` function is not that simple comparing the `map` but it ensure it will work and I can also add tests to that to ensure we don't forget to update that.

schema/meta.go
yevgenypats: changed to `ID`

plugins/source_metrics_test.go
```diff
@@ -0,0 +1,37 @@
...
+	other := &SourceMetrics{
+		TableClient: make(map[string]map[string]*TableClientMetrics),
+	}
+	other.TableClient["test_table"] = make(map[string]*TableClientMetrics)
+	other.TableClient["test_table"]["testExecutionClient"] = &TableClientMetrics{
+		Resources: 1,
+		Errors:    2,
+		Panics:    3,
+	}
+	if !s.Equal(other) {
```
erezrokah, line 34: I don't this this test will catch if someone adds a field but forgets to add it to `Equal`

plugins/source_metrics.go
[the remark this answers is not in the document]
erezrokah: If it's only used in the tests I would say we don't need it as a part of the struct and just do the equality in the test (via library or a helper function).

Again this is not blocking for the PR, just seems there's already existing code we can use to compare structs in tests, so we don't need to re-implement it

plugins/source_metrics_test.go
yevgenypats: This wont catch it but this should be caught in code-review. Also reflection can miss things and/or have incorrect results. I think there is no good way to ensure this apart from code-review but I would say this is reasonable for an equality function.

plugins/source_scheduler_dfs.go
yevgenypats: this wont block because unlike `table` semaphores, `resource` semaphore are being released the moment resource is resolved as we use a bit more memory for that (still capped).

erezrokah: Yes, agree it won't block, just might impact concurrency as table 2 won't finish resolving before table 1 is finished.

plugins/source_metrics.go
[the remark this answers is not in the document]
hermanschaaf: +1 to what @erezrokah said; not a blocker, but I'd vote for go-cmp in tests rather than maintain equality operators if we don't need them in our (non-test) code.

hermanschaaf (commented):
I read through, and it all looks good to me; the concurrency also looks good, though I'd like to tweak it a bit further once we have benchmarks in place in the coming weeks 👍

PR #320, opened 2022-10-30
feat: Add CQ type system to support multiple destinations

yevgenypats:
This add a type system to CloudQuery SDK.

This is mandatory to support multiple destinations.

Also, this found quite a few bugs where we were sending random stuff over the wire without any validation apart from maybe when we were hitting the database and then failing batches all together.

In CloudQuery type system I used heavily https://github.com/jackc/pgtype and kept the license and the copyright in it's own package `cqtype`.

This is a continue of this PR https://github.com/cloudquery/plugin-sdk/pull/298 where I split it into this one and https://github.com/cloudquery/plugin-sdk/pull/318


plugins/destination.go
```diff
@@ -79,55 +75,51 @@ func (p *DestinationPlugin) Init(ctx context.Context, logger zerolog.Logger, spe
...
+	return p.client.Read(ctx, table, sourceName, res)
+}
+
+func (p *DestinationPlugin) Write(ctx context.Context, tables schema.Tables, sourceName string, syncTime time.Time, res <-chan *schema.DestinationResource) error {
+	SetDestinationManagedCqColumns(tables)
+	ch := make(chan *schema.DestinationResource)
+	eg := &errgroup.Group{}
+	// given most destination plugins writing in batch we are using a worker pool to write in parallel
+	// it might not generalize well and we might need to move it to each destination plugin implementation.
+	for i := 0; i < writeWorkers; i++ {
```
erezrokah, line 92: Since `writeWorkers` is always `1` maybe we can remove this for now?

plugins/source_test.go
```diff
@@ -237,19 +260,26 @@ func testSyncTable(t *testing.T, tc syncTestCase) {
...
+	for resource := range resources {
+		if tc.data == nil {
+			t.Fatalf("Unexpected resource %v", resource)
+		}
+		if i > len(tc.data) {
+			t.Fatalf("expected %d resources. got %d", len(tc.data), i)
+		}
+		// if !tc.data[i].Equal(resource.GetValues()) {
+		// 	t.Fatalf("expected in item %d %v. got %v", i, tc.data[i], resource.GetValues())
+		// }
```
erezrokah, line 273:
```suggestion
		// if !tc.data[i].Equal(resource.GetValues()) {
		// 	t.Fatalf("expected in item %d %v. got %v", i, tc.data[i], resource.GetValues())
		// }
```

plugins/destination.go
```diff
@@ -2,21 +2,23 @@ package plugins
...
 	"github.com/rs/zerolog"
+	"golang.org/x/sync/errgroup"
 )
 
 type NewDestinationClientFunc func(context.Context, zerolog.Logger, specs.Destination) (DestinationClient, error)
 
 type DestinationClient interface {
 	Migrate(ctx context.Context, tables schema.Tables) error
-	Write(ctx context.Context, table string, data map[string]interface{}) error
+	Read(ctx context.Context, table *schema.Table, sourceName string, res chan<- *schema.DestinationResource) error
```
erezrokah, line 18: Where do we use `Read`?

erezrokah (changes requested):
Added a few comments and will continue to review tomorrow 📖

clients/destination.go
```diff
@@ -188,18 +178,6 @@ func (c *DestinationClient) GetProtocolVersion(ctx context.Context) (uint64, err
 	return res.Version, nil
 }
 
-func (c *DestinationClient) GetMetrics(ctx context.Context) (*plugins.DestinationMetrics, error) {
```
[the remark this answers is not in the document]
yevgenypats: Good catch. seems like an issue in my split PR

clients/source.go
```diff
@@ -122,17 +122,8 @@ func (c *SourceClient) newManagedClient(ctx context.Context, path string) error
...
 	if err := cmd.Start(); err != nil {
-		return fmt.Errorf("failed to start plugin %s: %w", path, err)
+		return fmt.Errorf("failed to start source plugin %s: %w", path, err)
 	}
-
-	c.wg.Add(1)
-	go func() {
-		defer c.wg.Done()
-		if err := cmd.Wait(); err != nil {
-			c.cmdWaitErr = err
```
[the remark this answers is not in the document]
yevgenypats: I think there is no need to do it in a goroutine but rather we can do it in the `Terminate` function synchorunsly and there wont be need to add workgroups and so on

internal/servers/source.go
```diff
@@ -23,7 +23,7 @@ type SourceServer struct {
 
 func (*SourceServer) GetProtocolVersion(context.Context, *pb.GetProtocolVersion_Request) (*pb.GetProtocolVersion_Response, error) {
 	return &pb.GetProtocolVersion_Response{
-		Version: 1,
+		Version: 2,
```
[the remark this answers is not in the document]
yevgenypats: This is fine as this is the error that should happen if an old cli tries to connect to a new plugin. servers always serve one version while new clients can work with old and new version. So we need to release new CLI first that can work with both v1 and v2 and then all plugins. https://github.com/cloudquery/cloudquery/pull/3323

plugins/destination.go
yevgenypats: yeah we can. I was playing with that, wasn't sure what it should be and how it generalize to most plugins.

plugins/destination.go
yevgenypats: This is very important for testing otherwise it produces a lot of boilerplate code in destination plugins. This way we can test round-trip values and data consistency. https://github.com/cloudquery/plugin-sdk/pull/320/files#diff-e6872d0aa4c7b0a55579fa1fd579f89364e68173611f8bf7d574f95c9d8cc65c

clients/source.go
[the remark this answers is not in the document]
erezrokah: I don't think `Terminate` is called if someone runs `go run main.go serve` and then Ctrl+C (e.g. for unmanaged plugins)

internal/servers/source.go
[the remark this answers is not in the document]
erezrokah: Yes, but the error users get is wrong `Try updating source plugin`

clients/destination.go
```diff
@@ -118,17 +116,9 @@ func (c *DestinationClient) newManagedClient(ctx context.Context, path string) e
...
 	if err := cmd.Start(); err != nil {
-		return fmt.Errorf("failed to start plugin %s: %w", path, err)
+		return fmt.Errorf("failed to start destination plugin %s: %w", path, err)
 	}
 
-	c.wg.Add(1)
-	go func() {
-		defer c.wg.Done()
-		if err := cmd.Wait(); err != nil {
-			c.cmdWaitErr = err
```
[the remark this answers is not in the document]
yevgenypats: It is done in the `Terminate` function

clients/destination.go
```diff
@@ -296,24 +305,37 @@ func (c *DestinationClient) Terminate() error {
...
 		if err := c.conn.Close(); err != nil {
-			c.logger.Error().Err(err).Msg("failed to close gRPC connection")
+			c.logger.Error().Err(err).Msg("failed to close gRPC connection to destination plugin")
 		}
+		c.conn = nil
 	}
 	if c.cmd != nil && c.cmd.Process != nil {
-		if err := c.cmd.Process.Kill(); err != nil {
-			c.logger.Error().Err(err).Msg("failed to kill process")
+		if err := c.cmd.Process.Signal(os.Interrupt); err != nil {
```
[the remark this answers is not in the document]
yevgenypats: Fixed.

clients/destination.go
```diff
@@ -296,24 +305,37 @@ func (c *DestinationClient) Terminate() error {
...
+			c.logger.Error().Err(err).Msg("failed to send interrupt signal to destination plugin")
+		}
+		timer := time.AfterFunc(5*time.Second, func() {
+			if err := c.cmd.Process.Kill(); err != nil {
```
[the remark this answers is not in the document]
yevgenypats: we stop the timer if `wait` returns so it wont be a problem

clients/source.go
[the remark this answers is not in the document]
yevgenypats: yeah but even if this runs in a goroutine or in terminate this wont help setting `cmdWaitErr` as it's not used anywhere so there is no need for this goroutine as it just makes code more complex

internal/servers/source.go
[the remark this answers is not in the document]
yevgenypats: Looks like you are right but it's a bug in old CLI (should be the other way around). unfortunately not much we can do anymore about it. 

I fixed it in new cli and and SDK

cqtypes/convert.go
```diff
@@ -0,0 +1,229 @@
...
+	"time"
+)
+
+const (
+	maxUint = ^uint(0)
+	maxInt  = int(maxUint >> 1)
+	minInt  = -maxInt - 1
+)
+
+// var kindTypes map[reflect.Kind]reflect.Type
```
hermanschaaf, line 15: Looks like this and the `init` function below can be removed

cqtypes/convert.go
```diff
@@ -0,0 +1,229 @@
...
+			return nil, false
+		}
+		convVal := refVal.Elem().Interface()
+		return convVal, true
+	}
+
+	return nil, false
+}
+
+// func init() {
```
hermanschaaf, line 212: Looks like this can be removed

cqtypes/inet.go
```diff
@@ -0,0 +1,158 @@
...
+	"strings"
+)
+
+// Network address family is dependent on server socket.h value for AF_INET.
+// In practice, all platforms appear to have the same value. See
+// src/include/utils/inet.h for more information.
+const (
+// defaultAFInet  = 2
+// defaultAFInet6 = 3
+)
```
hermanschaaf, line 17: Looks like this constant block is now unused 🤔

internal/servers/destinations.go
```diff
@@ -56,69 +56,67 @@ func (s *DestinationServer) Migrate(ctx context.Context, req *pb.Migrate_Request
...
+	var tables schema.Tables
+	if err := json.Unmarshal(r.Tables, &tables); err != nil {
+		return fmt.Errorf("write: failed to unmarshal tables: %w", err)
 	}
+	sourceName := r.Source
+	syncTime := r.Timestamp.AsTime()
 
+	eg := errgroup.Group{}
+	eg.Go(func() error {
+		return s.Plugin.Write(context.Background(), tables, sourceName, syncTime, resources)
```
hermanschaaf, line 84: Just trying to understand: why is this write now using `context.Background()` instead of `msg.Context()` ?

internal/servers/destinations.go
```diff
@@ -56,69 +56,67 @@ func (s *DestinationServer) Migrate(ctx context.Context, req *pb.Migrate_Request
...
+			}
+			return fmt.Errorf("failed to receive msg: %w", err)
 		}
-		var resource *schema.Resource
+		var resource *schema.DestinationResource
 		if err := json.Unmarshal(r.Resource, &resource); err != nil {
 			close(resources)
-			wg.Wait()
+			if err := eg.Wait(); err != nil {
+				s.Logger.Error().Err(err).Msg("failed to wait 2")
```
hermanschaaf, line 108:
```suggestion
				s.Logger.Error().Err(err).Msg("failed to wait 3")
```

Looks like the third instance of an error occurring while waiting in this function, we should probably differentiate it in the same way

internal/servers/destinations.go
```diff
@@ -56,69 +56,67 @@ func (s *DestinationServer) Migrate(ctx context.Context, req *pb.Migrate_Request
...
 		}
 		select {
 		case resources <- resource:
-		case <-msg.Context().Done():
+		case <-ctx.Done():
 			close(resources)
-			wg.Wait()
-			return msg.Context().Err()
+			if err := eg.Wait(); err != nil {
+				s.Logger.Error().Err(err).Msg("failed to wait")
```
hermanschaaf, line 117:
```suggestion
				s.Logger.Error().Err(err).Msg("failed to wait 4")
```

internal/servers/destinations.go
hermanschaaf: nit: I don't have good wording, but I think rather than numbers it would be better if we can make each error unique through its description, like `failed to wait after reaching io.EOF`

plugins/destination_testing.go
```diff
@@ -0,0 +1,246 @@
...
+			// 	Type: schema.TypeInetArray,
+			// },
+			// {
+			// 	Name: "cidr",
+			// 	Type: schema.TypeCIDR,
+			// },
+			// {
+			// 	Name: "cidr_array",
+			// 	Type: schema.TypeCIDRArray,
+			// },
```
hermanschaaf, line 82: Should we uncomment these for testing purposes?

plugins/destination_testing.go
```diff
@@ -0,0 +1,246 @@
...
+	// 	panic(err)
+	// }
+	// cidrColumn := &cqtypes.CIDR{}
+	// if err := cidrColumn.Set("192.0.2.1/24"); err != nil {
+	// 	panic(err)
+	// }
+	// cidrArrayColumn := &cqtypes.CIDRArray{}
+	// if err := cidrArrayColumn.Set([]string{"192.0.2.1/24", "192.0.2.1/24"}); err != nil {
+	// 	panic(err)
+	// }
```
hermanschaaf, line 163: Wondering why this is commented out

schema/column.go
```diff
@@ -46,7 +25,7 @@ type Column struct {
 	// Value Type of column i.e String, UUID etc'
 	Type ValueType `json:"type,omitempty"`
 	// Description about column, this description is added as a comment in the database
-	Description string `json:"description,omitempty"`
+	Description string `json:"-"`
```
hermanschaaf, line 28: Why is description being removed from the JSON?

schema/resource.go
```diff
@@ -13,64 +13,138 @@ type Resources []*Resource
...
+// and panics in case it fails
+func (r *Resource) Set(columnName string, value interface{}) error {
+	index := r.Table.Columns.Index(columnName)
+	if index == -1 {
+		// we panic because we want to distinguish between code error and api error
+		// this also saves additional checks in our testing code
+		panic(columnName + " column not found")
+	}
+	if err := r.data[index].Set(value); err != nil {
+		panic(fmt.Errorf("failed to set column %s: %w", columnName, err))
```
hermanschaaf, line 146: Should this be a panic? It seems like if we set a column and it fails, we should return an error but otherwise leave it empty.

serve/source.go
```diff
@@ -85,6 +91,9 @@ func newCmdSourceServe(source *sourceServe) *cobra.Command {
 					return fmt.Errorf("failed to listen %s:%s: %w", network, address, err)
 				}
 			}
+			// source plugins can only accept one connection at a time
+			// unlike destinatino plugin that can accept multiple connections
```
hermanschaaf, line 95:
```suggestion
			// unlike destination plugins that can accept multiple connections
```

hermanschaaf (changes requested):

plugins/source_test.go
hermanschaaf: Was this an accidental suggestion @erezrokah, or is there something different between the two versions that I'm not seeing? 👓

erezrokah: Ah, I meant to remove those if not used, good catch @hermanschaaf

internal/servers/destinations.go
yevgenypats: Good catch. I think I just forgot this from some of my debugging playground

clients/source_terminate_windows.go
```diff
@@ -0,0 +1,20 @@
...
+	if err := c.cmd.Process.Kill(); err != nil {
+		c.logger.Error().Err(err).Msg("failed to kill source plugin")
+	}
+	st, err := c.cmd.Process.Wait()
+	if err != nil {
+		return err
+	}
+	if !st.Success() {
+		// on windows there is no way to shutdown gracefully via signal. Maybe we can do it via grpc api?
+		// though it is a bit strange to expose api to shutdown a server :thinking?:
```
hermanschaaf, line 15: This sounds like a reasonable way to do it, in the absence of OS support. But let's kick that can down the road for another day

clients/destination.go
```diff
@@ -195,7 +186,7 @@ func (c *DestinationClient) GetMetrics(ctx context.Context) (*plugins.Destinatio
 	}
 	var stats plugins.DestinationMetrics
 	if err := json.Unmarshal(res.Metrics, &stats); err != nil {
-		return nil, fmt.Errorf("failed to unmarshal destination metrics: %w", err)
+		return nil, fmt.Errorf("failed to unmarshal source stats: %w", err)
```
hermanschaaf, line 189:
```suggestion
		return nil, fmt.Errorf("failed to unmarshal destination metrics: %w", err)
```

This was right before, wasn't it?

plugins/source_testing.go
```diff
@@ -101,53 +93,23 @@ func validateResources(t *testing.T, resources []*schema.Resource) {
...
 			}
-
-			columnsWithValues[columnName] = true
-
-			switch resource.Table.Columns.Get(columnName).Type {
-			case schema.TypeJSON:
-				switch value.(type) {
-				case string, []byte:
-					t.Errorf("table: %s JSON column %s is being set with a string or byte slice. Either the unmarhsalled object should be passed in, or the column type should be changed to string", resource.Table.Name, columnName)
-					continue
```
hermanschaaf, line 129: This check has caught a number of bugs already, but seems like it's missing from the new version. Does the JSON type disallow strings / bytes being passed to it?

yevgenypats: our new type system knows to take care of that so it tries to unmarshal and if it doesn't work it panics and if not then then it just stores it and the user doesn't need to take care of that.

hermanschaaf: Nevermind - I checked the JSON type definition now and it looks like it's handled there 👍  (It checks if the type is string or byte and then doesn't call json.Marshal on it)

hermanschaaf: @yevgenypats Nice! 👍

schema/column.go
yevgenypats: Just saving some bandwidth, I dont think we use it anywhere. but I can get it back.

plugins/destination_testing.go
yevgenypats: added back

plugins/destination_testing.go
yevgenypats: added back

plugins/source_test.go
yevgenypats: added back

schema/column.go
hermanschaaf: Yeah not a big deal, like you say, I don't think we are using this anywhere right now.

cqtypes/inet.go
yevgenypats: yes. removed. it is used in pgx to encode to postgresql binary format but we don't need it as we just need the decode part and not from postgresql

internal/servers/destinations.go
yevgenypats: Good catch. I think just got confused in the context maze

hermanschaaf (approved):

PR #336, opened 2022-11-01
fix: Parsing timestamptz default string

yevgenypats:
Another edge-case

### DIFF
--- a/cqtypes/timestamptz.go
+++ b/cqtypes/timestamptz.go
@@ -12,6 +12,7 @@ import (
 
 // this is the default format used by time.Time.String()
 const defaultStringFormat = "2006-01-02 15:04:05.999999999 -0700 MST"
+
 // const microsecFromUnixEpochToY2K = 946684800 * 1000000
 
 const (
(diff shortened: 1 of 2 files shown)


cqtypes/timestamptz.go
```diff
@@ -121,6 +121,11 @@ func (dst *Timestamptz) DecodeText(src []byte) error {
 	default:
 		var tim time.Time
 		var err error
+
+		if len(sbuf) > len(defaultStringFormat)+1 && sbuf[len(defaultStringFormat)+1] == 'm' {
+			sbuf = sbuf[:len(defaultStringFormat)]
+		}
```
hermanschaaf, line 127: Can we add a test for this case?

disq:
What's the edge case, can you comment in the code? is `2006-01-02 15:04:05.999999999 -0700 MSTm` valid?

cqtypes/timestamptz.go
yevgenypats: Added

yevgenypats:
> What's the edge case, can you comment in the code? is `2006-01-02 15:04:05.999999999 -0700 MSTm*` valid?

It's not valid, if you look at `time.String()` they just added manually so when you parse back you have to cut it manually as well

hermanschaaf (approved):

PR #339, opened 2022-11-03
fix: Check for missing parent tables

disq:
Previously this didn't work because `p.tables` does not contain child tables.

### DIFF
--- a/plugins/source.go
+++ b/plugins/source.go
@@ -134,8 +134,8 @@ func filterParentTables(tables schema.Tables, filter []string) schema.Tables {
 		return tables
 	}
 	for _, name := range filter {
-		if tables.Get(name) != nil {
-			res = append(res, tables.Get(name))
+		if t := tables.Get(name); t != nil {
+			res = append(res, t)
 		}
 	}
 	return res
--- a/plugins/source_validate_test.go
+++ b/plugins/source_validate_test.go
@@ -83,12 +83,6 @@ func TestSourcePlugin_listAndValidateAllResources(t *testing.T) {
 			configurationTables: []string{"sub_table"},
 			wantErr:             true,
 		},
-		{
-			name:                "should return both tables if both child and parent are specified",
-			plugin:              SourcePlugin{tables: []*schema.Table{{Name: "main_table", Relations: []*schema.Table{{Name: "sub_table", Parent: &schema.Table{Name: "main_table"}}}}}},
-			configurationTables: []string{"main_table", "sub_table"},
-			want:                []string{"main_table", "sub_table"},
-		},
 	}
 
 	for _, tt := range tests {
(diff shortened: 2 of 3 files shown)


erezrokah (approved):

plugins/source.go
```diff
@@ -134,8 +134,8 @@ func filterParentTables(tables schema.Tables, filter []string) schema.Tables {
 		return tables
 	}
 	for _, name := range filter {
-		if tables.Get(name) != nil {
-			res = append(res, tables.Get(name))
+		if t := tables.Get(name); t != nil {
```
disq, line 137: Why do `tables.Get` twice?

hermanschaaf (approved):
Nice fix 👍

disq:
> @disq Since it's a bit tricky, it might be worth adding a test for this case to `source_validate_test.go` ?

yeah as an afterthought I'll fix the test in the test file.

disq:
> @disq Since it's a bit tricky, it might be worth adding a test for this case to `source_validate_test.go` ?

@hermanschaaf done

hermanschaaf (changes requested):
Wait let me just check something...

plugins/source_validate_test.go
```diff
@@ -79,21 +79,27 @@ func TestSourcePlugin_listAndValidateAllResources(t *testing.T) {
...
 		{
 			name:                "should return an error if child table is without its parent",
-			plugin:              SourcePlugin{tables: []*schema.Table{{Name: "table 1", Parent: &schema.Table{Name: "table 2"}}, {Name: "table 2"}}},
-			configurationTables: []string{"table 1"},
+			plugin:              SourcePlugin{tables: []*schema.Table{{Name: "main_table", Relations: []*schema.Table{{Name: "sub_table", Parent: &schema.Table{Name: "main_table"}}}}}},
+			configurationTables: []string{"sub_table"},
 			wantErr:             true,
 		},
+		{
+			name:                "should return both tables if both child and parent are specified",
```
hermanschaaf, line 87: Yeah hmm, actually I think this should either:
 - return an error if a child table is specified, or
 - remove the child table from the list as long as the parent is specified, since it will get fetched anyway

I don't think we want to return the child table, since if we try to resolve it on its own the parent value will be `nil`.

disq: Removed that test case as this seems like another issue.

hermanschaaf: Yeah, you're right, looks like this is a separate issue. But we should address it, otherwise we'll probably get a panic with some configurations.

hermanschaaf (approved):

plugins/source_validate_test.go
hermanschaaf: Created an issue to track it: https://github.com/cloudquery/plugin-sdk/issues/340

disq: I never said we shouldn't address it.

hermanschaaf: Neither did I try to imply that, sorry if it came across that way :bow:

PR #342, opened 2022-11-03
fix: Disallow child tables

### LINKED ISSUE #340: listAndValidateTables should not return child tables
`listAndValidateTables` can in some circumstances return child tables right now, and this will probably cause a panic. We should either disallow child tables from being listed in the config, or filter the child tables out before returning the list if their parents are included (or both)

See comment here for details: https://github.com/cloudquery/plugin-sdk/pull/339#discussion_r1012748251_

disq:
Fixes #340

Also displays the name of the main table to include.

### DIFF
--- a/plugins/source_validate.go
+++ b/plugins/source_validate.go
@@ -77,15 +77,16 @@ func (p *SourcePlugin) listAndValidateTables(tables, skipTables []string) ([]str
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


hermanschaaf (approved):
Nice, yeah I think this is good enough for now, we can revisit filtering of child tables as a separate feature later

yevgenypats (approved):

Issue #309, opened 2022-10-25
feat: Ability to dump to supported tables by a plugin in a machine readable way


erezrokah:
In order to help with people building orchestration on top of CloudQuery, it would be nice to be able to dump all supported tables in a machine readable way as the [tables readme](https://github.com/cloudquery/cloudquery/blob/b3a18cbbf79c234fe659bbdc63d64eb50dee98b3/plugins/source/aws/docs/tables/README.md) is a bit hard to parse using code (or copy paste even).

We could add a new command or add a `--json` flag to the docs command to output the result as a JSON file

erezrokah:
> Can we start with documentation here? maybe add a `json` file as part of our docs generation ?

💯 That's what's the issue suggests

erezrokah:
@hermanschaaf suggested this should be a CLI command and not a plugin command (which is the current state), for example `cloudquery tables <plugin-name>`.

I ⬆️ should make it easier for users to consume the tables (otherwise they first need to download the plugins manually or via sync)

PR #347, opened 2022-11-04
feat: Add option to plugin doc command to output tables as JSON

hermanschaaf:
Fixes https://github.com/cloudquery/plugin-sdk/issues/309

This adds a `--format json` option to the source plugin `doc` command, which allows all tables to be dumped as a single JSON file. This can then be further parsed with `jq` for various purposes (such as comparing two versions of a plugin to see what changed). 

### Example: List first 5 top-level tables of AWS plugin

```
./aws doc --format json .
```

```
$ cat __tables.json | jq -r '.[].name' | head -n 5
aws_accessanalyzer_analyzers
aws_acm_certificates
aws_apigateway_api_keys
aws_apigateway_client_certificates
aws_apigateway_domain_names
```


plugins/source_docs.go
```diff
@@ -14,20 +15,101 @@ import (
...
+	Description string `json:"description"`
+}
+
+func (p *SourcePlugin) renderTablesAsJSON(dir string) error {
+	tables := p.jsonifyTables(p.Tables())
+	b, err := json.MarshalIndent(tables, "", "  ")
+	if err != nil {
+		return fmt.Errorf("failed to marshal tables as json: %v", err)
+	}
+	outputPath := filepath.Join(dir, "__tables.json")
```
hermanschaaf, line 81: I made this name start with a double-underscore, so that we have the option of dumping a json file for each table in the future without having conflicts with table names.

disq: Why not output to stdout?

hermanschaaf: Hmm, that's also an option, but the `doc` command takes a directory as a required argument. So if we do that, we should probably put this under a different command entirely (maybe `json`?)

disq: Mmmm, then it only makes sense to write to a file, to not break consistency.

plugins/source_docs.go
```diff
@@ -14,20 +15,98 @@ import (
...
+type jsonTable struct {
+	Name        string       `json:"name"`
+	Description string       `json:"description"`
+	Columns     []jsonColumn `json:"columns"`
+	Relations   []jsonTable  `json:"relations"`
+}
+
+type jsonColumn struct {
+	Name string `json:"name"`
+	Type string `json:"type"`
```
disq, line 70: Maybe add `IsPK bool` or something to that effect?

hermanschaaf: Ah yes, good call 👍

disq (approved):

plugins/source_docs.go
```diff
@@ -14,20 +15,100 @@ import (
...
+var SourceDocsFormats = []SourceDocsFormat{
+	SourceDocsFormatMarkdown,
+	SourceDocsFormatJSON,
+}
+
+func (s SourceDocsFormat) String() string {
+	return string(s)
+}
+
+func (s SourceDocsFormat) Validate() error {
```
yevgenypats, line 34: I think this can be just an enum (https://github.com/cloudquery/plugin-sdk/blob/main/serve/source.go#L67) which already include the validation.

plugins/source_docs.go
```diff
@@ -14,20 +15,100 @@ import (
...
+	case SourceDocsFormatMarkdown:
+		return p.renderTablesAsMarkdown(dir)
+	case SourceDocsFormatJSON:
+		return p.renderTablesAsJSON(dir)
+	default:
+		return fmt.Errorf("unsupported format: %v", format.String())
+	}
+}
+
+type jsonTable struct {
```
yevgenypats, line 61: I think this is fine but can't we re-use the same tables definition that we already have? we also encode them to json or we want a different structure.

yevgenypats (changes requested):
Sorry for the delayed review, forgot about it. It looks good. One nit and one question.

plugins/source_docs.go
hermanschaaf: I tried this initially, and if I recall correctly the issue I faced was that column type came out as an integer, which isn't very useful without the mappings. So I figured that it would be good to decouple the GRPC representation from the one we use here, but I could be convinced that we shouldn't do that, and figure out a way to turn the type to a string here instead?

plugins/source_docs.go
hermanschaaf: Good point, but I think the enum code is private to the `serve` package so I'll have to refactor a bit first.

plugins/source_docs.go
yevgenypats: Yeah I don't have a good idea right now so maybe we can merge this as is and maybe fix later on if it becomes a maintains burden. But actually now that I think about it why the gRPC representation is not good ? especially if it's intended to be read by computers? Maybe we can have some flag that will marshal the `Type` in a different way.

yevgenypats:
@hermanschaaf I think we can do just the enum and we can merge it/roll this out so user can use it.

plugins/source_docs.go
hermanschaaf: I think one use case for this is going to be outputting the schema in a different format than the markdown we support. Maybe a different markdown format, maybe HTML? But having a number for the type is not very useful then, you'd either need access to the plugin-sdk or copy-paste the list of constants to know how it maps to a string. I also think it's probably good if we can evolve the internal representations separately. We may even move away from JSON for internal communication purposes if it turns out to be a drag on performance, but we'd still want to keep this.

hermanschaaf:
@yevgenypats Updated to use enums, PTAL

plugins/.snapshots/TestGenerateSourcePluginDocs-JSON-__tables.json
```diff
@@ -0,0 +1,96 @@
...
+        "type": "TypeInt",
+        "is_primary_key": true
+      },
+      {
+        "name": "id_col2",
+        "type": "TypeInt",
+        "is_primary_key": true
+      }
+    ],
+    "relations": [
```
erezrokah, line 37: Sorry for jumping late to this, but can we also have the relations at the top level? Might be easier to process it that way. For example if I was to get a list of all tables, I'll need to recurse into the JSON at the moment.

We can add a another property `parent` to link to the parent table, and keep it `nil` for top level ones, WDYT?

hermanschaaf: Yeah, I was in two minds about it 🤔  I think having it flattened will be good for some use cases, and not-so-good for others. I'm fine with changing it to flattened though. @yevgenypats WDYT?

erezrokah: Maybe we can have both?

hermanschaaf: why not both

yevgenypats (approved):

PR #348, opened 2022-11-04
fix: Make CQTypes to Schema and introduce versioning.

yevgenypats:
This make the following changes:

- moves cqtype package into schema because we need the original `ValueType` reside inside the same package as the types. Given I didn't want to update all the plugins I moved the types into `schema` and not the otherway around
- introduces semantic versioning for sdk and protobuf. This starts from v2 (because in go v1 is the same as no version so not possible to have a `/v1`)
- Streamlines testing for destination plugins
- Introduce `Transformer` and `ReverseTransformer` for destination plugins. This way it is easy for the SDK to have round trip tests with those transformers and the `Read` capability.


codegen/transformers.go
```diff
@@ -43,10 +43,6 @@ func defaultGoTypeToSchemaType(v reflect.Type) (schema.ValueType, error) {
 		return schema.TypeBool, nil
 	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64,
 		reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64:
-		// need to check for duration here, as Go time.Duration is int64
-		if v == reflect.TypeOf(time.Duration(0)) {
-			return schema.TypeTimeInterval, nil
-		}
```
hermanschaaf, line 49: Oh I see it's also removed here. Could this break an existing plugin by any chance?

schema/types.go
```diff
@@ -15,77 +14,241 @@ const (
...
+	if err := json.Unmarshal(b, &res); err != nil {
+		return err
+	}
+	cqTypes := make(CQTypes, 0, len(res))
+	for i := range res {
+		var t ValueType
+		if err := json.Unmarshal(res[i]["type"], &t); err != nil {
+			return err
+		}
+		if t <= TypeInvalid || t >= TypeEnd {
```
hermanschaaf, line 158: Interesting technique to check `t >= TypeEnd`, but I like it 😄

hermanschaaf (approved):
LGTM 👍 Only thing is that we should probably mark this as `fix!` so that the version gets bumped to the next major version?

erezrokah (approved):
What @hermanschaaf said 👍

schema/types.go
disq: it's very C (`MAX_TYPE` would be the C counterpart)

internal/pb/source.proto
```diff
@@ -15,48 +14,22 @@ service Source {
   rpc GetVersion(GetVersion.Request) returns (GetVersion.Response);
   // Get all tables the source plugin supports
   rpc GetTables(GetTables.Request) returns (GetTables.Response);
-  // GetSyncSummary returns the latest sync summary of the source plugin. we don't want to send the summary on
-  // every sync request.
-  rpc GetSyncSummary(GetSyncSummary.Request) returns (GetSyncSummary.Response);
-  // Fetch resources
-  rpc Sync(Sync.Request) returns (stream Sync.Response);
   // Sync2 is a new sync API that supports CQ Types. It is not backward compatible with Sync.
```
disq, line 17: Doc looks a bit outdated, should explain by itself not referring to older sync

yevgenypats:
Closing in favour of https://github.com/cloudquery/plugin-sdk/pull/369

Issue #350, opened 2022-11-06
Bug(config): Can't use multiple sources in the same sync


erezrokah:
### Describe the bug

With the following config:

```yml
kind: source
spec:
  name: aws
  version: v4.6.1
  destinations: [postgresql]
  path: cloudquery/aws
---
kind: source
spec:
  name: azure
  version: v1.3.3
  destinations: [postgresql]
  path: cloudquery/azure
---
kind: destination
spec:
  name: postgresql
  version: v1.6.3
  spec:
    connection_string: postgresql://postgres:pass@localhost:5432/postgres
```

The CLI errors out with `Error: failed to load spec(s) from <path>. Error: destination postgresql is used by multiple sources cloudquery/azure with different versions`

### Expected Behavior

Using multiple sources should work

### Steps to Reproduce

Create a YAML file with the above configuration, run `cloudquery sync <file>`

### Possible Solution

Fix the config validation

### Provider and CloudQuery version

v1.5.5

PR #353, opened 2022-11-06
fix: Include source path in dest to source map key

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/plugin-sdk/issues/350

---

- [ ] Update or add tests 🧪

### DIFF
--- a/specs/spec_reader.go
+++ b/specs/spec_reader.go
@@ -83,9 +83,10 @@ func (r *SpecReader) validate() error {
 			if r.Destinations[destination] == nil {
 				return fmt.Errorf("source %s references unknown destination %s", source.Name, destination)
 			}
-			if destinationSourceMap[destination] == "" {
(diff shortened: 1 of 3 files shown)


specs/spec_reader.go
```diff
@@ -83,9 +83,10 @@ func (r *SpecReader) validate() error {
 			if r.Destinations[destination] == nil {
 				return fmt.Errorf("source %s references unknown destination %s", source.Name, destination)
 			}
-			if destinationSourceMap[destination] == "" {
-				destinationSourceMap[destination] = source.Path + "@" + source.Version
-			} else if destinationSourceMap[destination] != source.Path+"@"+source.Version {
+			destinationToSourceKey := fmt.Sprintf("%s-%s", destination, source.Path)
```
erezrokah, line 86: It's ok to `destination` name and not `destination.Path` as those destination names are unique by definition in a single sync

yevgenypats (approved):

PR #357, opened 2022-11-07
fix: Move cqtypes

yevgenypats:
This is a split of this PR https://github.com/cloudquery/plugin-sdk/pull/348 so it will be easier to review. 

Also, Im going to push the move to v2 maybe another week as we dont really have anything that requires protocol changes for now. even though it is the right way to do it and will make the cli much easier including the need to things like that - https://github.com/cloudquery/cloudquery/pull/3657 as it will be all handles on protobuf and go versioning side

### DIFF
--- a/schema/LICENSE
+++ b/schema/LICENSE
@@ -1,0 +1,23 @@
+Copyright (c) 2013-2021 Jack Christensen
+Copyright (c) 2022 CloudQuery inc.
+
+MIT License
+
+Permission is hereby granted, free of charge, to any person obtaining
+a copy of this software and associated documentation files (the
+"Software"), to deal in the Software without restriction, including
+without limitation the rights to use, copy, modify, merge, publish,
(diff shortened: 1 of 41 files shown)


disq:
Looks like `schema` package has way too many files now? Maybe prefix files coming from cqtypes with something? `type_*.go`?

hermanschaaf (approved):
I actually ended up reviewing [the other PR](https://github.com/cloudquery/plugin-sdk/pull/348) first in any case :+1: I'll leave it up to you how you want to merge the changes.

yevgenypats:
> I actually ended up reviewing [the other PR](https://github.com/cloudquery/plugin-sdk/pull/348) first in any case 👍 I'll leave it up to you how you want to merge the changes.

Thanks! I'll go with that first and then open a new one on top as I ended up making it backward compatible and not breaking the protocol.

PR #363, opened 2022-11-07
fix: Allow managed clients to disable sentry logging

hermanschaaf:
This change allows managed clients to disable sentry logging, which until now was enabled for plugins by accident (so panics in plugins were being logged when they shouldn't have been). The CLI will be able to use this option to disable sentry logging. 

Plugins will now also respect the `CQ_TELEMETRY_LEVEL` environment variable.


serve/source.go
```diff
@@ -63,6 +64,13 @@ func newCmdSourceServe(source *sourceServe) *cobra.Command {
 	var noSentry bool
 	logLevel := newEnum([]string{"trace", "debug", "info", "warn", "error"}, "info")
 	logFormat := newEnum([]string{"text", "json"}, "text")
+	telemetryLevel := newEnum([]string{"none", "errors", "stats", "all"}, "all")
```
yevgenypats, line 67: Im abit confused. we don't have telemetry on the plugin side. Can we just add no-sentry? or just have `none/errors`

yevgenypats: Actually seems like there is already a `no-sentry` flag, so maybe let just use that?

yevgenypats (changes requested):
Looks good. few minor suggestions.

serve/source.go
hermanschaaf: Yes, this PR is re-using the existing `no-sentry` flag, and this isn't adding a new flag. But what is changing is that the plugins now respect `CQ_TELEMETRY_LEVEL`, so that developers who have that set in their environment don't accidentally send crash reports if they forget to run a plugin with the `--no-sentry` flag.

clients/source.go
```diff
@@ -61,6 +62,13 @@ func WithSourceGRPCConnection(userConn *grpc.ClientConn) func(*SourceClient) {
 	}
 }
 
+func WithSourceNoSentry() func(*SourceClient) {
+	return func(c *SourceClient) {
+		// we use a different variable here because we don't want to close a connection that wasn't created by us.
```
hermanschaaf, line 67:
```suggestion
```

clients/destination.go
```diff
@@ -35,6 +35,7 @@ type DestinationClient struct {
 	conn           *grpc.ClientConn
 	grpcSocketName string
 	wg             *sync.WaitGroup
+	noSentry       bool
```
hermanschaaf, line 38:
```suggestion
	noSentry       bool
	wg             *sync.WaitGroup
```

yevgenypats (approved):

yevgenypats (approved):

PR #366, opened 2022-11-07
fix: Revert "chore: Start SDK semantic versioning from v1"

hermanschaaf:
Reverts cloudquery/plugin-sdk#362

I think the change to use v1 is invalid, as Go expects versioning to be used only from v2 onwards. When importing it in `cloudquery/cloudquery`, I'm getting this error:

```
require github.com/cloudquery/plugin-sdk/v1: version "v1.0.1" invalid: malformed module path "github.com/cloudquery/plugin-sdk/v1"
```


erezrokah (approved):

erezrokah:
TODO after this:
1. Delete the v1 releases https://github.com/cloudquery/plugin-sdk/releases
2. Delete the v2 tags https://github.com/cloudquery/plugin-sdk/tags
3. Remove the `autorelease` labels from https://github.com/cloudquery/plugin-sdk/pull/361 and https://github.com/cloudquery/plugin-sdk/pull/365 so they don't get re-released

yevgenypats (approved):

yevgenypats:
What are the v2 tags? I dont see any.

Regarding deleting the previous ones. I think no need as no one will import them and we just bump a minor version to v1.0.2 and all should work. 

starting v2 in any case we need to have `/v2` and it has to match `v2.x.z`

Issue #256, opened 2022-10-07
Detach githb path generation from `name` of the plugin


amanenk:
Make `path` argument required for plugin configuration. Stop generating default path from name field

Issue #317, opened 2022-10-30
Proposal: Use a dedicated `id` property instead of forcing a unique `name` 


erezrokah:
## Background

To support `overwrite-delete-stale` we've added a constraint for plugins names to be unique.
Also as a result we're planning to make the `path` required, as to allow users to change the `name`. 

## Problem

I believe it's quite [hard to communicate](https://github.com/cloudquery/cloudquery/pull/3276) the reason `name` has to be unique, and easy for users to miss that requirement (also if users split their config across multiple machines we can't really validate it either).

Additionally since we're planning to make `path` required due to the uniqueness of `name`, it adds friction for people onboarding CloudQuery.

## Proposal

Instead of making `name` unique we can add a dedicated `id` configuration (it can default to `name`) and use that instead.
As a result we can achieve the following:

1. Keep the meaning of `name` and `path` the same
2. Keep `path` optional, as we can resolve it from `name`
3. Users that don't need to run CloudQuery in parallel don't need to know anything about uniqueness
4. Easier to understand why an `id` property needs to be unique

TLDR: Only users that split their configuration (e.g. run in parallel) should care about having a unique identifier.

erezrokah:
After an internal discussion we've decided to close this issue and stick with making `path` and `name` unique.
The reason is that are other use cases for needing a field to uniquely identify multiple sources with the same path in a single sync:
1. Distinguish them in the logs
2. Distinguish them in the database

So the meaning of `name` is not actually the plugin name, but the name of the configuration for that source, hence it should be decoupled from path, and path should not infer from name.

PR #368, opened 2022-11-07
fix: Make path a required config parameter

hermanschaaf:
This updates (and replaces) this earlier PR for making `path` a required config parameter: https://github.com/cloudquery/plugin-sdk/pull/265

We have received numerous user reports of the distinction between name and path being confusing, so rather than infer path from name in some circumstances (and not in some others), this change makes things consistent by always requiring both. I've also added a small hint for users running into this error when upgrading from earlier versions.

We will not consider this a breaking SDK change, as it doesn't change any protocols, but it will be a breaking CLI change.

- relates to https://github.com/cloudquery/plugin-sdk/issues/317 
- closes https://github.com/cloudquery/plugin-sdk/issues/256

### DIFF
--- a/specs/source.go
+++ b/specs/source.go
@@ -5,6 +5,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"strings"
+
+	"github.com/thoas/go-funk"
 )
 
 const (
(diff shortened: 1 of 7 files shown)


erezrokah (approved):

specs/source.go
```diff
@@ -79,6 +75,15 @@ func (s *Source) Validate() error {
 	if s.Name == "" {
 		return fmt.Errorf("name is required")
 	}
+	if s.Path == "" {
+		msg := "path is required"
+		// give a small hint to help users transition from the old config format that didn't require path
+		officialPlugins := []string{"aws", "azure", "gcp", "digitalocean", "github", "heroku", "k8s", "okta", "terraform", "cloudflare"}
+		if funk.ContainsString(officialPlugins, s.Name) {
+			msg += fmt.Sprintf(". Hint: try setting path to cloudquery/%s in your config", s.Name)
```
erezrokah, line 83: giphy

yevgenypats (approved):

PR #369, opened 2022-11-07
feat: Add Testing suite for destination plugins

yevgenypats:
Closes https://github.com/cloudquery/plugin-sdk/pull/348

Introduce testing suite for destination plugins as well as number of methods to including `transformers` and `reverseTransformers` to help testing suite be able to test "end-to-end".

This also adds more testing with a `memoryDB` plugin to test destination plugin and destination plugin testing suite


plugins/destination_testing.go
```diff
@@ -194,52 +223,205 @@ func TestData() schema.CQTypes {
...
+	if !s.skipTestOverWriteDeleteStale {
+		if err := p.DeleteStale(ctx, tables, sourceName, syncTime); err != nil {
+			return fmt.Errorf("failed to delete stale data: %w", err)
+		}
 	}
-	if readErr != nil {
-		return readErr
+
+	if err := p.writeAll(ctx, tables, sourceName, syncTime, resources); err != nil {
+		return fmt.Errorf("failed to write one: %w", err)
```
hermanschaaf, line 274:
```suggestion
	if err := p.writeAll(ctx, tables, sourceName, syncTime, resources); err != nil {
		return fmt.Errorf("failed to write all: %w", err)
```

plugins/destination_testing.go
```diff
@@ -194,52 +223,205 @@ func TestData() schema.CQTypes {
...
+	if len(resourcesRead) != 2 {
+		return fmt.Errorf("expected 2 resource, got %d", len(resourcesRead))
+	}
+
+	if resource.Data.Equal(resourcesRead[0]) {
+		return fmt.Errorf("expected data to be %v, got %v", resource.Data, resourcesRead[0])
+	}
+
+	if resource.Data.Equal(resourcesRead[1]) {
+		return fmt.Errorf("expected data to be %v, got %v", resource.Data, resourcesRead[0])
```
hermanschaaf, line 391:
```suggestion
		return fmt.Errorf("expected data to be %v, got %v", resource.Data, resourcesRead[1])
```

plugins/destination_testing.go
```diff
@@ -194,52 +223,205 @@ func TestData() schema.CQTypes {
...
+		return fmt.Errorf("failed to delete stale data: %w", err)
+	}
+
+	resourcesRead, err := p.readAll(ctx, tables[0], sourceName)
+	if err != nil {
+		return fmt.Errorf("failed to read all second time: %w", err)
+	}
+
+	if len(resourcesRead) != 2 {
+		return fmt.Errorf("expected 2 resource, got %d", len(resourcesRead))
```
hermanschaaf, line 383:
```suggestion
		return fmt.Errorf("expected 2 resources, got %d", len(resourcesRead))
```

plugins/destination_testing.go
```diff
@@ -194,52 +223,205 @@ func TestData() schema.CQTypes {
...
+	}
+
+	if resource.Data.Equal(resourcesRead[0]) {
+		return fmt.Errorf("expected data to be %v, got %v", resource.Data, resourcesRead[0])
+	}
+
+	if resource2.Data.Equal(resourcesRead[1]) {
+		return fmt.Errorf("expected data to be %v, got %v", resource.Data, resourcesRead[0])
```
hermanschaaf, line 314:
```suggestion
		return fmt.Errorf("expected data to be %v, got %v", resource.Data, resourcesRead[1])
```

plugins/destination_testing.go
```diff
@@ -194,52 +223,205 @@ func TestData() schema.CQTypes {
...
+		return fmt.Errorf("failed to write one second time: %w", err)
+	}
+
+	resourcesRead, err = p.readAll(ctx, tables[0], sourceName)
+	if err != nil {
+		return fmt.Errorf("failed to read all second time: %w", err)
+	}
+
+	if len(resourcesRead) != 2 {
+		return fmt.Errorf("expected 2 resource, got %d", len(resourcesRead))
```
hermanschaaf, line 306:
```suggestion
		return fmt.Errorf("expected 2 resources, got %d", len(resourcesRead))
```

plugins/destination_testing.go
```diff
@@ -194,52 +223,205 @@ func TestData() schema.CQTypes {
...
+	}
+
+	if resource.Data.Equal(resourcesRead[0]) {
+		return fmt.Errorf("expected data to be %v, got %v", resource.Data, resourcesRead[0])
+	}
+
+	if resource2.Data.Equal(resourcesRead[1]) {
+		return fmt.Errorf("expected data to be %v, got %v", resource.Data, resourcesRead[0])
```
hermanschaaf, line 291:
```suggestion
		return fmt.Errorf("expected data to be %v, got %v", resource.Data, resourcesRead[1])
```

erezrokah:
Looks like the newly added tests on Windows are failing https://github.com/cloudquery/plugin-sdk/actions/runs/3412572062/jobs/5678256683

yevgenypats:
> Looks like the newly added tests on Windows are failing https://github.com/cloudquery/plugin-sdk/actions/runs/3412572062/jobs/5678256683

wow so weird. maybe something related to `time`. seems like it cannot delete stale data on windows or fails to do so.

yevgenypats:
Ok I think found the bug on windows actions. seems like either it is "too fast" but more likely `time.Now` returns the same thing 🤷  per run.

hermanschaaf (approved):

plugins/destination_testing.go
```diff
@@ -3,243 +3,241 @@ package plugins
...
+func DestinationPluginTestSuiteRunner(t *testing.T, p *DestinationPlugin, spec specs.Destination, options ...DestinationPluginTestSuiteOption) {
+	t.Helper()
+	suite := &DestinationPluginTestSuite{}
+	for _, option := range options {
+		option(suite)
+	}
+	ctx := context.Background()
+	logger := getTestLogger(t)
+
+	t.Run("TestWriteOverwrite", func(t *testing.T) {
```
erezrokah, line 222: Maybe we should use a list of enums instead of multiple booleans? Then this should will be a for loop of the suites we want to run (instead of running, then skipping).

We can default to all if the list is `nil`

plugins/destination_testing.go
```diff
@@ -3,243 +3,241 @@ package plugins
...
+	t.Helper()
+	zerolog.TimeFieldFormat = zerolog.TimeFormatUnixMs
+	return zerolog.New(zerolog.NewTestWriter(t)).Output(
+		zerolog.ConsoleWriter{Out: os.Stderr, TimeFormat: time.StampMicro},
+	).Level(zerolog.DebugLevel).With().Timestamp().Logger()
 }
 
-func DestinationPluginTestHelper(ctx context.Context, p *DestinationPlugin, logger zerolog.Logger, spec specs.Destination) error {
+func (s *DestinationPluginTestSuite) destinationPluginTestWriteOverwrite(ctx context.Context, p *DestinationPlugin, logger zerolog.Logger, spec specs.Destination) error {
+	if s.skipTestOverwrite {
```
erezrokah, line 52: This seems redundant as it was checked by the caller

plugins/destination_testing.go
```diff
@@ -3,243 +3,241 @@ package plugins
...
+		resource2,
 	}
-	syncTime := time.Now()
-	sourceName := "test_helper_" + syncTime.String()
 
-	resources := make(chan *schema.DestinationResource, 1)
-	expectedResource := &schema.DestinationResource{
-		TableName: tables[0].Name,
-		Data:      TestData(),
+	if !s.skipTestOverWriteDeleteStale {
```
erezrokah, line 84: These lines don't impact the outcome of the test (you can remove them and `TestDestinationPlugin` passes

plugins/destination_testing.go
```diff
@@ -3,243 +3,241 @@ package plugins
...
+		table,
+	}
+	if err := p.Migrate(ctx, tables); err != nil {
+		return fmt.Errorf("failed to migrate tables: %w", err)
+	}
+
+	sourceName := uuid.NewString()
+	resource := schema.DestinationResource{
+		TableName: table.Name,
+		Data:      testdata.TestData(),
```
erezrokah, line 173: I'm not that happy about different tests sharing the same data. I think it's better to have it duplicated than creating a dependency between the different tests, WDYT?

plugins/destination_testing.go
```diff
@@ -3,243 +3,241 @@ package plugins
...
+	}
+	resource2 := schema.DestinationResource{
+		TableName: table.Name,
+		Data:      testdata.TestData(),
+	}
+	_ = resource2.Data[5].Set("00000000-0000-0000-0000-000000000007")
```
erezrokah, line 78: This might a bit hard to read. Maybe we can struct the test data in a way that it's clear what is its type?

plugins/destination_testing.go
```diff
@@ -3,243 +3,241 @@ package plugins
...
+func getTestLogger(t *testing.T) zerolog.Logger {
+	t.Helper()
+	zerolog.TimeFieldFormat = zerolog.TimeFormatUnixMs
+	return zerolog.New(zerolog.NewTestWriter(t)).Output(
+		zerolog.ConsoleWriter{Out: os.Stderr, TimeFormat: time.StampMicro},
+	).Level(zerolog.DebugLevel).With().Timestamp().Logger()
 }
 
-func DestinationPluginTestHelper(ctx context.Context, p *DestinationPlugin, logger zerolog.Logger, spec specs.Destination) error {
+func (s *DestinationPluginTestSuite) destinationPluginTestWriteOverwrite(ctx context.Context, p *DestinationPlugin, logger zerolog.Logger, spec specs.Destination) error {
```
erezrokah, line 51: Doesn't this test only `OverWriteDeleteStale`?

If I remove the code that handles `skipTestOverWriteDeleteStale` (both sections of it) the test fails.
I'm not sure who this tests only overwrite mode.

May I suggest splitting this suite to 2 distinct suites, 1 for overwrite mode and another for overwrite with delete stale?

plugins/destination_test.go
```diff
@@ -0,0 +1,14 @@
...
+
+import (
+	"testing"
+
+	"github.com/cloudquery/plugin-sdk/specs"
+)
+
+func TestDestinationPlugin(t *testing.T) {
+	p := NewDestinationPlugin("test", "development", NewTestDestinationMemDBClient)
+	DestinationPluginTestSuiteRunner(t, p, specs.Destination{
```
erezrokah, line 11: I'm a bit confused by this. `DestinationPluginTestSuiteRunner` runs `TestWriteOverwrite` and `TestWriteAppend`, but we set the plugin write mode to `WriteModeAppend`.

I'd expect it so each suite will run on a specific plugin write mode

erezrokah (changes requested):
I think the helper methods and memory plugin make sense.

I might be missing something with the test suites. Maybe we could make them more readable by having a clear structure of `setup, act, assert`, and splitting the suites (especially the overwrite one), so each one tests a single use case?

plugins/destination_testing.go
yevgenypats: Not sure I understand. some plugins dont support `DeleteStale`

plugins/destination_testing.go
yevgenypats: They dont share it. it creates new data every call to `TestData`

plugins/destination_test.go
yevgenypats: The testsuite will re-run `Init`. maybe I need to change it to get just the spec of the plugin instead of the destination.

plugins/destination_testing.go
erezrokah: By share I mean if someone changes `TestData` it impacts all tests (i.e. I can add data to test something and it will break other tests accidentally)

plugins/destination_testing.go
yevgenypats: Tried a different approach now.

yevgenypats:
@erezrokah Thanks! Can you please take another look?

plugins/destination_testing.go
erezrokah: I mean you can remove the entire block (including `DeleteStale`) and the test still passes

plugins/destination_testing.go
```diff
@@ -3,243 +3,225 @@ package plugins
...
+		resource,
+		resource2,
+	}
+
+	if s.tests.DeleteStale {
+		if err := p.DeleteStale(ctx, tables, sourceName, syncTime); err != nil {
+			return fmt.Errorf("failed to delete stale data: %w", err)
+		}
+	}
+
```
erezrokah, line 70:
```suggestion
```

I mean you can remove this block and the test still passes so it seems redundant

plugins/destination_testing.go
yevgenypats: But I want to run deletestale for plugins that support it so why remove it? or we are talking about a different block.

plugins/destination_testing.go
yevgenypats: Oh you mean the function itself. yes but no one else is using it so let's keep it as is now given we didn't have tests before.

plugins/destination_testing.go
yevgenypats: I added a struct with options instead of `with`

plugins/destination_testing.go
erezrokah: That is what's confusing for me, why do we need it if the test passes without it?

plugins/destination_testing.go
yevgenypats: Oh I see there were two delete stales. The first one is a redundant yes.

plugins/destination_testing.go
```diff
@@ -3,243 +3,225 @@ package plugins
...
+	}
+	suite := &destinationTestSuite{
+		tests: tests,
 	}
+	ctx := context.Background()
+	logger := getTestLogger(t)
+
+	t.Run("TestWriteOverwrite", func(t *testing.T) {
+		t.Helper()
+		if suite.tests.Overwrite {
```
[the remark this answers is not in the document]
yevgenypats: I think not because otherwise I cant call t.skip which will give indicator on what tests were skipped natively.

plugins/destination_testing.go
yevgenypats: I removed it I was referring to a deletestale lateron where I check if it was deleted after that via read

plugins/destination_testing.go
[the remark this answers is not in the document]
erezrokah: Ah, cool but then it should be `if !suite.tests.Overwrite` to skip when the condition is false?

PR #373, opened 2022-11-08
fix(faker): Use `MarshalText` for faker timestamps

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/3627.

**The TLDR** is that we should `MarshalText` when faking time data, as that follows an RFC.

**The DON'T READ part is**:
If we don't do it and default to `String()` that sometimes creates a string that cannot be parsed by `const defaultStringFormat = "2006-01-02 15:04:05.999999999 -0700 MST"`.

You can see an example in https://github.com/cloudquery/cloudquery/issues/3627 that the `2110-10-19 08:53:50.21670207 +0000 UTC m=+2775422040.494306874` string was created.
The reason that the string can't be parsed is because the nano seconds part has 8 digits and not 9, that is `21670207` instead of `216702070`.
The nano seconds string needs to be padded only when adding the `m=+2775422040.494306874` part. This was added in go 1.19 (see https://pkg.go.dev/time#Time.String).
To summarize:
1. `time.Parse("2006-01-02 15:04:05.999999999 -0700 MST", "2110-10-19 08:53:50.21670207 +0000 UTC")` works
2. `time.Parse("2006-01-02 15:04:05.999999999 -0700 MST", "2110-10-19 08:53:50.216702070 +0000 UTC  m=+2775422040.494306874")` works
3. `time.Parse("2006-01-02 15:04:05.999999999 -0700 MST", "2110-10-19 08:53:50.21670207 +0000 UTC  m=+2775422040.494306874")` fails

So technically this could be a bug in go, that needs to pad the nano seconds when adding the `m=+` part, though there is no guarantee on the format of `time.String()`.

I added another test `TestTimestamptzSetString` that shows this behavior. The only way I could find to create a string that has the `m=+` part is using `time.Now()`. It's in the PR to demonstrate the issue and I'll remove it before merging

---

- [ ] Update or add tests 🧪

### DIFF
--- a/schema/timestamptz_test.go
+++ b/schema/timestamptz_test.go
@@ -1,7 +1,6 @@
 package schema
 
 import (
-	"math/rand"
 	"testing"
 	"time"
 
@@ -37,17 +36,5 @@ func TestTimestamptzSet(t *testing.T) {
 		err := r.Set(tt.source)
 		require.NoError(t, err)
 		require.Equal(t, tt.result, r)
-	}
-}
-
-func TestTimestamptzSetString(t *testing.T) {
-	for i := 0; i < 100000; i++ {
-		timeString := time.Now().Add(time.Duration(rand.Int63())).String()
-		_, err := time.Parse(defaultStringFormat, timeString)
-		if err == nil {
-			t.Errorf("time.Parse(%q) should return error", timeString)
-		} else {
-			t.Errorf("time.Parse(%q) should not return error", timeString)
-		}
 	}
 }
--- a/schema/timestamptz_test.go
+++ b/schema/timestamptz_test.go
@@ -1,12 +1,18 @@
 package schema
 
 import (
+	"math/rand"
 	"testing"
 	"time"
+
+	"github.com/stretchr/testify/assert"
 )
 
 func TestTimestamptzSet(t *testing.T) {
 	type _time time.Time
+
+	timeInstance := time.Date(2105, 7, 23, 22, 23, 37, 750076110, time.UTC)
+	timeRFC3339NanoBytes, _ := timeInstance.MarshalText()
 
 	successfulTests := []struct {
 		source interface{}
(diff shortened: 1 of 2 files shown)


schema/timestamptz_test.go
```diff
@@ -21,19 +27,23 @@ func TestTimestamptzSet(t *testing.T) {
...
-		}
+		require.NoError(t, err)
+		require.Equal(t, tt.result, r)
+	}
+}
 
-		if !r.Equal(&tt.result) {
-			t.Errorf("%d: %v != %v", i, r, tt.result)
-		}
+func TestTimestamptzSetString(t *testing.T) {
```
erezrokah, line 43: I'll remove this before merging the PR. This is here just to show the bug

erezrokah:
>The reason that the string can't be parsed is because the nano seconds part has 8 digits and not 9, that is 21670207 instead of 216702070.

This is only partially correct. I found other use cases the parsing fails (even with 9 digits).

schema/timestamptz_test.go
```diff
@@ -21,19 +27,27 @@ func TestTimestamptzSet(t *testing.T) {
...
+	}
+}
 
-		if !r.Equal(&tt.result) {
-			t.Errorf("%d: %v != %v", i, r, tt.result)
+func TestTimestamptzSetString(t *testing.T) {
+	for i := 0; i < 100000; i++ {
+		timeString := time.Now().Add(time.Duration(rand.Int63())).String()
+		_, err := time.Parse(defaultStringFormat, timeString)
+		if err == nil {
```
erezrokah, line 47: Both errors are printed

hermanschaaf (approved):
Amazing work! 👏

erezrokah:
> So technically this could be a bug in go, that needs to pad the nano seconds when adding the `m=+` part, though there is no guarantee on the format of `time.String()`.

I posted a message about this in https://groups.google.com/g/golang-nuts (I think it needs to be approved first)

yevgenypats (approved):

PR #375, opened 2022-11-08
fix: Deadlock on Context Timeout

bbernays:
#### Summary

---

- [ ] Update or add tests 🧪

### DIFF
--- a/plugins/source_scheduler_dfs.go
+++ b/plugins/source_scheduler_dfs.go
@@ -122,10 +122,9 @@ func (p *SourcePlugin) resolveResourcesDfs(ctx context.Context, table *schema.Ta
 				defer wg.Done()
 				//nolint:all
 				resolvedResource := p.resolveResource(ctx, table, client, parent, resourcesSlice[i])
-				if resolvedResource == nil {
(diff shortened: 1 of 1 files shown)


hermanschaaf (commented):
@bbernays Could you explain the fix here? I might need more coffee, but I've been staring at it a while and I can't tell what the difference is from before that would fix a deadlock 👀

plugins/source_scheduler_dfs.go
```diff
@@ -122,10 +122,9 @@ func (p *SourcePlugin) resolveResourcesDfs(ctx context.Context, table *schema.Ta
 				defer wg.Done()
 				//nolint:all
 				resolvedResource := p.resolveResource(ctx, table, client, parent, resourcesSlice[i])
-				if resolvedResource == nil {
-					return
+				if resolvedResource != nil {
+					resourcesChan <- resolvedResource
```
candiduslynx, line 126: is this just a style change?

candiduslynx (changes requested):
only a style change, drop it or make a chore?

Issue #299, opened 2022-10-17
feat(codegen): `codegen` should fail if column names clash (after adding extra columns, naturally)


candiduslynx:
```go
type A struct {
    Abc string
}
```

If we feed `struct A` to `codegen` it will naturally produce column named `abc`.
If we add an extra column with the same name and forget to add filed `Abc` to be skipped, we end up with duplicate names in the resulting resource table.

Now, this will result in `cloudquery` CLI failing on the sync/init part as it checks for the data integrity then. There will be something like `panic: found duplicate columns in source plugin: test: duplicate column abc in table test_resource`.

The proposal is to add this check to `codegen`, too, so that such issues are highlighted during the development.

disq:
While this is a quality-of-life feature, it's easy to miss an unintended column overwrite if this feature is enabled implicitly.

candiduslynx:
Is there an example where we want this? Maybe, the data has column N that we overwrite as some meta-info, and we want to preserve the column info, too?
In this case, at least let's fail the generation if there are clashing column names, so that the developer will have to address this issue right away?

PR #376, opened 2022-11-09
feat(codegen): Add sanity check to `TableDefinition`

candiduslynx:
Add sanity check to `TableDefinition` (#299)

### DIFF
--- a/codegen/table.go
+++ b/codegen/table.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"reflect"
 
+	"github.com/cloudquery/plugin-sdk/schema"
 	"github.com/rs/zerolog"
 	"golang.org/x/exp/slices"
 )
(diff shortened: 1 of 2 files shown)


codegen/table.go
```diff
@@ -90,5 +92,50 @@ func NewTableFromStruct(name string, obj interface{}, opts ...TableOption) (*Tab
...
-	return t, nil
+	// add PK options
+	columns := make(ColumnDefinitions, 0, len(t.Columns))
+	for _, column := range t.Columns {
+		if _, ok := t.extraPKColumns[column.Name]; ok {
+			column.Options.PrimaryKey = true
+		}
+		columns = append(columns, column)
+	}
+	t.Columns = columns
```
hermanschaaf, line 103: I think this should raise an error if a specified PK column was not found, since it would indicate a configuration error

codegen/table.go
```diff
@@ -90,5 +92,50 @@ func NewTableFromStruct(name string, obj interface{}, opts ...TableOption) (*Tab
...
+			return fmt.Errorf("%s->%s: invalid column type", t.Name, column.Name)
+		case columns[column.Name]:
+			return fmt.Errorf("%s->%s: duplicate column name", t.Name, column.Name)
+		default:
+			columns[column.Name] = true
+			hasPK = hasPK || column.Options.PrimaryKey
+		}
+	}
+
+	if !hasPK {
```
erezrokah, line 136: Is this always an error? I think we can have tables without PKs and then we use a random ID via `_cq_id`.

Maybe we can return different errors so the caller can check and allow for some (for example X table can not have a primary key)

erezrokah (approved):
Nice, one comment on erroring on missing primary keys (not sure it's blocking)

codegen/table.go
```diff
@@ -90,5 +92,50 @@ func NewTableFromStruct(name string, obj interface{}, opts ...TableOption) (*Tab
...
+		return fmt.Errorf("no PK for table %s", t.Name)
+	}
```
hermanschaaf, line 138: Does this get called before or after `_cq_id` is added? Because we have a few tables where there is no PK specified and it then defaults to `_cq_id`, and this is currently considered valid.

hermanschaaf (changes requested):

codegen/table.go
candiduslynx: Nice point

codegen/table.go
candiduslynx: I'll drop the PK check for now, then

codegen/table.go
candiduslynx: Dropping this check for now

erezrokah (approved):

yevgenypats (approved):

PR #377, opened 2022-11-09
fix: Context cancelled too early for delete stale mode

hermanschaaf:
This is a fix for this error, encountered when postgresql plugin is run with overwrite-delete-stale mode:

```
Error: failed to write for gcp->postgresql: failed to CloseAndRecv client: rpc error: code = Canceled desc = got EOF. failed to wait for plugin: failed to execute batch: context canceled
2022-11-09T10:11:35Z ERR exiting with error error="failed to write for gcp->postgresql: failed to CloseAndRecv client: rpc error: code = Canceled desc = got EOF. failed to wait for plugin: failed to execute batch: context canceled" module=cli
```

The reason it didn't work before is that the grouperr context gets cancelled as soon as Wait() returns. We should be using the original `ctx` for the subsequent delete stale call.

This also fixes this error:

```
2022-11-09T10:02:22Z ERR failed to remove source socket file error="remove /var/folders/_c/hm9lkvfj7fz_rykn71rmr8gh0000gn/T/cq-WDwWDgDPiakfRkxi.sock: no such file or directory" module=cli
```

by using `os.RemoveAll` so that the file is removed if it is there, or ignored otherwise.

### DIFF
--- a/clients/destination.go
+++ b/clients/destination.go
@@ -289,7 +289,7 @@ func (c *DestinationClient) Write2(ctx context.Context, tables schema.Tables, so
 		if err := saveClient.Send(&pb.Write2_Request{
 			Resource: resource,
 		}); err != nil {
-			return fmt.Errorf("failed to call Write.Send: %w", err)
+			return fmt.Errorf("failed to call Write2.Send: %w", err)
 		}
 	}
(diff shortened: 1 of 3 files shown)


erezrokah (approved):
🚀

disq (approved):

PR #381, opened 2022-11-09
fix: Revert "fix(faker): Use `MarshalText` for faker timestamps (#373)"

erezrokah:
This reverts commit a29143861b22432c81cdc8b04650d9d8d0ac9671.

#### Summary

Reverts https://github.com/cloudquery/plugin-sdk/issues/373 as I applied that fix in the wrong place. I think the fix should be where we serialize/deserialize `time.time`

---

- [ ] Update or add tests 🧪

### DIFF
--- a/faker/faker_test.go
+++ b/faker/faker_test.go
@@ -2,6 +2,7 @@ package faker
 
 import (
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 )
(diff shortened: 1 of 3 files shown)


faker/faker_test.go
```diff
@@ -21,6 +23,7 @@ func TestFaker(t *testing.T) {
 	assert.NotEmpty(t, a.A)
 	assert.NotEmpty(t, a.B)
 	assert.NotEmpty(t, a.C)
+	assert.NotEmpty(t, a.D)
```
erezrokah, line 26: Added a test case to catch this

erezrokah:
Going to merge this to un-break the SDK

PR #382, opened 2022-11-09
fix: Use MarshalText when serializing timestamps when applicable

erezrokah:
#### Summary

Proper fix for https://github.com/cloudquery/plugin-sdk/pull/373

---

- [ ] Update or add tests 🧪

### DIFF
--- a/schema/timestamptz.go
+++ b/schema/timestamptz.go
@@ -2,6 +2,7 @@
 package schema
 
 import (
+	"encoding"
 	"fmt"
 	"time"
 )
@@ -22,10 +23,6 @@ const (
 
 type TimestamptzTransformer interface {
 	TransformTimestamptz(*Timestamptz) interface{}
-}
-
-type TextMarshaler interface {
-	MarshalText() ([]byte, error)
 }
 
 type Timestamptz struct {
@@ -94,7 +91,7 @@ func (dst *Timestamptz) Set(src interface{}) error {
 		if originalSrc, ok := underlyingTimeType(src); ok {
 			return dst.Set(originalSrc)
 		}
-		if value, ok := value.(TextMarshaler); ok {
+		if value, ok := value.(encoding.TextMarshaler); ok {
 			s, err := value.MarshalText()
 			if err == nil {
 				return dst.Set(string(s))
--- a/schema/timestamptz.go
+++ b/schema/timestamptz.go
@@ -96,6 +96,7 @@ func (dst *Timestamptz) Set(src interface{}) error {
 			if err == nil {
 				return dst.Set(string(s))
 			}
+			// fall through to String() method
 		}
 		if value, ok := value.(fmt.Stringer); ok {
 			s := value.String()
--- a/schema/timestamptz.go
+++ b/schema/timestamptz.go
@@ -24,6 +24,10 @@ type TimestamptzTransformer interface {
 	TransformTimestamptz(*Timestamptz) interface{}
 }
 
+type TextMarshaler interface {
+	MarshalText() ([]byte, error)
+}
+
 type Timestamptz struct {
 	Time             time.Time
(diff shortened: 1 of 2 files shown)


schema/timestamptz.go
```diff
@@ -90,6 +94,12 @@ func (dst *Timestamptz) Set(src interface{}) error {
 		if originalSrc, ok := underlyingTimeType(src); ok {
 			return dst.Set(originalSrc)
 		}
+		if value, ok := value.(TextMarshaler); ok {
+			s, err := value.MarshalText()
+			if err == nil {
```
candiduslynx, line 99: What happens if `err` is not nil? Do we lose this info?

candiduslynx (changes requested):

schema/timestamptz.go
```diff
@@ -24,6 +24,10 @@ type TimestamptzTransformer interface {
 	TransformTimestamptz(*Timestamptz) interface{}
 }
 
+type TextMarshaler interface {
```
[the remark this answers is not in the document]
erezrokah: Ah nice I'll use that

schema/timestamptz.go
erezrokah: Yeah, we can't really print a warning and using `String()` might just work

schema/timestamptz_test.go
```diff
@@ -21,19 +30,16 @@ func TestTimestamptzSet(t *testing.T) {
 		{source: _time(time.Date(1970, 1, 1, 0, 0, 0, 0, time.Local)), result: Timestamptz{Time: time.Date(1970, 1, 1, 0, 0, 0, 0, time.Local), Status: Present}},
 		{source: Infinity, result: Timestamptz{InfinityModifier: Infinity, Status: Present}},
 		{source: NegativeInfinity, result: Timestamptz{InfinityModifier: NegativeInfinity, Status: Present}},
-		{source: "2105-07-23 22:23:37.133334968 +0300 UTC m=+2610529275.484148261", result: Timestamptz{Time: time.Date(2105, 7, 23, 22, 23, 37, 133334968, time.UTC), Status: Present}},
-		// {source: "2020-04-05 06:07:08Z", result: Timestamptz{Time: time.Date(2020, 4, 5, 6, 7, 8, 0, time.UTC), Status: Present}},
+		{source: string(timeRFC3339NanoBytes), result: Timestamptz{Time: time.Date(2105, 7, 23, 22, 23, 37, 750076110, time.UTC), Status: Present}},
```
erezrokah, line 33: Re-added the reverted tests and added a new one for a wrapped `time.Time`

candiduslynx (approved):

schema/timestamptz_test.go
```diff
@@ -21,19 +30,16 @@ func TestTimestamptzSet(t *testing.T) {
...
 		var r Timestamptz
 		err := r.Set(tt.source)
-		if err != nil {
-			t.Errorf("%d: %v", i, err)
-		}
-
-		if !r.Equal(&tt.result) {
-			t.Errorf("%d: %v != %v", i, r, tt.result)
-		}
+		require.NoError(t, err)
```
yevgenypats, line 42: Let's not use testify.

yevgenypats (changes requested):
Let's avoid using testify it doesn't bring a lot of value and can use regular `if err` where we can also use better error messages.

schema/timestamptz_test.go
erezrokah: Any reason not to use it? It prints better error messages (a better diff between the expected and actual values)

erezrokah: I reverted that change as it's unrelated to this PR. However we do use `testify` in many other places

yevgenypats (approved):

PR #388, opened 2022-11-10
fix: Add -race when running tests

hermanschaaf:
This adds the `-race` flag to `go test`, so we can detect data races in the tests. There were two instances that I had to fix for this to succeed (not really a bug: just the test accessing a variable that another goroutine is writing to) - for now fixed using a lock around the variable.

Some other test were failing when I ran `go test -race -count 3 ./...` because of global variables being modified in the test. This also fixes this issue.  

The only real functional change is that we now only add a `_cq_id` column if it is not already present.


plugins/source.go
```diff
@@ -46,6 +46,10 @@ const (
 // Add internal columns
 func addInternalColumns(tables []*schema.Table) {
 	for _, table := range tables {
+		c := table.Column("_cq_id")
+		if c != nil {
```
disq, line 50: nit: since `c` is only used in this block, I would use the `if c := ...; c != nil {` syntax.

yevgenypats (approved):

Issue #354, opened 2022-11-06
Investigate process termination on Windows


erezrokah:
Related to https://github.com/cloudquery/plugin-sdk/pull/352.

Looks like we fail to kill the plugin process on Windows:
https://github.com/cloudquery/plugin-sdk/actions/runs/3404741718/jobs/5662201307#step:6:49
![image](https://user-images.githubusercontent.com/26760571/200178266-2cc0eb48-bb80-4c7f-92b6-01ed157710b4.png)

We should investigate, fix, then remove `//go:build !windows` from `clients/destination_test.go` and `clients/source_test.go`

PR #391, opened 2022-11-10
fix: Close zip archive when we're done with it

hermanschaaf:
I suspect this will help fix https://github.com/cloudquery/plugin-sdk/issues/354


erezrokah (approved):

erezrokah:
Maybe we can remove https://github.com/cloudquery/plugin-sdk/blob/b76f2cc21d8d78af400441a71d264bb717626287/clients/destination_test.go#L1 and https://github.com/cloudquery/plugin-sdk/blob/b76f2cc21d8d78af400441a71d264bb717626287/clients/source_test.go#L1 to verify

hermanschaaf:
@erezrokah Thanks, looks like it works!

erezrokah:
> @erezrokah Thanks, looks like it works!

Great catch, nice work 🪟

yevgenypats (approved):

PR #395, opened 2022-11-10
fix: Exit early if all Write workers have stopped

hermanschaaf:
This is a stop-gap solution to ensure that the `sync` operation doesn't hang, and instead returns the underlying error that was returned by the worker(s).

I think this is probably not the final solution and this part of the code will need some further thought around how to handle Write worker failures.

Related to https://github.com/cloudquery/cloudquery/issues/4070


candiduslynx (approved):

erezrokah (approved):

yevgenypats (changes requested):
I think this wont solve the potential deadlock. `writeWorkers` is also set statically to 1 so Im not sure there is an issue right here. But in any case I think the right way is to just select on `gctx.Done` in addition to the channel

hermanschaaf:
> I think this wont solve the potential deadlock. `writeWorkers` is also set statically to 1 so Im not sure there is an issue right here. But in any case I think the right way is to just select on `gctx.Done` in addition to the channel

@yevgenypats Thanks! The `gctx.Done` fix is (almost) equivalent but simpler, thanks for the suggestion (updated). But I just want to say that there definitely is a deadlock here, and this does solve it. I have made a reproduction of the issue. To do so, I added a flag to the destination test plugin that makes it return an error on write ([PR here](https://github.com/cloudquery/cloudquery/pull/4378)).

Config file:

```
kind: source
spec:
  name: "test"
  version: "v1.3.14"
  registry: "github"
  path: "cloudquery/test"
  tables: 
    - "*"
  destinations: 
    - "test"
---
kind: destination
spec:
  name: "test"
  version: "v1.7.1"
  registry: grpc
  path: "localhost:7777"
  write_mode: "overwrite-delete-stale"
  spec:
    error_on_write: true
```

Including only the `error_on_write` change in the destination plugin, but not the fix in this PR (hangs forever):

```
cloudquery sync test.yml
Loading spec(s) from test.yml
Starting migration for: test -> [test]
Migration completed successfully.
Starting sync for:  test -> [test]
| Syncing resources... (1/-, 294 resources/s) [0s]
```

Including both the `error_on_write` change and the fix in this PR in the destination plugin:

```
cloudquery sync test.yml
Loading spec(s) from test.yml
Starting migration for: test -> [test]
Migration completed successfully.
Starting sync for:  test -> [test]
Error: failed to write for test->test: failed to CloseAndRecv client: rpc error: code = Unknown desc = got EOF. failed to wait for plugin: error_on_write is true
```

plugins/destination.go
```diff
@@ -156,6 +156,9 @@ func (p *DestinationPlugin) Write(ctx context.Context, tables schema.Tables, sou
 			Data:      schema.TransformWithTransformer(p.client, r.Data),
 		}
 		select {
+		case <-gctx.Done():
```
yevgenypats, line 159: I think that was the culprit. the line below is not needed as I used `ctx` instead of `gctx`. If the main context will be cancelled then `gctx` also will be cancelled.

yevgenypats (changes requested):
Looks good. One more nit.

yevgenypats:
> > I think this wont solve the potential deadlock. `writeWorkers` is also set statically to 1 so Im not sure there is an issue right here. But in any case I think the right way is to just select on `gctx.Done` in addition to the channel
> 
> @yevgenypats Thanks! The `gctx.Done` fix is (almost) equivalent but simpler, thanks for the suggestion (updated). But I just want to say that there definitely is a deadlock here, and this does solve it. I have made a reproduction of the issue. To do so, I added a flag to the destination test plugin that makes it return an error on write ([PR here](https://github.com/cloudquery/cloudquery/pull/4378)).
> 
> Config file:
> 
> ```
> kind: source
> spec:
>   name: "test"
>   version: "v1.3.14"
>   registry: "github"
>   path: "cloudquery/test"
>   tables: 
>     - "*"
>   destinations: 
>     - "test"
> ---
> kind: destination
> spec:
>   name: "test"
>   version: "v1.7.1"
>   registry: grpc
>   path: "localhost:7777"
>   write_mode: "overwrite-delete-stale"
>   spec:
>     error_on_write: true
> ```
> 
> Including only the `error_on_write` change in the destination plugin, but not the fix in this PR (hangs forever):
> 
> ```
> cloudquery sync test.yml
> Loading spec(s) from test.yml
> Starting migration for: test -> [test]
> Migration completed successfully.
> Starting sync for:  test -> [test]
> | Syncing resources... (1/-, 294 resources/s) [0s]
> ```
> 
> Including both the `error_on_write` change and the fix in this PR in the destination plugin:
> 
> ```
> cloudquery sync test.yml
> Loading spec(s) from test.yml
> Starting migration for: test -> [test]
> Migration completed successfully.
> Starting sync for:  test -> [test]
> Error: failed to write for test->test: failed to CloseAndRecv client: rpc error: code = Unknown desc = got EOF. failed to wait for plugin: error_on_write is true
> ```

Going to look into this deadlock

plugins/destination.go
hermanschaaf: Ah yeah, good point

yevgenypats (approved):

PR #398, opened 2022-11-10
feat: Add support for glob matching in config

hermanschaaf:
This adds wildcard matching to the tables / skip_tables config entries. It also adds the ability to skip child tables, which may sometimes be useful if the child tables are expensive to fetch.

This will be useful to reduce the number of entries needed in config. For example, to fetch all ec2-related resources, but skip vpn gateways:

```
  tables: 
    - "aws_ec2_*"
  skip_tables:
    - "aws_ec2_vpn_gateways"
```

Instead of  20+ entries that it would have required before.

# Behavior
This is how table configs will work after this change:
 - Glob matches will match against both top-level and descendent tables
 - If you want to sync a table, it needs to be explicitly mentioned. This includes descendent (child) tables. This is the one and only breaking change, because previously child tables needed to be excluded from config.
 - Similarly, if you want to skip a table or a descendent table, it needs to be explicitly skipped in the `skip_tables` section, either through a glob match or by using its exact name.

This will require some config updates, but I believe this is a more consistent approach: if you want a table to by synced, include it. If you want it to be skipped, skip it. There are no implicit inclusions anymore.

# Breaking Change

This will not be considered a breaking plugin-sdk change (no change to the protocol), but it *will* be considered a breaking change in all source plugins, where we should bump the major version. Old configs will continue to work, but the behavior will change slightly for some of them, since top-level tables that were explicitly included will now stop fetching their descendent tables by default. 

Relates to https://github.com/cloudquery/cloudquery/issues/3963


yevgenypats:
Looks great! Seems like it's one of the most requested features :)

Suggestion: Let's import this library https://github.com/ryanuber/go-glob/blob/master/glob.go directly and keep the license/copyright so we can fix if any issues and avoid another dependency especially because the code there is quite simple and the library is not seems like it is maintained (it's not really a full glob but rather a `*` support)

plugins/source_validate.go
```diff
@@ -27,80 +27,73 @@ func (p *SourcePlugin) validate() error {
...
-	// return an error if child table is included
-	for _, t := range tables {
-		tt := p.tables.Get(t)
-		if tt.Parent != nil {
-			pt := tt.Parent
-			for pt.Parent != nil {
-				pt = pt.Parent
+	// return an error if child table is included without its parent
+	for _, t := range remainingTables {
+		missingParents := make([]string, 0)
```
disq, line 85:
```suggestion
		var missingParents []string
```

disq (changes requested):
One nit, and a question: Would something like `aws_ec2_instances*` work? include aws_ec2_instances and all child tables. The point is, the table name is fully mentioned before the ending `*` char. If so, config specifying only parent tables can be updated to add "*" in the end and it would work without needing to either "delete one char and add *" or "add every relation".

hermanschaaf:
> Would something like aws_ec2_instances* work? 

@disq I think I'd rather keep things simple / unsurprising. Users are going to need to make an update to their configs either way (if they want the same behavior as before), and I don't think it really matters whether they have to type one character or two 🤷

hermanschaaf:
We should also follow this up with some CLI UX improvements to show the number and names of matched tables. ⚡

disq:
> - "aws_ec2_instance*"

Should work out fine since the tables are supposed to be plurally named.

disq (approved):

PR #400, opened 2022-11-11
feat: Allow setting threads value

candiduslynx:


yevgenypats (changes requested):
I don't think this is the right fix.

Seems we have a problem here. a thread can have as many goroutines as 10,000 or even more especially for network I/O. if we do `100,000` max threads then instead of having a normal error that we can debug the operating system will kill CloudQuery due to out of memory.

Thread size is 1MB (depending on configuration) and 100,000 is now 100GB of RAM which I dont think we want.

disq:
My hot take: If it's in the `debug` package it shouldn't be in production... Would be in `runtime` if it was something that needed adjusting on a daily basis.

hermanschaaf:
@candiduslynx Are you okay for us to close this?

candiduslynx:
> @candiduslynx Are you okay for us to close this?

Yep, sure

PR #404, opened 2022-11-13
fix: Allow searching relations by name

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/4489, and we'll need to update all dest plugins to use `GetAll`.
We could also pass an option to `Get`.

Maybe leave this open so others can review too

---

- [ ] Update or add tests 🧪

### DIFF
--- a/schema/table.go
+++ b/schema/table.go
@@ -83,6 +83,20 @@ func (tt Tables) Get(name string) *Table {
 	return nil
 }
 
+// GetAll return table by name. Returns top-level table and relations.
+func (tt Tables) GetAll(name string) *Table {
+	for _, t := range tt {
+		if t.Name == name {
+			return t
+		}
+		table := t.Relations.GetAll(name)
+		if table != nil {
+			return table
+		}
+	}
+	return nil
+}
+
 // GlobMatch returns a list of tables that match a given glob. If a parent table is matched,
 // it will also return all its descendants as part of the list.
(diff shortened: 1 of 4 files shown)


schema/table.go
```diff
@@ -83,6 +83,20 @@ func (tt Tables) Get(name string) *Table {
 	return nil
 }
 
+// GetAll returns a table by name. Returns top-level tables and relations.
+func (tt Tables) GetAll(name string) *Table {
```
disq, line 87: I would make this the default `Get` and add a `GetTopLevel` etc. that looks only in the top level list.

erezrokah: Maybe both `GetTopLevel` and `GetAll`?

disq: Keeping the `Get` name would keep compatibility across plugins, even though there's only a handful dest plugins (if not just PG) that's using it.

hermanschaaf: `Get` (same as before) + `GetTopLevel` (new function) would get my vote

hermanschaaf (approved):

schema/table.go
erezrokah: Cool, I'm good with either. Let me update the PR

erezrokah:
I'll go ahead and force merge @yevgenypats so we can un-break the SDK

PR #406, opened 2022-11-14
fix: Update libs

candiduslynx:


erezrokah (approved):
This looks good to me, but you'll need to change it to `fix:` or `feat:` to trigger a release.
It is a user facing change so a `chore:` doesn't fit here I think

serve/source.go
```diff
@@ -108,10 +107,10 @@ func newCmdSourceServe(source *sourceServe) *cobra.Command {
 			limitListener := netutil.LimitListener(listener, 1)
 			// See logging pattern https://github.com/grpc-ecosystem/go-grpc-middleware/blob/v2/providers/zerolog/examples_test.go
 			s := grpc.NewServer(
-				middleware.WithUnaryServerChain(
+				grpc.ChainUnaryInterceptor(
```
yevgenypats, line 110: Was it moved from `middleware` ?

candiduslynx: yep

yevgenypats (approved):

PR #408, opened 2022-11-14
test: Add more non-happy flow tests

yevgenypats:
More tests for source and destination plugins with non-happy flow case, context cancellation and other fun stuff.

### DIFF
--- a/plugins/destination_memdb_plugin.go
+++ b/plugins/destination_memdb_plugin.go
@@ -2,6 +2,7 @@ package plugins
 
 import (
 	"context"
+	"fmt"
 	"time"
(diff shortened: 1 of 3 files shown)


erezrokah (approved):
Thanks for adding these tests @yevgenypats 🚀 

There's quite a lot boilerplate and duplication here we can improve (waiting for the gRPC server for example).

Also, this test mimics quite quite a lot of the CLI logic (e.g. calling terminate), but I guess that makes sense, and we should also have CLI tests that ensure the logic there works.

hermanschaaf (approved):

PR #409, opened 2022-11-15
fix: Inet marshal/unmarshal didnt work

yevgenypats:
Looks like we hit a nasty Go bug - https://github.com/golang/go/issues/35727.

@shimonp21 can you please confirm it solves the error that was triggered by the k8s plugin ?


yevgenypats:
Closing as branched from incorrect place - https://github.com/cloudquery/plugin-sdk/pull/410

PR #410, opened 2022-11-15
fix: Workaround Go Inet marshal bug

yevgenypats:
Fix nasty bug triggered by Go bug

https://github.com/golang/go/issues/35727

@shimonp21  can you please confirm this solves the bug triggered by the k8s cidr plugin.

### DIFF
--- a/schema/inet.go
+++ b/schema/inet.go
@@ -160,7 +160,6 @@ func (dst *Inet) UnmarshalJSON(b []byte) error {
 		return err
 	}
 	*dst = Inet{Status: tmp.Status}
-	dst.Status = tmp.Status
 	if dst.Status == Present {
 		if err := dst.Set(tmp.IPNet.String()); err != nil {
 			return err
--- a/schema/inet.go
+++ b/schema/inet.go
@@ -14,7 +14,7 @@ type InetTransformer interface {
 }
 
 // workaround this Golang bug: https://github.com/golang/go/issues/35727
-type inetIntermidate struct {
+type inetWrapper struct {
 	IPNet  *net.IPNet
 	Status Status
 }
@@ -155,10 +155,11 @@ func maybeGetIPv4(input string, ip net.IP) net.IP {
 
 // workaround this Golang bug: https://github.com/golang/go/issues/35727
 func (dst *Inet) UnmarshalJSON(b []byte) error {
-	tmp := inetIntermidate{}
+	tmp := inetWrapper{}
 	if err := json.Unmarshal(b, &tmp); err != nil {
 		return err
 	}
+	*dst = Inet{Status: tmp.Status}
 	dst.Status = tmp.Status
 	if dst.Status == Present {
 		if err := dst.Set(tmp.IPNet.String()); err != nil {
--- a/schema/inet.go
+++ b/schema/inet.go
@@ -3,6 +3,7 @@ package schema
 
 import (
 	"encoding"
+	"encoding/json"
 	"fmt"
 	"net"
 	"strings"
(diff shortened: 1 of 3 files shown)


schema/inet.go
```diff
@@ -12,6 +13,12 @@ type InetTransformer interface {
 	TransformInet(*Inet) interface{}
 }
 
+// workaround this Golang bug: https://github.com/golang/go/issues/35727
+type inetIntermidate struct {
```
hermanschaaf, line 17:
```suggestion
type inetIntermediate struct {
```

typo fix :)

schema/inet.go
```diff
@@ -146,6 +153,22 @@ func maybeGetIPv4(input string, ip net.IP) net.IP {
 	return ip.To4()
 }
 
+// workaround this Golang bug: https://github.com/golang/go/issues/35727
+func (dst *Inet) UnmarshalJSON(b []byte) error {
+	tmp := inetIntermidate{}
```
hermanschaaf, line 158:
```suggestion
	tmp := inetIntermediate{}
```

erezrokah (approved):
I'm assuming this works, but I'll let @shimonp21 verify

hermanschaaf (approved):
Looks good 👍 

Just thinking we should maybe add a test that covers the marshal/unmarshal behavior for all the types to make sure this fix works as expected, doesn't regress and that there aren't other similar bugs lurking in other types.

shimonp21 (changes requested):
This doesn't work. Let's sync on this online?

yevgenypats:
@shimonp21 this now works. @hermanschaaf added tests.

schema/inet_test.go
```diff
@@ -59,3 +60,28 @@ func TestInetSet(t *testing.T) {
...
+	err = json.Unmarshal(b, &r2)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if !r.Equal(&r2) {
+		t.Errorf("%v != %v", r, r2)
+	}
+
+	// workaround this Golang bug: https://github.com/golang/go/issues/35727
+	if len(r.IPNet.Mask) != len(r2.IPNet.Mask) {
```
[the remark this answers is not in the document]
yevgenypats: because we compare the values in the line before that. the bug was that the bytes are in different length. I can also add a byte comparision.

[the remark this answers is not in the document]
hermanschaaf: Ah I see. No all good, just curious

hermanschaaf (approved):

schema/inet_test.go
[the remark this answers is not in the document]
yevgenypats: but good call. Added more thorugh tests and also checked that it failed before the fix and after

schema/inet.go
```diff
@@ -146,6 +153,23 @@ func maybeGetIPv4(input string, ip net.IP) net.IP {
...
 }
 
+// workaround this Golang bug: https://github.com/golang/go/issues/35727
+func (dst *Inet) UnmarshalJSON(b []byte) error {
+	tmp := inetWrapper{}
+	if err := json.Unmarshal(b, &tmp); err != nil {
+		return err
+	}
+	*dst = Inet{Status: tmp.Status}
+	dst.Status = tmp.Status
```
disq, line 163: Do we need this line?

schema/inet.go
```diff
@@ -12,6 +13,12 @@ type InetTransformer interface {
 	TransformInet(*Inet) interface{}
 }
 
+// workaround this Golang bug: https://github.com/golang/go/issues/35727
+type inetWrapper struct {
```
disq, line 17: Maybe add a comment that this is only used in Unmarshal.
Maybe order structs so that exported struct `Inet` is at the top.

shimonp21 (approved):
Approving at @yevgenypats request.
We may still have a bug in handling ipv6 though - testing that right now

PR #415, opened 2022-11-18
chore: Add source plugin benchmark

hermanschaaf:
This adds a source plugin benchmark that runs in-memory. It allows for different scenarios to be set up, but currently comes with only two: one scenario without child tables, and one with child tables. 

Here are the results on my machine:

```
BenchmarkDefaultConcurrency
BenchmarkDefaultConcurrency-8                     	       1	     11637 resources/s	     12626 targetResources/s	73510696 B/op	 1032412 allocs/op
BenchmarkTablesWithChildrenDefaultConcurrency
BenchmarkTablesWithChildrenDefaultConcurrency-8   	       1	       544.6 resources/s	     40606 targetResources/s	461292456 B/op	 6689992 allocs/op
PASS
```

The output is different from normal Go benchmarks, and I feel it better fits our use-case. It shows the actual `resources/s`, similar to what the CLI would show. It also shows a `targetResources/s`, an attempt to calculate the theoretical best value we could aim for. 

As the above results show, we are doing pretty well in the case with no child relations, but we are leaving a lot of room for improvement when child relations are involved. In theory, these could be sped up by almost 100x (this might use a lot of memory though, of course). In both cases it seems like we could also be doing better in terms of number of allocs and B/op, but that is pending further research.

Closes https://github.com/cloudquery/cloudquery/issues/3721

### DIFF
--- a/plugins/benchmark_test.go
+++ b/plugins/benchmark_test.go
@@ -113,7 +113,10 @@ func (s *SourceBenchmark) setup(b *testing.B) {
 	}
 	columnResolver := func(ctx context.Context, meta schema.ClientMeta, resource *schema.Resource, c schema.Column) error {
 		s.simulateAPICall(s.ResolverMin, s.ResolverStdDev, s.ResolverMean)
-		resource.Set(c.Name, "test")
+		err := resource.Set(c.Name, "test")
+		if err != nil {
+			return err
+		}
 		return nil
 	}
 	s.tables = make([]*schema.Table, s.Tables)
--- a/plugins/benchmark_test.go
+++ b/plugins/benchmark_test.go
@@ -113,11 +113,7 @@ func (s *SourceBenchmark) setup(b *testing.B) {
 	}
 	columnResolver := func(ctx context.Context, meta schema.ClientMeta, resource *schema.Resource, c schema.Column) error {
 		s.simulateAPICall(s.ResolverMin, s.ResolverStdDev, s.ResolverMean)
-		err := resource.Set(c.Name, "test")
-		if err != nil {
-			return err
-		}
-		return nil
+		return resource.Set(c.Name, "test")
 	}
 	s.tables = make([]*schema.Table, s.Tables)
 	for i := 0; i < s.Tables; i++ {
--- a/plugins/benchmark_test.go
+++ b/plugins/benchmark_test.go
@@ -212,6 +212,8 @@ func (s *SourceBenchmark) Run() {
 	s.b.ReportMetric(0, "ns/op") // drop default ns/op output
 	s.b.ReportMetric(float64(totalResources)/(end.Sub(start).Seconds()), "resources/s")
 	s.b.ReportMetric(float64(totalResources)/s.lowerBound().Seconds(), "targetResources/s")
+
+	// Enable the below metrics for more verbose information about the scenario:
 	//s.b.ReportMetric(float64(totalResources), "resources")
 	//s.b.ReportMetric(float64(s.apiCalls.Load()), "apiCalls")
 }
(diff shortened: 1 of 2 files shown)


disq:
```
goos: darwin
goarch: arm64
pkg: github.com/cloudquery/plugin-sdk/plugins
BenchmarkDefaultConcurrency
BenchmarkDefaultConcurrency-10                      	       1	     11994 resources/s	     12626 targetResources/s	73615560 B/op	 1032619 allocs/op
BenchmarkTablesWithChildrenDefaultConcurrency
BenchmarkTablesWithChildrenDefaultConcurrency-10    	       1	       540.1 resources/s	     40606 targetResources/s	461383256 B/op	 6690320 allocs/op
PASS
```

Same `targetResources` result? that's weird?

hermanschaaf:
@disq That's expected - it's a theoretical maximum we calculate, so you have something to compare the `resources/s` value to. It'll be unique for every test scenario, but it should never change (unless we refine the calculation)

plugins/benchmark_test.go
```diff
@@ -0,0 +1,322 @@
...
+		resource.Item = struct {
+			Column1 string
+		}{
+			Column1: "test-pre",
+		}
+		return nil
+	}
+	columnResolver := func(ctx context.Context, meta schema.ClientMeta, resource *schema.Resource, c schema.Column) error {
+		s.simulateAPICall(s.ResolverMin, s.ResolverStdDev, s.ResolverMean)
+		err := resource.Set(c.Name, "test")
```
disq, line 116: why not just `return resource.Set(...`

hermanschaaf: meh, yeah. Was just trying to satisfy the linter, this should never really return an error

hermanschaaf: Updated 👍

plugins/benchmark_test.go
```diff
@@ -0,0 +1,318 @@
...
+		return nil
+	}
+	columnResolver := func(ctx context.Context, meta schema.ClientMeta, resource *schema.Resource, c schema.Column) error {
+		s.simulateAPICall(s.ResolverMin, s.ResolverStdDev, s.ResolverMean)
+		return resource.Set(c.Name, "test")
+	}
+	s.tables = make([]*schema.Table, s.Tables)
+	for i := 0; i < s.Tables; i++ {
+		columns := make([]schema.Column, s.Columns)
+		for u := 0; u < s.Columns; u++ {
```
disq, line 121: Traditional second index variable is `j` but you chose `u`? interesting 👀

hermanschaaf: Hmm, now you're making me self-conscious. I've been using `u` for much of my life for some reason - I think basically just that it's next to `i` on the keyboard 😂

disq:
> @disq That's expected - it's a theoretical maximum we calculate, so you have something to compare the `resources/s` value to. It'll be unique for every test scenario, but it should never change (unless we refine the calculation)

Yes interesting calculation 👀

plugins/benchmark_test.go
```diff
@@ -175,42 +175,46 @@ func min(a, b int) int {
...
+		totalResources := 0
+		for range resources {
+			// read resources channel until empty
+			totalResources++
+		}
+		if err := g.Wait(); err != nil {
+			s.b.Fatal(err)
+		}
+
+		end := time.Now()
```
disq, line 209: `diff := time.Since(start)` (there's no call to `s.b.StopTimer()` now?)

yevgenypats (approved):
Looks great! we can add benchtest in follow-up PR.

Issue #358, opened 2022-11-07
feat: Resolve relation tables in parallel


erezrokah:
TDLR: run the following line in a new Go routine:
https://github.com/cloudquery/plugin-sdk/blob/014a2df1bf61c5b8c5669470371d90a6c646a23a/plugins/source_scheduler_dfs.go#L140

so relations are resolved in parallel instead of synchronously.

We should limit the number of Go routines based on the `Concurrency` configuration

hermanschaaf:
Relates to: https://github.com/cloudquery/cloudquery/issues/3721 - would be good to have that first so we can measure the impact on both performance and memory.

yevgenypats:
We need to support it but it's not as easy as just running the goroutine because we need to divide goroutines into each level otherwise it will get deadlock. and the performance might benchmark might be a good idea indeed to understand what is a good scheduling (i.e how. to divide it best)

disq:
We could use Log to reduce concurrency exponentially at each level. 1M top-level goroutines would give us 20 goroutines for Level 1, which would give us 4 goroutines for Level 2... Level 3 would get 2 goroutines if we ever had a Level 3, we rarely do. If the result < 2, set the goroutine count to 2. So further levels down still get 2 goroutines.

I'm not sure if we need a hard limit, we could introduce a feature flag removing the hard limit so that goroutines used in the lower levels are not reduced from the initial 1M. Another solution would be to have a `relational_concurrency` and start from that.

PR #416, opened 2022-11-18
feat: Resolve table relations in parallel

hermanschaaf:
This changes the resolver for table relations to work concurrently, like parent tables. To do so safely and without the risk of deadlock, we instantiate one semaphore per depth level. (The size of the semaphore is decreased logarithmically for each depth.)

This change will only improve performance for tables with child relations.

To compare the performance before and after, I used the benchmarks in https://github.com/cloudquery/plugin-sdk/pull/415

## Before
```
goos: darwin
goarch: arm64
pkg: github.com/cloudquery/plugin-sdk/plugins
BenchmarkDefaultConcurrency-8                     	       1	     11957 resources/s	     12626 targetResources/s	73597280 B/op	 1032425 allocs/op
BenchmarkTablesWithChildrenDefaultConcurrency-8   	       1	       545.9 resources/s	     40606 targetResources/s	461622584 B/op	 6690105 allocs/op
PASS
ok  	github.com/cloudquery/plugin-sdk/plugins	150.596s
```

## After
```
goos: darwin
goarch: arm64
pkg: github.com/cloudquery/plugin-sdk/plugins
BenchmarkDefaultConcurrency-8                     	       1	     11373 resources/s	     12626 targetResources/s	73692608 B/op	 1033614 allocs/op
BenchmarkTablesWithChildrenDefaultConcurrency-8   	       1	     30162 resources/s	     40606 targetResources/s	464285672 B/op	 6697508 allocs/op
PASS
ok  	github.com/cloudquery/plugin-sdk/plugins	6.241s
```

## Analysis

This change focuses on the `BenchmarkTablesWithChildrenDefaultConcurrency` case. The change improves `resources/s` from 545.9 to 30162, an improvement of about 55x. Memory and allocs are mostly the same.

The small downward change in `resources/s` in the `BenchmarkDefaultConcurrency` case is likely due to the few additional allocs needed. 

Closes https://github.com/cloudquery/plugin-sdk/issues/358

### DIFF
--- a/plugins/source.go
+++ b/plugins/source.go
@@ -32,7 +32,7 @@ type SourcePlugin struct {
 	// resourceSem is a semaphore that limits the number of concurrent resources being fetched
 	resourceSem *semaphore.Weighted
 	// tableSem is a semaphore that limits the number of concurrent tables being fetched
-	tableSem *semaphore.Weighted
+	tableSems []*semaphore.Weighted
 	// maxDepth is the max depth of tables
 	maxDepth uint64
 	// caser
@@ -101,7 +101,6 @@ func NewSourcePlugin(name string, version string, tables []*schema.Table, newExe
 	if p.maxDepth > maxAllowedDepth {
 		panic(fmt.Errorf("max depth of tables is %d, max allowed is %d", p.maxDepth, maxAllowedDepth))
 	}
-
 	return &p
 }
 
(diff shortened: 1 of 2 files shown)


yevgenypats:
@hermanschaaf Really awesome tests! Im still reviewing and playing with it as quite a lot to intake but posting some thoughts in the meanwhile. Maybe add golang.org/x/perf/cmd/benchstat to also compare those in CI and catch if anything change the performance significantly as otherwise we would forget to run it

yevgenypats:
> some tests! Im still reviewing and playing with it as quite a lot to intake but posting some thoughts in the meanwhile. Maybe add golang.org/x/perf/cmd/benchstat to also compare those in CI and catch if anything change the performance significantly as otherwise we would forget to run it

comment on the wrong thread :)

yevgenypats (approved):
Nice!

PR #418, opened 2022-11-20
feat: Support ${file:./path} expansion in spec

yevgenypats:
Per user request. we had a similar version of that in v0 but now it is generic and available to all specs and not only to a specific destination.

### DIFF
--- a/specs/spec_reader.go
+++ b/specs/spec_reader.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 )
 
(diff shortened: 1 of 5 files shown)


specs/spec_reader_test.go
```diff
@@ -93,3 +94,51 @@ func TestLoadSpecs(t *testing.T) {
...
+kind: source
+spec:
+	name: test
+	version: v1.0.0
+	spec:
+		credentials: ${file:./testdata/creds2.txt}
+		otherstuff: 2
+	`)
+	_, err = expandFileConfig(badCfg)
+	expectedErr := `open ./testdata/creds2.txt: no such file or directory`
```
disq, line 137: Why not use `errors.Is` and ErrNotExist etc.

yevgenypats: yeah that would be better. I basically copy pasted from other tests :)

specs/spec_reader.go
```diff
@@ -12,11 +13,31 @@ type SpecReader struct {
 	Destinations map[string]*Destination
 }
 
+var fileRegex = regexp.MustCompile(`\$\{file:(.+)\}`)
```
hermanschaaf, line 16:
```suggestion
var fileRegex = regexp.MustCompile(`\$\{file:([^}]+)\}`)
```

I think this regex should also handle the case where multiple variable replacements occur in the same line. 

E.g. before:

```
${file:abc}-${file:def}
```

would greedily match `abc}-${file:def` as the filename, instead of two separate matches `abc` and `def`. (Assuming this is a valid)

We may want to add a test case where two variable replacements happen as well (maybe one file replacement and one normal replacement)

hermanschaaf (changes requested):

specs/spec_reader.go
yevgenypats: Done.

disq: Why not do `var fileRegex = regexp.MustCompile(`\$\{file:(.+?)\}`)` for an ungreedy match?

hermanschaaf (approved):

disq (approved):

PR #421, opened 2022-11-22
fix: Skip very large GRPC messages, log when it happens

hermanschaaf:
Context: https://github.com/cloudquery/cloudquery/issues/4834

This change:
 - bumps the gRPC max message size from 50MiB to 100MiB
 - logs the name of the table to Sentry when we see a message over 50MiB
 - logs to the user if we drop a message because it exceeds 100MiB

### DIFF
--- a/clients/constants.go
+++ b/clients/constants.go
@@ -1,5 +1,5 @@
 package clients
 
 const (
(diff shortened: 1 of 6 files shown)


yevgenypats (approved):
Nice

PR #426, opened 2022-11-22
chore: Add more options to destination testing

yevgenypats:
Added another option needed for destination testing. Needed by this PR: https://github.com/cloudquery/cloudquery/pull/4902

### DIFF
--- a/plugins/destination_testing.go
+++ b/plugins/destination_testing.go
@@ -22,6 +22,7 @@ type DestinationTestSuiteTests struct {
 	SkipOverwrite   bool
 	SkipDeleteStale bool
 	SkipAppend      bool
+	SkipAppendTwice	bool
 }
 
 func getTestLogger(t *testing.T) zerolog.Logger {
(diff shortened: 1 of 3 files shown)


schema/table.go
```diff
@@ -65,6 +65,15 @@ var (
 	reValidColumnName = regexp.MustCompile(`^[a-z_][a-z\d_]*$`)
 )
 
+func (tt Tables) FlattenTables() Tables {
```
hermanschaaf, line 68: Looks good, but we should probably have a small test for this functionality :)

plugins/destination_testing.go
```diff
@@ -22,6 +22,7 @@ type DestinationTestSuiteTests struct {
 	SkipOverwrite   bool
 	SkipDeleteStale bool
 	SkipAppend      bool
+	SkipAppendTwice bool
```
hermanschaaf, line 25: Could we add a comment to explain what this option does? The others are fairly self-explanatory, but I'm not sure about this one

hermanschaaf: Oh okay, reading the code, maybe it could be called `SkipSecondAppend`?

hermanschaaf (changes requested):

plugins/destination_testing.go
hermanschaaf: Also, I'm curious why this is useful as a test case 🤔

yevgenypats: It is useful in cases like cloud storage where you can't append to an existing object the moment you closed the file as it's immutable while locally you can and of-course for PostgreSQL it's all valid.

hermanschaaf (approved):

PR #430, opened 2022-11-24
feat: Handle resolving of empty maps and slices

shimonp21:
In fields like 'Tags', 'Annotations' - it is more consistent if empty/null dictionaries show up as an empty JSON object `{}`, rather than being a postgres "null" or a JSON "null".

#### Summary

---

- [ ] Update or add tests 🧪

### DIFF
--- a/schema/json.go
+++ b/schema/json.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"errors"
+	"reflect"
 )
 
 type JSONTransformer interface {
@@ -63,6 +64,7 @@ func (dst *JSON) Set(src interface{}) error {
 		} else {
 			*dst = JSON{Bytes: value, Status: Present}
 		}
+
 	// Encode* methods are defined on *JSON. If JSON is passed directly then the
 	// struct itself would be encoded instead of Bytes. This is clearly a footgun
 	// so detect and return an error. See https://github.com/jackc/pgx/issues/350.
@@ -74,6 +76,20 @@ func (dst *JSON) Set(src interface{}) error {
 		if err != nil {
 			return err
 		}
+
+		// For map and slice jsons, it is easier for users to work with '[]' or '{}' instead of JSON's 'null'.
+		if bytes.Equal(buf, []byte(`null`)) {
+			if isEmptyStringMap(value) {
+				*dst = JSON{Bytes: []byte("{}"), Status: Present}
+				return nil
+			}
+
+			if isEmptySlice(value) {
+				*dst = JSON{Bytes: []byte("[]"), Status: Present}
+				return nil
+			}
+		}
+
 		*dst = JSON{Bytes: buf, Status: Present}
 	}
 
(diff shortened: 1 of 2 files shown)


schema/json.go
```diff
@@ -63,6 +63,19 @@ func (dst *JSON) Set(src interface{}) error {
...
 			*dst = JSON{Bytes: value, Status: Present}
 		}
+	// In the case of map[string]string (fields like 'Tags'), an empty map is easier for users to work with than
+	// postgres "null", or JSON "null".
+	case map[string]string:
+		if len(value) == 0 {
+			*dst = JSON{Bytes: []byte("{}"), Status: Present}
+			break
+		}
+		buf, err := json.Marshal(value)
```
disq, line 73: You could move this `case` just before the `default` and use `fallthrough` keyword but would make it harder to read and reason/refactor later on...

disq (approved):

schema/json.go
```diff
@@ -63,6 +63,19 @@ func (dst *JSON) Set(src interface{}) error {
 		} else {
 			*dst = JSON{Bytes: value, Status: Present}
 		}
+	// In the case of map[string]string (fields like 'Tags'), an empty map is easier for users to work with than
+	// postgres "null", or JSON "null".
+	case map[string]string:
```
yevgenypats, line 68: I think this is not generic enough because it can be `map[string]interface` and can be []interface{} and can be `[]string` all valid jsons.

we should then solve it on the `default` i.e to check if null and then make it `{}`

yevgenypats (changes requested):
Looks good but it needs to be implemented in different place.

schema/json.go
shimonp21: Agree :)

schema/json.go
shimonp21: PTAL at new version - more generic

schema/json.go
```diff
@@ -63,13 +64,24 @@ func (dst *JSON) Set(src interface{}) error {
...
+			*dst = JSON{Bytes: []byte("{}"), Status: Present}
+			break
+		}
+
+		if isEmptySlice(value) {
+			*dst = JSON{Bytes: []byte("[]"), Status: Present}
+			break
+		}
+
 		buf, err := json.Marshal(value)
```
yevgenypats, line 85: Let's do a small improvement in order:

1: do the `marshal` first. 
2: if error return error
3: check if the `buff` is null
4) if null do this check and turn `null` into `{}` or `[]`

yevgenypats (changes requested):
looks good. One more improvement.

schema/json.go
```diff
@@ -95,3 +111,29 @@ func (dst JSON) Get() interface{} {
 		return dst.Status
 	}
 }
+
+// isEmptyStringMap returns true if the value is a map from string to any (i.e. map[string]interface{}).
+// We need to use reflection for this, because it impossible to type-assert a map[string]string into a
```
hermanschaaf, line 116:
```suggestion
// We need to use reflection for this, because it is impossible to type-assert a map[string]string into a
```

hermanschaaf (approved):
With this change, is it possible to ever get a `null` entry in a column that has type `map` or `slice`? I guess not, so any queries checking for `null` to see if a particular config exists will need to be updated. That said, I think this change will solve more issues than it will create, and we have to choose one way or another, so I say we go for it.

schema/json.go
```diff
@@ -74,6 +76,20 @@ func (dst *JSON) Set(src interface{}) error {
 		if err != nil {
 			return err
 		}
+
+		// For map and slice jsons, it is easier for users to work with '[]' or '{}' instead of JSON's 'null'.
+		if bytes.Equal(buf, []byte(`null`)) {
+			if isEmptyStringMap(value) {
+				*dst = JSON{Bytes: []byte("{}"), Status: Present}
+				break
```
yevgenypats, line 84: Does this break works? should be return imo as otherwise it will override itself with line 93? in any case lets use return.

yevgenypats (approved):

yevgenypats (changes requested):
wanted to request small change

yevgenypats (approved):

candiduslynx (approved):

PR #431, opened 2022-11-24
fix(test): Values check test should account for `IgnoreInTests` column option

candiduslynx:


hermanschaaf (approved):

hermanschaaf (approved):

plugins/source_testing.go
```diff
@@ -107,8 +107,15 @@ func validateResources(t *testing.T, resources []*schema.Resource) {
...
-		if !hasValue && !(table.Columns[i].Name == "_cq_parent_id" && table.Parent == nil) {
-			t.Errorf("table: %s column %s has no values", table.Name, table.Columns[i].Name)
+		col := table.Columns[i]
+		switch {
+		case col.IgnoreInTests,
+			hasValue,
+			col.Name == "_cq_parent_id" && table.Parent != nil:
+		// nop
+		default:
+			return fmt.Errorf("table: %s column %s has no values", table.Name, table.Columns[i].Name)
```
yevgenypats, line 117: I don't think it should return an error but rather call `t.Errorf`

yevgenypats: Also I don't think a switch case is needed here but better just a simple `if`

yevgenypats (changes requested):
almost there.

plugins/source_testing.go
candiduslynx: > call `t.Errorf`
It makes impossible creating a negative test

> simple `if`
The `if` clause started to look too overloaded

plugins/source_testing.go
```diff
@@ -107,8 +116,15 @@ func validateResources(t *testing.T, resources []*schema.Resource) {
...
-			t.Errorf("table: %s column %s has no values", table.Name, table.Columns[i].Name)
+		col := table.Columns[i]
+		switch {
+		case col.IgnoreInTests,
+			hasValue,
+			col.Name == "_cq_parent_id" && table.Parent != nil:
+		// nop
+		default:
+			errs = append(errs, fmt.Errorf("table: %s column %s has no values", table.Name, table.Columns[i].Name))
 		}
```
hermanschaaf, line 127:
```suggestion
		emptyExpected := col.Name == "_cq_parent_id" && table.Parent == nil
		if !col.IgnoreInTests && !emptyExpected && !hasValue {
		errs = append(errs, fmt.Errorf("table: %s column %s has no values", table.Name, table.Columns[i].Name))
		}
```

Maybe this ☝️  (I think this also fixes a bug related to the parent columns being null?)

yevgenypats (changes requested):
Few issues:
* no need to use `testify` package
* no need to change function signature. just call `t.Errorf` directly instead of passing errors around.
* use simple `if else` instead of switch

yevgenypats (approved):

hermanschaaf (approved):

PR #437, opened 2022-11-25
chore: Remove benchmark comments for now

hermanschaaf:
The fix in https://github.com/cloudquery/plugin-sdk/pull/436 didn't work; looks like you can't use the GH bot token from a forked repo PR either.

### DIFF
--- a/.github/workflows/unittest.yml
+++ b/.github/workflows/unittest.yml
@@ -31,12 +31,13 @@ jobs:
       - name: Run benchmark
         if: matrix.os == 'ubuntu-latest'
         run: make benchmark-ci
-      - name: Delta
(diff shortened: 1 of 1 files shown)


yevgenypats (approved):

danez:
One solution could be to do what we (netlify) do here in this repo. It is a little bit complicated for what it does in the end but it works.

It calculates the delta and uploads it, then uses the workflow_run event and posts the comment.
https://github.com/netlify/cli/blob/main/.github/workflows/benchmark.yml
https://github.com/netlify/cli/blob/main/.github/workflows/benchmark-post.yml

erezrokah:
Thanks @danez, that should work well for us 🥇

PR #443, opened 2022-11-29
fix: Add GetFlatTableList helper

shimonp21:
#### Summary

Required to fix DeleteStale bug.

---

- [ ] Update or add tests 🧪

### DIFF
--- a/helpers/table.go
+++ b/helpers/table.go
@@ -1,0 +1,16 @@
+package helpers
+
+import "github.com/cloudquery/plugin-sdk/schema"
+
+// GetFlatTablesList returns a flat list of *Table objects from a non-flat *Table slice.
+// This doesn't modify the Table objects themselves! Neither should the caller!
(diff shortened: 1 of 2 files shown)


yevgenypats (changes requested):
There is already such function here - https://github.com/cloudquery/plugin-sdk/blob/main/schema/table.go#L68

disq:
Why not recurse in the DeleteStale call? Could even make an intermediary fn that creates larger batches (take a batch, iterate given schema.Tables, add to the batch and return, without SendBatch... so that multiple levels of tables can be batched)

erezrokah:
> There is already such function here - https://github.com/cloudquery/plugin-sdk/blob/main/schema/table.go#L68

And also `GlobMatch` with `*` should work? https://github.com/cloudquery/plugin-sdk/blob/c8c824d0c1842e6b85a6e0d4e0e4102dc1a5ab50/schema/table.go#L111

shimonp21:
Closing this one :)

PR #446, opened 2022-11-30
chore: Refactor some goroutine-related code

hermanschaaf:
Refactoring some code that I initially thought had a bug, but actually didn't. 😄

### DIFF
--- a/internal/servers/source.go
+++ b/internal/servers/source.go
@@ -13,6 +13,7 @@ import (
 	"github.com/cloudquery/plugin-sdk/specs"
 	"github.com/getsentry/sentry-go"
 	"github.com/rs/zerolog"
+	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/proto"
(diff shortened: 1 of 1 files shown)


erezrokah (approved):

disq (approved):
🤦

yevgenypats (changes requested):
I can see how this is a potential improvement but Im not sure there was any race condition here.  the `return` statement would only be called after `close(resources)` would be called and this would only happen after `syncErr` is called so Im requesting changes mostly for a discussion as this can also work but I dont think there was any bug before that.

hermanschaaf:
@yevgenypats Ah, you're absolutely right. I spent too much time staring at this code today, beginning to see bugs where there aren't any 😅 Yeah, looks like there was no bug here, because the deferred closing should mean the outer function always finishes after the variable was already set by the inner goroutine.

Happy to close this, but maybe for future readers this updated version requires a bit less thinking to understand, so I'll leave it up to you whether you think it's an improvement or not. Marking the PR as `chore` in any case.

yevgenypats:
I think in that case let's close it for now given we don't want to touch something that works especially around tricky areas to avoid introducing new bugs :) .

PR #447, opened 2022-11-30
fix(codegen)!: Column type for slices

candiduslynx:
#### Summary

* `[]*string` will become `schema.TypeStringArray`
* `[]*int(8|16|32|64)` will become `schema.TypeIntArray`

---

- [x] Update or add tests 🧪

### DIFF
--- a/codegen/table_test.go
+++ b/codegen/table_test.go
@@ -32,7 +32,7 @@ type (
 			StringCol string `json:"string_col,omitempty"`
 		}
 		IntArrayCol    []int  `json:"int_array_col,omitempty"`
-		IntArrayColToo []*int `json:"int_array_col_too,omitempty"`
+		IntPointerArrayCol []*int `json:"int_pointer_array_col,omitempty"`
 
 		StringArrayCol    []string  `json:"string_array_col,omitempty"`
 		StringArrayColToo []*string `json:"string_array_col_too,omitempty"`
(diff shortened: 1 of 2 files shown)


codegen/table_test.go
```diff
@@ -31,8 +31,12 @@ type (
 			IntCol    int    `json:"int_col,omitempty"`
 			StringCol string `json:"string_col,omitempty"`
 		}
-		IntArrayCol    []int      `json:"int_array_col,omitempty"`
-		StringArrayCol []string   `json:"string_array_col,omitempty"`
+		IntArrayCol    []int  `json:"int_array_col,omitempty"`
+		IntArrayColToo []*int `json:"int_array_col_too,omitempty"`
```
erezrokah, line 35:
```suggestion
		IntPointerArrayCol []*int `json:"int_pointer_array_col,omitempty"`
```

[non blocking] Since the convention is to use the type of the field, maybe keep that? See `TimePointerCol *time.Time` a few lines below.

Also might be better to use a name that describes the field, and not what's it's meant to be transformed to

codegen/table_test.go
```diff
@@ -31,8 +31,12 @@ type (
...
+
+		StringArrayCol    []string  `json:"string_array_col,omitempty"`
+		StringArrayColToo []*string `json:"string_array_col_too,omitempty"`
```
erezrokah, line 38:
```suggestion
		StringPointerArrayColToo []*string `json:"string_pointer_array_col,omitempty"`
```

Similar here

erezrokah (approved):
This looks good, added a non blocking comment.
Please merge this as `fix(codegen)!` as this is a breaking change

yevgenypats (approved):

PR #453, opened 2022-12-01
fix: Add error-handling to transformer

shimonp21:
#### Summary

Required to fix batch-write-failures due to postgresql TEXT fields containing NULL bytes.

```
2022-11-30T12:00:11Z ERR failed to execute batch with pgerror error="ERROR: invalid byte sequence for encoding \"UTF8\": 0x00 (SQLSTATE 22021)" module=pg-dest
```


yevgenypats (changes requested):
Good find. I think we should take a different approach here see  comment: https://github.com/cloudquery/cloudquery/pull/5249#discussion_r1036761693

shimonp21:
Closing in favor of https://github.com/cloudquery/cloudquery/pull/5249

PR #455, opened 2022-12-01
fix: Sync Metrics for Long running plugins

bbernays:
#### Summary

Prior to this change resource, error, and panic count were only reset when a plugin was relaunched. This means if a plugin is being run via direct gRPC calls then the summary was a total of all of the runs.

This PR reinitializes the metrics for each sync

---

- [ ] Update or add tests 🧪

### DIFF
--- a/plugins/source_metrics.go
+++ b/plugins/source_metrics.go
@@ -54,15 +54,15 @@ func (s *SourceMetrics) Equal(other *SourceMetrics) bool {
 }
 
 func (s *SourceMetrics) initWithTables(tables schema.Tables) {
+	s.TableClient = make(map[string]map[string]*TableClientMetrics)
 	for _, table := range tables {
-		if _, ok := s.TableClient[table.Name]; !ok {
-			s.TableClient[table.Name] = make(map[string]*TableClientMetrics)
-		}
+		s.TableClient[table.Name] = make(map[string]*TableClientMetrics)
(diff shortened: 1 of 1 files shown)


yevgenypats (approved):

plugins/source_metrics.go
```diff
@@ -64,9 +64,7 @@ func (s *SourceMetrics) initWithTables(tables schema.Tables) {
 
 func (s *SourceMetrics) initWithClients(table *schema.Table, clients []schema.ClientMeta) {
 	for _, client := range clients {
-		if _, ok := s.TableClient[table.Name][client.ID()]; !ok {
-			s.TableClient[table.Name][client.ID()] = &TableClientMetrics{}
-		}
+		s.TableClient[table.Name][client.ID()] = &TableClientMetrics{}
```
hermanschaaf, line 67: Should this then technically also be clearing the map before it loops over the clients, ie. `s.TableClient[table.Name] = make(map[string]*TableClientMetrics)` so that it's impossible to end up with outdated clients in the map?

hermanschaaf: (We may want to also make a similar fix to `initWithTables` above)

bbernays: Good catches! Have incorporated both into the pr

hermanschaaf (approved):

yevgenypats (approved):

PR #456, opened 2022-12-02
fix: TablesForSpec should only return top-level tables

hermanschaaf:
Related to https://github.com/cloudquery/cloudquery/pull/5294


candiduslynx (approved):

disq (approved):

hermanschaaf:
> It includes the relations in the parent table when going over the wire right?

@yevgenypats Right 👍

PR #460, opened 2022-12-05
fix(destinations): Stop writing resources when channel is closed

erezrokah:
#### Summary

Related to https://github.com/cloudquery/cloudquery/issues/5152.

This fixes an issue when if the channel gets closed (for example there's an error during write), we report the `EOF` error instead of the original error.

Scenario from the issue:
1. Users runs `sync --no-migrate` on a non initialized database
2. [`Write`](https://github.com/cloudquery/plugin-sdk/blob/ab7ca972e0b187a7dfb66132a03f07479cd29bb7/internal/servers/destinations.go#L84) fails on the first table write, returning an error from this function:
https://github.com/cloudquery/plugin-sdk/blob/ab7ca972e0b187a7dfb66132a03f07479cd29bb7/internal/servers/destinations.go#L65, causing the channel to close
3.  The next iteration of `for resource := range resources` starts, receiving an `EOF` error since the channel is closed, and `EOF` error is reported.

> This is only a patch. I think we should separate application level errors (e.g. write) from protocol/communication level errors. I'll open a separate issue for that

**Also, [another related fix coming](https://github.com/cloudquery/plugin-sdk/pull/461)**

---

- [x] Update or add tests 🧪


clients/destination.go
```diff
@@ -264,6 +264,10 @@ func (c *DestinationClient) Write(ctx context.Context, source string, syncTime t
 			Source:    source,
 			Timestamp: timestamppb.New(syncTime),
 		}); err != nil {
+			if err == io.EOF {
```
yevgenypats, line 267: Im not sure i understand what this solved. give in othercase this just returns an error and doesn't write to the channel while with break it actually sends `CloseAndRecv`.

yevgenypats (changes requested):
Had one question given I didn't understand the fix.

Also, I don't know if possible but given that those affect all plugin can we add tests to that (e2e in memory tests available here https://github.com/cloudquery/plugin-sdk/tree/main/serve) ?

clients/destination.go
erezrokah: `Write` works in stream mode (it's not a synchronous RPC), so if we want the correct response/error from the write request we need to call `RecvMsg` which is only done in `CloseAndRecv` (and only once actually).

So `SendMsg` will never return application level error as it doesn't wait for the message to be received by the server.
Technically I think we should call `RecvMsg` for each `SendMsg` per the docs:
> SendMsg does not wait until the message is received by the server. An untimely stream closure may result in lost messages. To ensure delivery, users should ensure the RPC completed successfully using RecvMsg.

erezrokah:
> can we add tests to that (e2e in memory tests available here https://github.com/cloudquery/plugin-sdk/tree/main/serve) ?

Sure I'll see how we can add a test for this

erezrokah:
@yevgenypats added a test in https://github.com/cloudquery/plugin-sdk/pull/460/commits/03ab5d992b2cb76222aab82aaba3f73d39568ead.
Note sure how to use the memory plugin for this, as in order to reproduce the issue we need to set up both a client and a server (I guess I can create a test method that starts the memory plugin as a local binary).

Without the fix the error will just be `"EOF"`.

Once https://github.com/cloudquery/plugin-sdk/pull/461 is merged the error should change from `context canceled` to `table not exists`

clients/destination.go
```diff
@@ -295,6 +299,10 @@ func (c *DestinationClient) Write2(ctx context.Context, tables schema.Tables, so
 		if err := saveClient.Send(&pb.Write2_Request{
 			Resource: resource,
 		}); err != nil {
+			if err == io.EOF {
```
candiduslynx, line 302:
```suggestion
			if errors.Is(err, io.EOF) {
```
`errors.Is` is much more reliable as the errors are wrapped more and more often.

clients/destination.go
candiduslynx, line 267:
```suggestion
			if errors.Is(err, io.EOF) {
```
`errors.Is` is much more reliable as the errors are wrapped more and more often.

candiduslynx (changes requested):

clients/destination.go
erezrokah: I don't think we need to be overly defensive in this case, as checking `== io.EOF` is the officially documented way to check if the stream has ended, see https://grpc.io/docs/languages/go/basics/#client-side-streaming-rpc

erezrokah:
Still happens and reported on Discord https://discord.com/channels/872925471417962546/873606591335759872/1057191097648033833

yevgenypats (approved):

PR #463, opened 2022-12-06
refactor: Reduce nesting in destination server code `write`

erezrokah:
#### Summary

We can check `err == io.EOF` outside of `err != nil` to make it easier to follow the code

---

- [x] Update or add tests 🧪


hermanschaaf (approved):

disq (approved):

internal/servers/destinations.go
```diff
@@ -86,14 +86,14 @@ func (s *DestinationServer) Write2(msg pb.Destination_Write2Server) error {
...
 	for {
 		r, err := msg.Recv()
-		if err != nil {
-			if err == io.EOF {
-				close(resources)
-				if err := eg.Wait(); err != nil {
-					return fmt.Errorf("got EOF. failed to wait for plugin: %w", err)
-				}
-				return msg.SendAndClose(&pb.Write2_Response{})
+		if err == io.EOF {
```
candiduslynx, line 89:
```suggestion
		if errors.Is(err, io.EOF) {
```
`errors.Is` is much more reliable as the errors are wrapped more and more often.

internal/servers/destinations.go
```diff
@@ -86,14 +86,14 @@ func (s *DestinationServer) Write2(msg pb.Destination_Write2Server) error {
...
+			close(resources)
```
candiduslynx, line 90: Let's move this outside the `for` loop and do a `defer` on it?
All exit routes do it, so I see no point in several places doing it.

internal/servers/destinations.go
```diff
@@ -86,14 +86,14 @@ func (s *DestinationServer) Write2(msg pb.Destination_Write2Server) error {
...
+			if err := eg.Wait(); err != nil {
+				return fmt.Errorf("got EOF. failed to wait for plugin: %w", err)
 			}
+			return msg.SendAndClose(&pb.Write2_Response{})
+		}
+		if err != nil {
```
candiduslynx, line 96: Maybe we can do this with a `switch`?
```go
switch {
	case errors.Is(err, io.EOF):
		// finish
	case err != nil:
		// error handling
	default:
		// process
}
```

candiduslynx (changes requested):

internal/servers/destinations.go
erezrokah: Maybe we can do each refactor in its own PR? That would make it easier to review and keep the changes in each PR small

candiduslynx: Ass this PR is about restructuring code for optimizing nesting, I think moving chan close outside would also allow to optimize the same code a bit better.

erezrokah: Sure it's about refactoring, but if we can split the refactoring into multiple smaller PRs, is there a reason not to do it?

candiduslynx (approved):

candiduslynx (approved):

erezrokah:
Done in https://github.com/cloudquery/plugin-sdk/pull/529

PR #464, opened 2022-12-06
test: Fix destination test

shimonp21:
#### Summary

Fixes all the tests failing silently due to a missing `!` https://github.com/cloudquery/plugin-sdk/blob/538211c863ec2d4b719b83086a842e89ecc396d3/plugins/destination_testing.go#L91. 

Works on https://github.com/cloudquery/cloudquery/issues/5248.

### DIFF
--- a/plugins/destination.go
+++ b/plugins/destination.go
@@ -88,6 +88,7 @@ func (p *DestinationPlugin) Migrate(ctx context.Context, tables schema.Tables) e
 	return p.client.Migrate(ctx, tables)
 }
 
+// nolint: unparam
 func (p *DestinationPlugin) readAll(ctx context.Context, table *schema.Table, sourceName string) ([]schema.CQTypes, error) {
 	var readErr error
 	ch := make(chan schema.CQTypes)
--- a/plugins/destination_testing.go
+++ b/plugins/destination_testing.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"sort"
 	"testing"
 	"time"
 
@@ -30,11 +31,6 @@ type DestinationTestSuiteTests struct {
 	// SkipAppend skips testing for "append" mode. Use if the destination
 	// plugin doesn't support this feature.
 	SkipAppend bool
-
-	// SkipSecondAppend skips the second append step in the test.
-	// This is useful in cases like cloud storage where you can't append to an
-	// existing object after the file has been closed.
-	SkipSecondAppend bool
 }
 
 func getTestLogger(t *testing.T) zerolog.Logger {
@@ -46,6 +42,7 @@ func getTestLogger(t *testing.T) zerolog.Logger {
 }
 
 func (s *destinationTestSuite) destinationPluginTestWriteOverwrite(ctx context.Context, p *DestinationPlugin, logger zerolog.Logger, spec specs.Destination) error {
+	// ----------------------- Write two resources -----------------------
 	spec.WriteMode = specs.WriteModeOverwrite
 	if err := p.Init(ctx, logger, spec); err != nil {
 		return fmt.Errorf("failed to init plugin: %w", err)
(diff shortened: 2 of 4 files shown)


plugins/destination_testing.go
```diff
@@ -111,36 +149,49 @@ func (s *destinationTestSuite) destinationPluginTestWriteOverwrite(ctx context.C
...
-	}
+	sort.Slice(resourcesRead, func(i, j int) bool {
+		// We add '+2' to uuidFieldIndex because we add 2 columns (_cq_sync_time and source_name)
+		return resourcesRead[i][uuidFieldIndex+2].String() < resourcesRead[j][uuidFieldIndex+2].String()
+	})
 
-	if resource2.Data.Equal(resourcesRead[1]) {
-		return fmt.Errorf("expected data to be %v, got %v", resource.Data, resourcesRead[1])
+	if !expectedResourceData[0].Equal(resourcesRead[0]) {
+		return fmt.Errorf("expected data to be %v, got %v", expectedResourceData[0], resourcesRead[0])
```
hermanschaaf, line 158:
```suggestion
		return fmt.Errorf("expected data[0] to be %v, got %v", expectedResourceData[0], resourcesRead[0])
```

plugins/destination_testing.go
```diff
@@ -30,11 +31,6 @@ type DestinationTestSuiteTests struct {
 	// SkipAppend skips testing for "append" mode. Use if the destination
 	// plugin doesn't support this feature.
 	SkipAppend bool
-
-	// SkipSecondAppend skips the second append step in the test.
-	// This is useful in cases like cloud storage where you can't append to an
-	// existing object after the file has been closed.
-	SkipSecondAppend bool
```
hermanschaaf, line 37: Why is this option no longer necessary?

plugins/destination_testing.go
```diff
@@ -155,53 +206,61 @@ func (s *destinationTestSuite) destinationPluginTestWriteAppend(ctx context.Cont
...
+
+	if !expectedResourceData[0].Equal(resourcesRead[0]) {
+		return fmt.Errorf("expected data to be %v, got %v", expectedResourceData[0], resourcesRead[0])
 	}
-	if !s.tests.SkipSecondAppend {
-		if resource.Data.Equal(resourcesRead[1]) {
-			return fmt.Errorf("expected data to be %v, got %v", resource.Data, resourcesRead[1])
-		}
+	if !expectedResourceData[1].Equal(resourcesRead[1]) {
+		return fmt.Errorf("expected data to be %v, got %v", expectedResourceData[1], resourcesRead[1])
```
hermanschaaf, line 263:
```suggestion
		return fmt.Errorf("expected data[1] to be %v, got %v", expectedResourceData[1], resourcesRead[1])
```

Readers of the test error should be able to differentiate it from the case above somehow

plugins/destination_testing.go
```diff
@@ -46,6 +42,13 @@ func getTestLogger(t *testing.T) zerolog.Logger {
 }
 
 func (s *destinationTestSuite) destinationPluginTestWriteOverwrite(ctx context.Context, p *DestinationPlugin, logger zerolog.Logger, spec specs.Destination) error {
+	// These are the indexes for fields in the test-table (without _cq_sync_time and _cq_source_name)
+	// To get the indices in structs that include those fields, add +2
+	const uuidFieldIndex = 5
+	const intFieldIndex = 3
+	const cqSyncTimeFieldIndex = -1
```
hermanschaaf, line 49: Is there a way we can infer these indexes from the table instead of hardcoding them?

plugins/destination_testing.go
```diff
@@ -88,17 +95,48 @@ func (s *destinationTestSuite) destinationPluginTestWriteOverwrite(ctx context.C
...
+		// We add '+2' to uuidFieldIndex because we add 2 columns (_cq_sync_time and source_name)
+		return expectedResourceData[i][uuidFieldIndex+2].String() < expectedResourceData[j][uuidFieldIndex+2].String()
+	})
+	sort.Slice(resourcesRead, func(i, j int) bool {
+		// We add '+2' to uuidFieldIndex because we add 2 columns (_cq_sync_time and source_name)
+		return resourcesRead[i][uuidFieldIndex+2].String() < resourcesRead[j][uuidFieldIndex+2].String()
+	})
+
+	if !expectedResourceData[0].Equal(resourcesRead[0]) {
+		return fmt.Errorf("expected data to be %v, got %v", expectedResourceData[0], resourcesRead[0])
```
hermanschaaf, line 117:
```suggestion
		return fmt.Errorf("expected data[0] to be %v, got %v", expectedResourceData[0], resourcesRead[0])
```

Ideally here the message should give enough context to tell the reader where it happened. Maybe something like `fmt.Errorf("expected data after first write to be %v, got %v", ...)`. Or another option is to put each section into a `t.Run` function such that they still run in order, but the error message more clearly indicates which part of the test failed.

hermanschaaf (changes requested):
Had a quick read through and left some comments. What was the bug that is being fixed here? Could you add some details about that to the description to help with review?

plugins/destination_testing.go
shimonp21: Was never used - and complicated the code quite a bit...

candiduslynx (approved):

plugins/destination.go
```diff
@@ -88,6 +88,7 @@ func (p *DestinationPlugin) Migrate(ctx context.Context, tables schema.Tables) e
 	return p.client.Migrate(ctx, tables)
 }
 
+// nolint: unparam
```
erezrokah, line 91: Not a blocker but we've discussed this internally and realized the linter was complaining since the private function `readAll` always got the same `sourceName` so technically we could remove that param and inline the value.

If we do that, we should probably move `readAll` to the testing file

PR #465, opened 2022-12-06
fix(sentry): Use HTTPSyncTransport, remove flush

erezrokah:
#### Summary

This PR switches Sentry to work in synchronous mode (like we have in the [CLI](https://github.com/cloudquery/cloudquery/blob/a98e20d3c13b4f9a4ef3a08cf00daa48648935fe/cli/cmd/sentry.go#L10)).

We can't rely on `sentry.Flush` begin called in a consistent way, especially on Windows since it doesn't support sending an interrupt signal and kills the process immediately:
https://github.com/cloudquery/plugin-sdk/blob/464f75f0b586def0c85a501dccc9e6235eebff0e/clients/source_terminate_windows.go#L6

This might have a performance tradeoff, but since we use it for errors/panics only I think that should be ok.
Another solution @disq suggested would be to call `sentry.Flush` after the `sync` for example in source plugins. 

---

- [ ] Update or add tests 🧪

### DIFF
--- a/serve/destination.go
+++ b/serve/destination.go
@@ -49,11 +49,9 @@ func Destination(plugin *plugins.DestinationPlugin, opts ...DestinationOption) {
 	}
 	if err := newCmdDestinationRoot(s).Execute(); err != nil {
 		sentry.CaptureMessage(err.Error())
-		sentry.Flush(flushTimeout)
 		fmt.Println(err)
 		os.Exit(1)
 	}
-	sentry.Flush(flushTimeout)
 }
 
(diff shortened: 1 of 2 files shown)


disq (approved):
Once we verify this is the fix and start getting the panic reports again, we can do another pass to optimize (maybe fall back to `HTTPTransport` without the sync, but do a blocking `Flush()` in critical points etc)

hermanschaaf (approved):
Nice find!

yevgenypats (approved):

PR #467, opened 2022-12-06
fix: Concurrent read,write to a map

yevgenypats:
Im not sure how we didn't hit this before but I was able to hit this while working on azure number of times. Prob will need another set of eye to make sure it's good now.

### DIFF
--- a/plugins/source_scheduler_dfs.go
+++ b/plugins/source_scheduler_dfs.go
@@ -48,11 +48,11 @@ func (p *SourcePlugin) syncDfs(ctx context.Context, spec specs.Source, client sc
 
 	var wg sync.WaitGroup
 	for _, table := range tables {
-		if table.Parent != nil {
-			// skip descendent tables here - they will be handled by the recursive
-			// depth-first-search later.
-			continue
-		}
+		// if table.Parent != nil {
+		// 	// skip descendent tables here - they will be handled by the recursive
+		// 	// depth-first-search later.
+		// 	continue
+		// }
 		table := table
 		clients := []schema.ClientMeta{client}
 		if table.Multiplex != nil {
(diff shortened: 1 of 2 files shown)


plugins/source_scheduler_dfs.go
```diff
@@ -48,7 +58,7 @@ func (p *SourcePlugin) syncDfs(ctx context.Context, spec specs.Source, client sc
 		if table.Multiplex != nil {
 			clients = table.Multiplex(client)
 		}
-		p.metrics.initWithClients(table, clients)
+		// p.metrics.initWithClients(table, clients)
```
erezrokah, line 61: Isn't this already called outside the Go routine? Can you share the stack trace of the concurrency error?

I might need to look at this in the morning with fresh 👀

erezrokah (approved):
I think this is ok but will take another look tomorrow morning.
One comment is that maybe we can re-use the `clients` we generate to avoid calling `Multiplex` again

disq (approved):

candiduslynx (approved):

plugins/source_scheduler_dfs.go
erezrokah, line 61:
```suggestion
```

plugins/source_scheduler_dfs.go
```diff
@@ -36,6 +36,16 @@ func (p *SourcePlugin) syncDfs(ctx context.Context, spec specs.Source, client sc
...
 	p.resourceSem = semaphore.NewWeighted(int64(resourceConcurrency))
 
+	for _, table := range tables {
+		clients := []schema.ClientMeta{client}
+		if table.Multiplex != nil {
+			clients = table.Multiplex(client)
+		}
+		// we do this here to avoid locks so we initial the metrics structure once in the main goroutines
+		// and then we can just read from it in the other goroutines concurrently given we are not writing to it.
+		p.metrics.initWithClients(table, clients)
```
hermanschaaf, line 46: `initWithClients` recursively traverses the tables, so this still has the (minor) issue that some tables will be initialized twice. Maybe let's add `if table.Parent != nil {` to the top of this loop for now so that it's correct, and we can clean up/refactor later?

disq: oh didn't catch that! Yes, definitely only init parent tables as children will be handled by the initWithClients call.

PR #471, opened 2022-12-07
chore: Better channel close

disq:
Otherwise if the `close` panics it won't be caught. But could be unnecessary.

### DIFF
--- a/plugins/source_scheduler_dfs.go
+++ b/plugins/source_scheduler_dfs.go
@@ -103,8 +103,8 @@ func (p *SourcePlugin) resolveTableDfs(ctx context.Context, allIncludedTables sc
 				p.logger.Error().Interface("error", err).Str("stack", stack).Msg("table resolver finished with panic")
 				atomic.AddUint64(&tableMetrics.Panics, 1)
 			}
-			close(res)
 		}()
+		defer close(res)
 		if err := table.Resolver(ctx, client, parent, res); err != nil {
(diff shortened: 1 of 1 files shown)


erezrokah (approved):

yevgenypats:
What effect this PR has? also seem tests are failing?

disq:
> What effect this PR has? also seem tests are failing?

Made it betterer with [`0636bef` (#471)](https://github.com/cloudquery/plugin-sdk/pull/471/commits/0636bef47e982183eceb43947ebabcc1ba1a0273)

Not necessarily a bugfix but makes it easier to read

yevgenypats (changes requested):
Let's avoid for now pushing code that has no effect to the SDK given how hard it is to test things at scale and this code works (also I think I had it this way but in a review I got it was requested to move this to this place so it will just continue to move around) .

plugins/source_scheduler_dfs.go
```diff
@@ -102,14 +102,14 @@ func (p *SourcePlugin) resolveTableDfs(ctx context.Context, allIncludedTables sc
 				})
 				p.logger.Error().Interface("error", err).Str("stack", stack).Msg("table resolver finished with panic")
 				atomic.AddUint64(&tableMetrics.Panics, 1)
+				close(res) // assuming the panic is from the resolver, don't leave the channel open
```
hermanschaaf, line 105: I guess the panic could have also been from closing the channel, in which case we'd be causing another panic here 😅

PR #473, opened 2022-12-07
fix: Initialise clients only once

yevgenypats:
This fix a panic that was caused if plugins misbehave and return difference results for difference runs with the same config.

### DIFF
--- a/plugins/source.go
+++ b/plugins/source.go
@@ -26,7 +26,7 @@ type SourcePlugin struct {
 	// Tables is all tables supported by this source plugin
 	tables schema.Tables
 	// status sync metrics
-	metrics SourceMetrics
+	metrics *SourceMetrics
 	// Logger to call, this logger is passed to the serve.Serve Client, if not defined Serve will create one instead.
 	logger zerolog.Logger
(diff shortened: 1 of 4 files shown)


yevgenypats:
Still testing this

erezrokah (changes requested):
The fix looks good, but I won't kill the process because we can't send metrics. Maybe only log it?

yevgenypats:
I think there were number of issues here and after fixing the initial issue with making sure `Multiplex` is called only once to combat with clients that return random clients on each call I was hit again by the flattening and global list. I'll look into refactoring this somehow next week

plugins/source.go
```diff
@@ -26,7 +26,7 @@ type SourcePlugin struct {
 	// Tables is all tables supported by this source plugin
 	tables schema.Tables
 	// status sync metrics
-	metrics SourceMetrics
+	metrics *SourceMetrics
```
erezrokah, line 29: Is this change to a pointer required for the fix?

yevgenypats: I think yes

erezrokah (approved):

hermanschaaf (approved):

Issue #470, opened 2022-12-07
Refactor code related to flattenened tables


hermanschaaf:
Flattened lists of tables have caused a few bugs now, because once they are returned it's hard to know what kind of list you are working with. Is it a flattened list or a normal list? Calling code that expects one but not the other can cause issues like tables getting traversed twice.

`listAndValidateTables` is one example of a function that returns a flattened list, and has so far been the main culprit https://github.com/cloudquery/plugin-sdk/blob/7474c90415119082bdb1cdb145bd16d1ef51a3b2/plugins/source_validate.go#L33

Flattened lists of tables are useful, but we should clearly distinguish flattened lists of tables from normal top-level lists. My reading is that we currently only use flattened lists for two things:
 - getting the total number of tables
 - checking if a table is in the top-level list by using `GetTopLevel`

These two things can both be achieved by a "set" type. If `listAndValidateTables` returned an abstracted "set" type that hid its internals and exposed only two methods: `Size() int` and `Contains(table) bool`, most of the bugs we've had so far would have been avoided. I think we should try to refactor the code to work something along those lines, so that it is easier to reason about and harder to make mistakes.

hermanschaaf:
I realized now that we have another bug related to this 😮‍💨 

Initially `TableForSpec` returned a flattened list. This led to bugs in some destination plugins that flattened the list a second time. So we updated `TableForSpec` to match `Tables` in returning a top-level list. This fixed the immediate bug, but it was a mistake, because now we have no way of telling whether a child table was specifically selected or not. The consequence is that child relations of selected tables will always be migrated, and will always be cleared of  stale data, even when those child relations are skipped by config.

I think that the original `TablesForSpec` implementation was correct: it _should_ return a flattened list, but we should use a different type (like a set) to distinguish it. Callers should then call `Tables()` for the full list of available tables

---

The core of the issue is this: we have a forest (multiple of trees) structure of tables:

```
A         E
|  \        |
B  C    F
|
D
```

We need to be able to convey information about which of these tables the user intended to select. Let's say the user wants to select `A`, `E` and `F`, but none of `A`'s children. A flattened list would convey it like this:

```
[A, E, F]
```

(with links to child tables still in tact) but the list is hard to distinguish from a list of all top-level tables:

```
[A, E]
```

and right now `TableForSpec` returns `[A, E]`, so information about which descendant tables should be synced is lost.

If we had a set, we could write the user's choice like this:

```
{A, E, F}
```

[...]

PR #475, opened 2022-12-08
feat: Add TableSet type

hermanschaaf:
This implements the set idea from https://github.com/cloudquery/plugin-sdk/issues/470

It doesn't change any behavior, and doesn't fix the bug in `TableForSpec`. It's just for readability: `listAndValidateTables` now returns a set type which only supports a very limited set of operations. This set type is passed to `syncDfs` rather than a flattened list. `syncDfs` and its related functions must reference `p.tables` to traverse the tables, but can now check `selectedTables.Contains(table.Name)` to see whether a given table should be included or skipped. I think this makes the code quite a bit simpler (e.g. no more checks for whether a table is top-level or not) and less error-prone.

### DIFF
--- a/plugins/source.go
+++ b/plugins/source.go
@@ -96,7 +96,6 @@ func NewSourcePlugin(name string, version string, tables []*schema.Table, newExe
 	if err := p.validate(); err != nil {
 		panic(err)
 	}
-	p.metrics.initWithTables(p.tables)
 	p.maxDepth = maxDepth(p.tables)
 	if p.maxDepth > maxAllowedDepth {
 		panic(fmt.Errorf("max depth of tables is %d, max allowed is %d", p.maxDepth, maxAllowedDepth))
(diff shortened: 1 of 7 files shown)


plugins/source.go
```diff
@@ -119,18 +119,17 @@ func (p *SourcePlugin) TablesForSpec(spec specs.Source) (schema.Tables, error) {
...
 	}
-	tables, err := p.listAndValidateTables(spec.Tables, spec.SkipTables)
+	selectedTables, err := p.listAndValidateTables(spec.Tables, spec.SkipTables)
 	if err != nil {
 		return nil, err
 	}
-	// listAndValidateTables returns a flattened list - we only want to return
-	// the top-level tables from this function.
+	// listAndValidateTables returns a set of all tables - we only want to return
+	// the top-level tables from the set in this function.
```
hermanschaaf, line 127: Technically I think this behavior should change, but I'm not changing it in this PR.

schema/table.go
```diff
@@ -65,6 +65,33 @@ var (
 	reValidColumnName = regexp.MustCompile(`^[a-z_][a-z\d_]*$`)
 )
 
+type TableSet struct {
+	tables map[string]bool
```
disq, line 69: why bool, use `struct{}`? bool looks simpler but it's confusing if used for a set type.

yevgenypats:
Closing in favour of https://github.com/cloudquery/plugin-sdk/pull/488

Issue #462, opened 2022-12-05
feat: Return application level errors as a part of the gRPC response


erezrokah:
TLDR: We should consider having application level errors as a part of the gRPC responses, to make it easier to identify them.

For example in the following line:
https://github.com/cloudquery/plugin-sdk/blob/ab7ca972e0b187a7dfb66132a03f07479cd29bb7/clients/destination.go#L303
`failed to CloseAndRecv` is not necessarily correct, as `err` is whatever is returned from the following function:
https://github.com/cloudquery/plugin-sdk/blob/ab7ca972e0b187a7dfb66132a03f07479cd29bb7/internal/servers/destinations.go#L65.

Not sure what's the best solution (need to do some research), but we could add an error(s) field to gRPC responses so we have a clear separation between application level errors and gRPC/communication level ones.

Also due to https://github.com/grpc/grpc-go/issues/2934 we can't unwrap errors.

erezrokah:
Looks like we could use `WithDetails` for this https://pkg.go.dev/google.golang.org/grpc/internal/status#Status.WithDetails

yevgenypats:
I understand the need for application level errors but I don't think it should have anything with grpc errors. depending on which errors we want to send we need to add it to the protobuf. grpc errors are mostly used when something bad happened on the server.

erezrokah:
> I understand the need for application level errors

Most errors returned from https://github.com/cloudquery/plugin-sdk/blob/62692b9cb8a3ff3465d9d14a1ec7cc801d3490af/internal/servers/destinations.go#L65 are application level.

The only ones that could be related to gRPC protocol are:
https://github.com/cloudquery/plugin-sdk/blob/62692b9cb8a3ff3465d9d14a1ec7cc801d3490af/internal/servers/destinations.go#L73
https://github.com/cloudquery/plugin-sdk/blob/62692b9cb8a3ff3465d9d14a1ec7cc801d3490af/internal/servers/destinations.go#L101, but also not necessarily depending on what error caused the stream to close

yevgenypats:
There can be also grpc errror returned after this function is returned that we actually saw when buffer is too big so there is no good way to distinguish both over the grpc error field.

erezrokah:
> no good way to distinguish both over the grpc error field

Agree there could be both application errors and gRPC errors. Right now we're mixing both so the purpose of this issue is to discuss a solution so we can distinguish between them.

[...]

PR #477, opened 2022-12-08
fix(sync): Extract error message from gRPC error

erezrokah:
#### Summary

Related to https://github.com/cloudquery/cloudquery/issues/4707.
Errors returned from a gRPC call have both a status and a message. We don't really use the status, but when we wrap the error it always gets printed, see:
https://github.com/grpc/grpc-go/blob/eeb9afa1f6b6388152955eeca8926e36ca94c768/internal/status/status.go#L140

We can easily extract just the message and print only that.
Before:
```
Error: failed to sync source aws: failed to fetch resources from stream: rpc error: code = Unknown desc = failed to sync resources: failed to create execution client for source plugin aws: no enabled accounts instantiated
```
After:
```
Error: failed to sync source aws: failed to sync resources: failed to create execution client for source plugin aws: no enabled accounts instantiated
```

We should do this for all gRPC calls, but I'll do that in follow up PRs so I can test it

---

- [ ] Update or add tests 🧪

### DIFF
--- a/clients/source.go
+++ b/clients/source.go
@@ -277,7 +277,7 @@ func (c *SourceClient) Sync(ctx context.Context, spec specs.Source, res chan<- [
 		Spec: b,
 	})
 	if err != nil {
-		return fmt.Errorf("failed to call Sync: %w", err)
+		return fmt.Errorf("failed to call Sync: %s", extractErrorMessage(err))
 	}
 	for {
 		r, err := stream.Recv()
@@ -285,7 +285,7 @@ func (c *SourceClient) Sync(ctx context.Context, spec specs.Source, res chan<- [
 			if err == io.EOF {
 				return nil
 			}
-			return fmt.Errorf("failed to fetch resources from stream: %w", err)
+			return errors.New(extractErrorMessage(err))
 		}
 		select {
 		case <-ctx.Done():
@@ -307,7 +307,7 @@ func (c *SourceClient) Sync2(ctx context.Context, spec specs.Source, res chan<- 
 		Spec: b,
 	})
 	if err != nil {
-		return fmt.Errorf("failed to call Sync: %w", err)
+		return fmt.Errorf("failed to call Sync: %s", extractErrorMessage(err))
 	}
 	for {
 		r, err := stream.Recv()
@@ -315,7 +315,7 @@ func (c *SourceClient) Sync2(ctx context.Context, spec specs.Source, res chan<- 
 			if err == io.EOF {
 				return nil
 			}
-			return fmt.Errorf("failed to fetch resources from stream: %w", err)
+			return errors.New(extractErrorMessage(err))
 		}
 		select {
 		case <-ctx.Done():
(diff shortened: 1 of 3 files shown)


clients/source.go
```diff
@@ -307,15 +307,15 @@ func (c *SourceClient) Sync2(ctx context.Context, spec specs.Source, res chan<-
...
-		return fmt.Errorf("failed to call Sync: %w", err)
+		return fmt.Errorf("failed to call Sync: %s", extractErrorMessage(err))
 	}
 	for {
 		r, err := stream.Recv()
 		if err != nil {
 			if err == io.EOF {
 				return nil
 			}
-			return fmt.Errorf("failed to fetch resources from stream: %w", err)
```
erezrokah, line 318: `failed to fetch resources from stream` is incorrect (similar to https://github.com/cloudquery/plugin-sdk/issues/462#issue-1476833829), as error can be anything returned from https://github.com/cloudquery/plugin-sdk/blob/cdcef5d9256ee7d43336e557d1c4db3eda2ee46b/internal/servers/source.go#L93 so not necessarily means something with the stream

disq (approved):

yevgenypats (changes requested):
What is the value of extracting it rather than printing it as is (without more code) with the `grpc=` given this error comes from grpc so basically we are hiding more information just because the string is a bit long?

erezrokah:
> given this error comes from grpc so basically we are hiding more information just because the string is a bit long?

This is not a gRPC error. This is an application level error and we almost always print `rpc error: code = Unknown desc =` which has no value. See also bullet 2. in https://github.com/cloudquery/plugin-sdk/pull/485#issue-1489972263 and issue https://github.com/cloudquery/plugin-sdk/issues/462

yevgenypats:
> > given this error comes from grpc so basically we are hiding more information just because the string is a bit long?
> 
> This is not a gRPC error. This is an application level error and we almost always print `rpc error: code = Unknown desc =` which has no value. See also bullet 2. in [#485 (comment)](https://github.com/cloudquery/plugin-sdk/pull/485#issue-1489972263) and issue #462.
> 
> Also explained in [#477 (comment)](https://github.com/cloudquery/plugin-sdk/pull/477#discussion_r1043569021)

Right now we don't have a way to send application level errors. We can implement this type of behaviour and add a field to the protocol but it's not the right place to do it over grpc errors.

erezrokah:
> We can implement this type of behaviour and add a field to the protocol but it's not the right place to do it over grpc errors.

Agree we should not do it as a part of gRPC errors, hence https://github.com/cloudquery/plugin-sdk/pull/487. We don't need to add anything to the protocol as `errdetails` already has all the things generated.

This PR doesn't change error handling, its meant to only remove the status print `rpc error: code = Unknown desc` as we're not using it anywhere nor it's useful to the user, pending a better solution.

erezrokah:
Closing since I think the fixes in https://github.com/cloudquery/plugin-sdk/pull/530 should be enough for now. We can re-visit this in the future

PR #479, opened 2022-12-08
fix: JSON empty string

bbernays:
#### Summary

---

- [ ] Update or add tests 🧪


schema/json.go
```diff
@@ -51,12 +51,20 @@ func (dst *JSON) Set(src interface{}) error {
 
 	switch value := src.(type) {
 	case string:
-		*dst = JSON{Bytes: []byte(value), Status: Present}
+		if value == "" {
```
disq, line 54: Can you test with `*string` and `string` cases removed from the `switch`? Just let it fall to `default`?

bbernays: If we do that, then all of those strings will just be turned into `{}` rather than `""`... Depends on the behavior we want...

schema/json.go
```diff
@@ -51,12 +51,20 @@ func (dst *JSON) Set(src interface{}) error {
...
+			*dst = JSON{Bytes: []byte(`""`), Status: Present}
```
shimonp21, line 55: I believe `JSON{Status: Null}` would be more correct here.
Current implementation results in a `""` value in the postgres column (which is kinda weird - because likely the policy is an object of some-sort like `{"some-policy": "some-value"}`).

Just null i think - no reason to distinguish between empty-string and nonexistent.

bbernays: I was trying to keep the same behavior that go has when we call `Marshal`...

`json.Marshal("")` produces `""`

schema/json.go
```diff
@@ -51,12 +51,20 @@ func (dst *JSON) Set(src interface{}) error {
...
+		} else {
+			*dst = JSON{Bytes: []byte(value), Status: Present}
```
candiduslynx, line 57: does it do anything?

schema/json.go
```diff
@@ -51,12 +51,20 @@ func (dst *JSON) Set(src interface{}) error {
...
+		}
 	case *string:
 		if value == nil {
 			*dst = JSON{Status: Null}
 		} else {
-			*dst = JSON{Bytes: []byte(*value), Status: Present}
+			if *value == "" {
+				*dst = JSON{Bytes: []byte(`""`), Status: Present}
```
candiduslynx, line 64: does it change anything?

candiduslynx (changes requested):

yevgenypats:
Closing in favour of https://github.com/cloudquery/plugin-sdk/pull/497

PR #486, opened 2022-12-11
fix(destinations): Set correct status code for context errors

erezrokah:
#### Summary

This ensures relevant status code (`codes.DeadlineExceeded` or `codes.Canceled`) is set for the gRPC error.
Without this fix the status will always be `codes.Unknown`.

---

- [ ] Update or add tests 🧪

### DIFF
--- a/internal/servers/destinations.go
+++ b/internal/servers/destinations.go
@@ -115,11 +115,7 @@ func (s *DestinationServer) Write2(msg pb.Destination_Write2Server) error {
 			if err := eg.Wait(); err != nil {
 				s.Logger.Error().Err(err).Msg("failed to wait")
 			}
-			st := status.FromContextError(ctx.Err())
-			if st == nil {
-				return nil
-			}
-			return st.Err()
+			return status.FromContextError(ctx.Err()).Err()
 		}
 	}
 }
--- a/internal/servers/destinations.go
+++ b/internal/servers/destinations.go
@@ -115,7 +115,11 @@ func (s *DestinationServer) Write2(msg pb.Destination_Write2Server) error {
 			if err := eg.Wait(); err != nil {
 				s.Logger.Error().Err(err).Msg("failed to wait")
 			}
-			return ctx.Err()
+			st := status.FromContextError(ctx.Err())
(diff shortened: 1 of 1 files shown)


internal/servers/destinations.go
```diff
@@ -115,7 +115,11 @@ func (s *DestinationServer) Write2(msg pb.Destination_Write2Server) error {
 			if err := eg.Wait(); err != nil {
 				s.Logger.Error().Err(err).Msg("failed to wait")
 			}
-			return ctx.Err()
+			st := status.FromContextError(ctx.Err())
+			if st == nil {
+				return nil
+			}
+			return st.Err()
```
candiduslynx, line 122:
```suggestion
			return status.FromContextError(ctx.Err()).Err()
```
https://github.com/grpc/grpc-go/blob/master/internal/status/status.go#L96-L101

candiduslynx (approved):

internal/servers/destinations.go
erezrokah: Is it guaranteed that that `ctx.Err()` is always not `nil`? Otherwise `st` can be `nil` too

candiduslynx: See this additional piece: https://github.com/grpc/grpc-go/blob/master/internal/status/status.go#L72-L77
Basically, if we pass a nil err (works either for `status.FromError` and for `status.FromContextError` we'll get `(*status.Status)(nil)`. However, as the receiver for `Code` & `Err` methods may be nil and the code accounts for that, you don't beed to do branching here – you'll just get the nil err on the output and it's what you're doing via `if` statement anyways.

erezrokah: OK so I think what I was missing is https://groups.google.com/g/golang-nuts/c/wcrZ3P1zeAk/m/WI88iQgFMvwJ. Thanks for explaining

candiduslynx (approved):

erezrokah:
Closing since I think the fixes in https://github.com/cloudquery/plugin-sdk/pull/530 should be enough for now. We can re-visit this in the future

PR #487, opened 2022-12-11
feat: Report application errors with details

erezrokah:
#### Summary

Related to https://github.com/cloudquery/plugin-sdk/issues/462.
Draft PR as meant to serve as an example on how to improve error reporting and handling.
I'm using `errdetails` but we could also write our own protobuf messages instead (don't think we should).

---

- [ ] Update or add tests 🧪

### DIFF
--- a/clients/destination.go
+++ b/clients/destination.go
@@ -19,6 +19,7 @@ import (
 	"github.com/cloudquery/plugin-sdk/schema"
 	"github.com/cloudquery/plugin-sdk/specs"
 	"github.com/rs/zerolog"
+	"google.golang.org/genproto/googleapis/rpc/errdetails"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials/insecure"
@@ -300,7 +301,20 @@ func (c *DestinationClient) Write2(ctx context.Context, tables schema.Tables, so
 	}
 	_, err = saveClient.CloseAndRecv()
 	if err != nil {
-		return fmt.Errorf("failed to CloseAndRecv client: %w", err)
+		st := status.Convert(err)
+		errorWithDetails := st.Message()
+		if len(st.Details()) > 0 {
+			errorWithDetails += "\nDetails:\n"
+			for _, d := range st.Details() {
+				switch t := d.(type) {
+				case *errdetails.ErrorInfo:
+					errorWithDetails += t.Reason + "\n"
+				default:
+					errorWithDetails += fmt.Sprintf("%v\n", d)
+				}
+			}
+		}
+		return errors.New(errorWithDetails)
 	}
 
 	return nil
--- a/internal/servers/destinations.go
+++ b/internal/servers/destinations.go
@@ -3,6 +3,7 @@ package servers
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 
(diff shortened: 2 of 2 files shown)


clients/destination.go
```diff
@@ -300,7 +301,20 @@ func (c *DestinationClient) Write2(ctx context.Context, tables schema.Tables, so
 	}
 	_, err = saveClient.CloseAndRecv()
 	if err != nil {
-		return fmt.Errorf("failed to CloseAndRecv client: %w", err)
+		st := status.Convert(err)
+		errorWithDetails := err.Error()
```
erezrokah, line 305: I'm concatenating errors into one string, but we can decide to do something else

internal/servers/destinations.go
```diff
@@ -86,36 +94,42 @@ func (s *DestinationServer) Write2(msg pb.Destination_Write2Server) error {
...
 		}
 		var resource schema.DestinationResource
-		if err := json.Unmarshal(r.Resource, &resource); err != nil {
+		if unmarshalError := json.Unmarshal(r.Resource, &resource); unmarshalError != nil {
 			close(resources)
-			if err := eg.Wait(); err != nil {
-				s.Logger.Error().Err(err).Msg("failed to unmarshal resource. failed to wait for plugin")
+			errWithDetails := addErrorDetails(errors.New("failed to unmarshal resource"), unmarshalError.Error())
+			if writeError := eg.Wait(); writeError != nil {
+				errWithDetails = addErrorDetails(errors.New("plugin write failed and failed to unmarshal resource"), writeError.Error())
```
erezrokah, line 116: `Details` is an array so we can send multiple errors

internal/servers/destinations.go
```diff
@@ -86,36 +94,42 @@ func (s *DestinationServer) Write2(msg pb.Destination_Write2Server) error {
...
 			if err := eg.Wait(); err != nil {
-				s.Logger.Error().Err(err).Msg("failed to wait")
+				return addErrorDetails(errors.New("plugin write failed"), err.Error())
+			}
+			ctxError := ctx.Err()
+			if ctxError == nil {
+				return nil
 			}
-			return ctx.Err()
+			return addErrorDetails(errors.New("context error"), ctxError.Error())
```
erezrokah, line 132: This is only an example, we should be using https://github.com/cloudquery/plugin-sdk/pull/486 instead

clients/destination.go
```diff
@@ -300,7 +301,20 @@ func (c *DestinationClient) Write2(ctx context.Context, tables schema.Tables, so
 	}
 	_, err = saveClient.CloseAndRecv()
 	if err != nil {
-		return fmt.Errorf("failed to CloseAndRecv client: %w", err)
+		st := status.Convert(err)
```
erezrokah, line 304: We should probably put this into a function and replace https://github.com/cloudquery/plugin-sdk/pull/477 so we can get both the message and the details

clients/destination.go
```diff
@@ -300,7 +301,20 @@ func (c *DestinationClient) Write2(ctx context.Context, tables schema.Tables, so
...
+		errorWithDetails := st.Message()
+		if len(st.Details()) > 0 {
+			errorWithDetails += "\nDetails:\n"
```
candiduslynx, line 307: can we just utilize [prototext](https://pkg.go.dev/google.golang.org/protobuf/encoding/prototext)?

erezrokah: We can try good catch. Please note this is only one example on how to report the errors. We might want to do something else than strings concatenation

erezrokah:
Closing this POC for now

PR #488, opened 2022-12-11
chore: Refactor glob filters

yevgenypats:
This simplifies our glob filtering and makes sure Tables are always stored in a tree like structured apart from when returned from `FlattenTables`. This solves a lot of potential bugs and places where we need to do things like `if parent != nil`.

I think this also closes https://github.com/cloudquery/plugin-sdk/pull/475 as there is no need for another data structure.

The reasoning for this is that our tables are tree like structured and it's easier imo to always keep it like this everywhere across the code and not have multiple data structures with similar methods and so on. This required a small update to our filtering logic which just works as a DFS and in-place filtering (for that to work I had to add a copy method which I think can be  useful for other things in the future).


plugins/source_test.go
```diff
@@ -322,59 +297,6 @@ func testSyncTable(t *testing.T, tc syncTestCase) {
 	}
 }
 
-func TestTablesForSpec(t *testing.T) {
```
hermanschaaf, line 325: Can we bring back this test in some form? Otherwise there is no test for `TablesForSpec`

hermanschaaf (approved):
Looks good, I think this should do the trick. :+1:

I would just like to make sure we don't leave a gap in our validation and testing:
 - we should have at least some basic test for `TablesForSpec`
 - we should bring back some of the validation that `listAndValidateTables` was doing, and the tests for it. It's fine if we put that in `Validate`, but it was checking for things like invalid table names (or globs that match nothing), which I think is a pretty useful user-facing feature

schema/table_test.go
```diff
@@ -19,3 +23,154 @@ func TestTablesFlatten(t *testing.T) {
...
+		},
+		{
+			name:                    "should return all tables when '*' is provided, excluding skipped tables",
+			tables:                  []*Table{{Name: "table1"}, {Name: "table2"}, {Name: "table3"}},
+			configurationTables:     []string{"*"},
+			configurationSkipTables: []string{"table1", "table3"},
+			want:                    []string{"table2"},
+		},
+		{
+			name: "should return the parent and all its descendents",
```
erezrokah, line 49:
```suggestion
			name: "should return the parent and all its descendants",
```

schema/table_test.go
```diff
@@ -19,3 +23,154 @@ func TestTablesFlatten(t *testing.T) {
...
+						},
+					},
+				},
+			},
+			configurationTables:     []string{"sub_sub_table"},
+			configurationSkipTables: []string{},
+			want:                    []string{"main_table", "sub_table", "sub_sub_table"},
+		},
+		{
+			name: "should return the parent and all its descendents, but skip children and their descendents if they are skipped",
```
erezrokah, line 70:
```suggestion
			name: "should return the parent and all its descendants, but skip children and their descendants if they are skipped",
```

schema/table.go
```diff
@@ -167,6 +179,36 @@ func (tt Tables) ValidateColumnNames() error {
...
+		if glob.Glob(skipTable, t.Name) {
+			matched = false
+			break
+		}
+	}
+	filteredRelations := make([]*Table, 0, len(t.Relations))
+	for _, r := range t.Relations {
+		filteredChild := r.filterDfs(tables, skipTables)
+		if filteredChild != nil {
+			matched = true
```
erezrokah, line 201: Won't this always return the parent table if the child matches, for example:
```go
               {
			name:                    "should skip parent table",
			tables:                  []*Table{{Name: "main_table", Relations: []*Table{{Name: "sub_table", Parent: &Table{Name: "main_table"}}}}},
			configurationTables:     []string{"*"},
			configurationSkipTables: []string{"main_table"},
			want:                    []string{""},
		},
```

The above test fails as it returns `main_table` and `sub_table`

erezrokah (changes requested):
I like this approach, but I think we're missing a use case

erezrokah:
Also @yevgenypats may I suggest we do this as a `fix` so it triggers a release, and also release it separately?
I mean:
1. Release https://github.com/cloudquery/plugin-sdk/pull/482
2. Release all plugins with https://github.com/cloudquery/plugin-sdk/pull/482
3. Then release this PR

We've had a bunch of bugs related to this logic, so maybe safer to do a dedicated release

schema/table.go
yevgenypats: Good catch. it depends on what we define on what has preference. if skip has preference then yes. I've added that logic.

erezrokah: Ah I think https://github.com/cloudquery/plugin-sdk/pull/488/files#diff-7e84c216b71c17b23bc66ee33042fe73fc9352c152dbc4fb4cd9a94d74e2bfb4R49 is the reason

schema/table.go
```diff
@@ -167,6 +179,35 @@ func (tt Tables) ValidateColumnNames() error {
...
+	var matched bool
+	for _, includeTable := range tables {
+		if glob.Glob(includeTable, t.Name) {
+			matched = true
+			break
+		}
+	}
+	for _, skipTable := range skipTables {
+		if glob.Glob(skipTable, t.Name) {
+			return nil
```
erezrokah, line 193: With this I think we can do the skip `for` loop first and the `tables` for later

schema/table.go
```diff
@@ -167,6 +179,35 @@ func (tt Tables) ValidateColumnNames() error {
...
+		}
+	}
```
erezrokah, line 195:
```suggestion
	for _, skipTable := range skipTables {
		if glob.Glob(skipTable, t.Name) {
			return nil
		}
	}
	
	var matched bool
	for _, includeTable := range tables {
		if glob.Glob(includeTable, t.Name) {
			matched = true
			break
		}
	}
```

erezrokah (approved):
I think this looks good now, added another blocking comment as I think we can exit early on skip

plugins/source_test.go
yevgenypats: I think this is already tested under `serve` so not sure there is need to test it twice. most tests should go under `serve` given our plugin is called only via `grpc`

hermanschaaf: Ah okay, yeah makes sense

Issue #454, opened 2022-12-01
Fix additional newlines in markdown files


hermanschaaf:
When generating markdown docs, we often get extra newlines that are hard to fix with Go templates themselves (while still handling all possible cases). The semantics of `{{-` and `-}}` seem too limited for this purpose. 

I'm open to anyone trying to fix all the cases of unnecessary newlines using templating, but another idea is to rather run a post-processor (maybe a markdown formatter?) on the files to make them uniform. It might be as simple as replacing 3+ occurrences of `\n`with `\n\n`.

PR #492, opened 2022-12-13
fix: Improve formatting of newlines in markdown files

hermanschaaf:
This neatens the newlines rendered for table docs, so we can no longer get multiple newlines in a row, like here: https://github.com/cloudquery/cloudquery/pull/5547/files#diff-7a51ae50d93ba1ae94ee3a2e3ff898ae0224a0a1cc32cafe0c667ff46cf080e6R1-R8

Closes https://github.com/cloudquery/plugin-sdk/issues/454

### DIFF
--- a/plugins/source/.snapshots/TestGeneratePluginDocs-Markdown-README.md
+++ b/plugins/source/.snapshots/TestGeneratePluginDocs-Markdown-README.md
@@ -1,5 +1,7 @@
 # Source Plugin: test
+
 ## Tables
+
 - [test_table](test_table.md)
   - [relation_table](relation_table.md)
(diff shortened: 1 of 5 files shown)


erezrokah (approved):

erezrokah:
Looks great, but if we need more formatting maybe use https://github.com/shurcooL/markdownfmt or https://github.com/DavidAnson/markdownlint with `--fix` and one of the CLIs it provides

hermanschaaf:
@erezrokah Yup, I looked at those, but it would be a fairly big dependency and I think this does what we need for now.

yevgenypats (approved):

PR #494, opened 2022-12-13
fix: Glob table filtering

yevgenypats:
If parent is selected all child selected as well unless explicitly skipped.

### DIFF
--- a/schema/table_test.go
+++ b/schema/table_test.go
@@ -63,7 +63,7 @@ func TestTablesFilterDFS(t *testing.T) {
 					},
 				},
 			},
-			configurationTables:     []string{"sub_sub_table"},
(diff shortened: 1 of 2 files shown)


hermanschaaf (approved):

schema/table_test.go
```diff
@@ -183,6 +183,16 @@ func TestTablesFilterDFS(t *testing.T) {
 			want:                    []string{},
 			err:                     "tables include a pattern main_table1 with no matches",
 		},
+		{
+			name:                    "skip child table but return sibilings",
```
hermanschaaf, line 187:
```suggestion
			name:                    "skip child table but return siblings",
```

erezrokah (approved):

PR #496, opened 2022-12-13
feat: Add basic periodic metric INFO logger

shimonp21:
- Logs very basic stats about the sync every 30 seconds.
- Used `ctx` and a new waitgroup - because it's importat to shut it down gracefully (concurrent map read-writes are dangerous in golang).
- Used `TotalResourcesAtomic` - see this stackoverflow answer: https://stackoverflow.com/a/46557083


plugins/source_scheduler_dfs.go
```diff
@@ -71,7 +73,21 @@ func (p *SourcePlugin) syncDfs(ctx context.Context, spec specs.Source, client sc
 			}()
 		}
 	}
+
+	// We start a goroutine that logs the metrics periodically.
+	// It needs its own waitgroup
+	var logWg sync.WaitGroup
+	logWg.Add(1)
```
yevgenypats, line 80: I don't think you need the waitgroup here

plugins/source_scheduler_dfs.go
```diff
@@ -71,7 +73,21 @@ func (p *SourcePlugin) syncDfs(ctx context.Context, spec specs.Source, client sc
...
+
+	logCtx, logCancel := context.WithCancel(ctx)
+	go p.periodicMetricLogger(logCtx, &logWg)
```
yevgenypats, line 83: you want to start this before you start spawning all the clients as it might be blocked if there are many tables/clients.
After you run `go p.periodicMetricLogger` you should just call `defer logCancel()`

yevgenypats (changes requested):
Few changes needed to make it work

plugins/source_scheduler_dfs.go
shimonp21: 1) Here is the race I am worried about:
- A sync starts and finishes, and `syncDfs` returns. However, periodic-metric-logger is still running (maybe he's still summing up in TotalResourcesAtomic, so not looking at `ctx` right now.
- A new sync starts (maybe we are in `grpc-mode`?), and `metrics.InitWithClients` is called.
- `periodicMetricLogger` is still running and summing up stuff from the map, so golang panics with `concurrent reads and writes to map`!

See the following code-snippet for a toy reproduction: 
https://gist.github.com/shimonp21/dda9ee70434ad5da25f3c3ecf5149b40

This produces a panic:
```
fatal error: concurrent map read and map write
fatal error: concurrent map read and map write

goroutine 31 [running]:
main.mapReader(0x0?)
        /Users/shimonp/projects/etc/main.go:25 +0x4e
created by main.main
        /Users/shimonp/projects/etc/main.go:12 +0x6e

goroutine 1 [runnable]:
main.main()
        /Users/shimonp/projects/etc/main.go:12 +0x8f

goroutine 17 [runnable]:
main.mapWriter(0xc0000547b8?)
        /Users/shimonp/projects/etc/main.go:18 +0x6c
created by main.main
        /Users/shimonp/projects/etc/main.go:8 +0x65
```

2. because logCancel is a child-context of `ctx` - it will be cancelled if the parent is cancelled. I don't want to put it in a `defer`, because, as mentioned in point 1 - I think we need to synchronously wait for the priodiclogger to finish.


Agree is a far-fetched race, but AFAIU it can definitly happen, so I'd rather err on the side of "no races in the code".

yevgenypats (approved):
Not sure I understood the write-up on the race condition but this looks fine to me now once it moved to the right place.

PR #508, opened 2022-12-16
fix(codegen): Preserve PK column order with `codegen.WithPKColumns` option

candiduslynx:
#### Summary

If we have a struct
```go
type A struct {
	A int
	B int
}
```
and use `codegen.WithPKColumns("b", "a")`, the resulting PK should be `(b, a)`.
However, prior to this fix it'd be `(a, b)`.

Additionally, if we have a struct
```go
type A struct {
	A int
	B int
	C int
}
```
and use the following options together:
* `codegen.WithPKColumns("c", "a")`
* `codegen.SkipFields([]string{"B"}`
* `codegen.WithExtraColumns(ColumnDefinitions{{Name:"b", …, Options: PK}})`
the resulting PK should be `(c, a, b)`.
However, prior to this fix it'd be `(a, b, c)`.
---

- [x] Update or add tests 🧪

### DIFF
--- a/codegen/column.go
+++ b/codegen/column.go
@@ -5,6 +5,7 @@ import (
 	"reflect"
 
 	"github.com/cloudquery/plugin-sdk/schema"
+	"golang.org/x/exp/slices"
 )
 
 type (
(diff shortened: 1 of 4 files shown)


erezrokah (commented):
Thanks for the PR @candiduslynx, I understand the change, but can you give an example where this is needed? Is there a reason not to keep the PK order the same as the struct properties order?

Also, can we do the `go.mod` and `go.sum` changes in a different PR? They seem unrelated and also we have a bot that those so no need to do them manually

candiduslynx:
@erezrokah 
1. This will be particularly beneficial in the relation resource case, where parent ID is part of the PK. This way we can do extra column + specify the correct order in the PK columns option
2. Yes, sure

erezrokah:
> This way we can do extra column + specify the correct order in the PK columns option

Can you show me some code please 🙌 
As long as the order is predicable, is there a reason to change it? Specially what makes the order correct?

candiduslynx:
When we have PK consisting of sevral columns, the column order is affecting the partitioning. I'm mostly concerned about hard cases like this: https://github.com/cloudquery/cloudquery/pull/5222/files#diff-9bebd76719512dca05a2eb0bf4be316ad6cddc72424731348c567847127ee050R22

erezrokah:
> When we have PK consisting of several columns, the column order is affecting the partitioning

Thank for the context, I got it. But since the order is predictable at the moment the partitioning still works.
I'm mostly concerned on the impact on existing code gen that uses this option, as this means we'll need to update existing code gen recipes to keep the current partitioning the same. Not sure how many places we'll need to update though

yevgenypats:
Codegen is deprecated. Closing.

PR #510, opened 2022-12-17
feat: Add size in bytes to CQ types

yevgenypats:
This makes sure every CQType implements size so we can report to the users also how many bytes/MB were synced in addition to number of resources/rows.

This will also be useful in our analytics.


schema/bool.go
```diff
@@ -19,6 +19,10 @@ func (*Bool) Type() ValueType {
 	return TypeBool
 }
 
+func (*Bool) Size() int {
+	return 2
```
candiduslynx, line 23: 1?

schema/cidr.go
```diff
@@ -12,6 +12,10 @@ func (*CIDR) Type() ValueType {
 	return TypeCIDR
 }
 
+func (*CIDR) Size() int {
```
candiduslynx, line 15: can't you just make CIDR an alias to Inet?
```go
type CIDR = Inet
```

schema/inet.go
```diff
@@ -29,6 +29,10 @@ func (*Inet) Type() ValueType {
 	return TypeInet
 }
 
+func (dst *Inet) Size() int {
+	return 32
```
candiduslynx, line 33:
```suggestion
	return len(dst.IPNet.IP) + len(dst.IPNet.Mask)
```

schema/inet_array.go
```diff
@@ -22,6 +22,10 @@ func (*InetArray) Type() ValueType {
 	return TypeInetArray
 }
 
+func (dst *InetArray) Size() int {
+	return len(dst.Elements) * 32
```
candiduslynx, line 26: has to be counted by each element len **or** need to account for IPv6

candiduslynx (changes requested):

erezrokah (commented):
Could we use https://github.com/DmitriyVTitov/size for this?
It seems calculating the size in bytes of values has a lot of gotchas, so maybe better to use a library that already has a few tests for it?

If not using a library, maybe use `reflect.Size()` https://go.dev/play/p/Sogem09ks3o

yevgenypats:
> Could we use https://github.com/DmitriyVTitov/size for this? It seems calculating the size in bytes of values has a lot of gotchas, so maybe better to use a library that already has a few tests for it?
> 
> If not using a library, maybe use `reflect.Size()` https://go.dev/play/p/Sogem09ks3o

We don't want to use https://github.com/DmitriyVTitov/size/blob/master/size.go as this uses reflection. There is no much gotchas for size calculation but also we mostly use it for a ballpark number it doesn't has to be the exact number of bytes. To have the exact number of bytes we need to `marshal` it and do `len`. If we will ever need it to be byte accurate then we can change to `marshal` and `len` approach but it will introduce overhead

schema/uuid_array.go
```diff
@@ -21,6 +21,10 @@ func (*UUIDArray) Type() ValueType {
 	return TypeUUIDArray
 }
 
+func (dst *UUIDArray) Size() int {
+	return 16 * len(dst.Elements)
```
erezrokah, line 25: Maybe call `UUID.Size()` here?

schema/timestamptz.go
```diff
@@ -35,6 +35,10 @@ func (*Timestamptz) Type() ValueType {
 	return TypeTimestamp
 }
 
+func (dst *Timestamptz) Size() int {
+	return 32
```
erezrokah, line 39: How is this calculated? Because `time.Time` seems to be 24 bytes:
https://go.dev/play/p/K7awGEnpngi.

schema/int8_array.go
```diff
@@ -21,6 +21,10 @@ func (*Int8Array) Type() ValueType {
 	return TypeIntArray
 }
 
+func (dst *Int8Array) Size() int {
+	return 8 * len(dst.Elements)
```
erezrokah, line 25: Can we call `Int8.Size()` here?

schema/cidr_array.go
```diff
@@ -22,6 +22,10 @@ func (*CIDRArray) Type() ValueType {
 	return TypeCIDRArray
 }
 
+func (dst *CIDRArray) Size() int {
+	return len(dst.Elements) * 32
```
erezrokah, line 26:
```suggestion
	return len(dst.Elements) * 32
```

Similar to under comments, maybe we can use the underlying type value instead of `32`, i.e. `CIDR.Size()`

erezrokah:
> To have the exact number of bytes we need to `marshal` it and do `len`. If we will ever need it to be byte accurate then we can change to `marshal` and `len` approach but it will introduce overhead

`Marshal` also uses reflection https://cs.opensource.google/go/go/+/refs/tags/go1.19.4:src/encoding/json/encode.go;l=330;drc=0b2ad1d815ea8967c49b32d848b2992d0c588d88.

I think the question is what we want, size in memory, or serialized size.

schema/timestamptz.go
candiduslynx: Maybe infinity modifier has to be included, too?

candiduslynx (approved):

schema/int8_array.go
hermanschaaf: I think this one is probably okay, the name even has `8` in it, and we're unlikely to redefine the size of 64-bit ints :) Otherwise we have to write:

```
if len(dst.Elements) == 0 {
     return 0
}
return dst.Elements[0] * len(dst.Elements)
```

And then there's still the assumption that all the elements have the same size. 

We could also do:

```
	totalSize := 0
	for _, element := range dst.Elements {
		totalSize += element.Size()
	}
	return totalSize
```

but that would be unnecessarily inefficient, being O(N) instead of O(1)

hermanschaaf (approved):

PR #512, opened 2022-12-17
fix: Improve error wrapping

yevgenypats:


internal/servers/destinations.go
```diff
@@ -87,16 +87,15 @@ func (s *DestinationServer) Write2(msg pb.Destination_Write2Server) error {
 	for {
 		r, err := msg.Recv()
 		if err != nil {
+			close(resources)
 			if err == io.EOF {
-				close(resources)
 				if err := eg.Wait(); err != nil {
-					return fmt.Errorf("got EOF. failed to wait for plugin: %w", err)
+					return fmt.Errorf("got EOF. plugin returned: %w", err)
```
erezrokah, line 93: `EOF` is not an error necessarily (it usually means the stream was closed), so I'm not sure we want to report it. Also to be consistent we should add `plugin returned` to everywhere we do `eg.Wait()`, like lines `105` and `114` in this file.

You can see https://github.com/cloudquery/plugin-sdk/pull/461/files for reference.

We can also reduce the nesting of handling `EOF`, but that can be done in another PR, see https://github.com/cloudquery/plugin-sdk/pull/463 for reference

plugins/destination.go
```diff
@@ -162,14 +162,16 @@ func (p *DestinationPlugin) Write(ctx context.Context, tables schema.Tables, sou
...
+				return fmt.Errorf("context done %v. write client returned: %w", gctx.Err(), err)
+			}
 		case ch <- clientResource:
 		}
 	}
 
 	close(ch)
 	if err := eg.Wait(); err != nil {
-		return err
+		return fmt.Errorf("write client returned: %w", err)
```
erezrokah, line 174:
```suggestion
		return fmt.Errorf("write client failed with error: %w", err)
```

Also shouldn't we use the same message above, i.e. instead `plugin returned` -> `write failed`

erezrokah:
An additional comment on this. [gRPC doesn't support error wrapping](https://github.com/grpc/grpc-go/issues/2934), so using `%w` doesn't really have any value here. You can see an example in https://github.com/cloudquery/plugin-sdk/pull/487/files how to chain multiple errors

yevgenypats:
Closing as this was addressed in additional PR.

PR #513, opened 2022-12-17
fix: Add better logging/metric per table

yevgenypats:
currently we silently drop metrics from child tables. This add them back when root table sync is finished so it is printed once per table/client (and not per object like it was before with millions of log lines)

### DIFF
--- a/plugins/source/scheduler_dfs.go
+++ b/plugins/source/scheduler_dfs.go
@@ -88,7 +88,7 @@ func (p *Plugin) resolveTableDfs(ctx context.Context, table *schema.Table, clien
 	logger := p.logger.With().Str("table", table.Name).Str("client", clientName).Logger()
 
 	if parent == nil { // Log only for root tables, otherwise we spam too much.
-		logger.Info().Msg("root table resolver started")
+		logger.Info().Msg("top level table resolver started")
 	}
 	tableMetrics := p.metrics.TableClient[table.Name][clientName]
 
--- a/plugins/source/scheduler_dfs.go
+++ b/plugins/source/scheduler_dfs.go
@@ -74,15 +74,24 @@ func (p *Plugin) syncDfs(ctx context.Context, spec specs.Source, client schema.C
 	wg.Wait()
 }
 
+func (p *Plugin) logTablesMetrics(tables schema.Tables, client schema.ClientMeta) {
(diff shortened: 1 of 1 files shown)


plugins/source_scheduler_dfs.go
```diff
@@ -74,15 +74,24 @@ func (p *SourcePlugin) syncDfs(ctx context.Context, spec specs.Source, client sc
...
+	}
+}
+
 func (p *SourcePlugin) resolveTableDfs(ctx context.Context, table *schema.Table, client schema.ClientMeta, parent *schema.Resource, resolvedResources chan<- *schema.Resource, depth int) {
 	clientName := client.ID()
 	logger := p.logger.With().Str("table", table.Name).Str("client", clientName).Logger()
 
 	if parent == nil { // Log only for root tables, otherwise we spam too much.
-		logger.Info().Msg("table resolver started")
+		logger.Info().Msg("root table resolver started")
```
erezrokah, line 91:
```suggestion
		logger.Info().Msg("top level table resolver started")
```

I think this is how we refer to these tables in other places

plugins/source_scheduler_dfs.go
```diff
@@ -109,8 +118,9 @@ func (p *SourcePlugin) resolveTableDfs(ctx context.Context, table *schema.Table,
 	}
 
 	// we don't need any waitgroups here because we are waiting for the channel to close
-	if parent == nil { // Log only for root tables, otherwise we spam too much.
-		logger.Info().Uint64("resources", tableMetrics.Resources).Uint64("errors", tableMetrics.Errors).Msg("fetch table finished")
+	if parent == nil { // Log only for root tables and relations only after resolving is done, otherwise we spam per object instead of per table.
+		logger.Info().Uint64("resources", tableMetrics.Resources).Uint64("errors", tableMetrics.Errors).Msg("table sync finished")
+		p.logTablesMetrics(table.Relations, client)
```
erezrokah, line 123:
```suggestion
		p.logTablesMetrics(table.Relations, client)
```

[non blocking] Maybe this could be `p.logTableMetrics(table, client)` and move the log in the above line into the method? This way we don't need to duplicate it

erezrokah (approved):
Looks good, added 2 non blocking comments

PR #514, opened 2022-12-17
fix: Use correct error codes

yevgenypats:
This is uses correct error codes for gRPC in most places. We use Internal and this is the intedded error code for server errors. This is also used by the grpc server as well but it doesn't mean it is only for the gRPC code. See documentation:

https://pkg.go.dev/google.golang.org/grpc/codes#Code
https://cloud.google.com/apis/design/errors

The right codes for internal error are `Unknown` (The default one) and `Internal`


candiduslynx (approved):

erezrokah (approved):
May I suggest we don't set the error code at all and only keep the default `Unknown` for now?

We don't use neither one as we never check the error code on the client side.

I don't have a strong opinion here so approving

yevgenypats:
We indeed are not using it either way as most of the error are internal and thus fatal but this will be just nicer to the user i.e seeing `internal error` rather then `unknown` imo as this put it inline wither other servers.

PR #516, opened 2022-12-17
fix: Move source & destination plugin code to separate packages

yevgenypats:
Im trying to add more functionality and tests to our destination plugin API but this because quite hard now when source and plugin reside in the same place and everything is suffixed with `source` or `destination`.

This should make it more "go" idiomatic by splitting `source` and `destination` into two different packages and help with follow-up PRs. 

This is a breaking change but relatively minor one as it is mostly affecting one file in each plugin (the `main.go` or `plugin.go`)


candiduslynx:
As this is a breaking change, the mod path should be suffixed with `/v2`
@yevgenypats @erezrokah

yevgenypats:
No need to move to `v2` on every breaking change unless it is significant in our case as most of the plugins are maintained by us anyway so this will just make us more work.

erezrokah (approved):
By making this a breaking change it will bump the Go package to `v2` and we'll need to update the imports for all plugins? Do we want that?

I'm good either way (not making it breaking and only updating complication errors)

yevgenypats:
> By making this a breaking change it will bump the Go package to `v2` and we'll need to update the imports for all plugins? Do we want that?
> 
> I'm good either way (not making it breaking and only fixing complication errors)

We don't want to move to v2. I marked it as non breaking.

PR #517, opened 2022-12-18
chore(go): Rename `interface{}` to `any` for readability

candiduslynx:
#### Summary

`any` is an alias to `interface{}` introduced in Go 1.18. It's shorter and using it instead of `interface{}` improves readability a lot.

Marking this PR as a chore, as the functionality is left intact.

---

- [x] Update or add tests 🧪


erezrokah (approved):
This looks good to me, but maybe we should have a lint rule for it?

candiduslynx:
They are interchangeable. Also gen code (see `..._pb.go`) sticks to `interface{}` for backward compatibility. I don't think we need a lint rule here.

erezrokah:
> I don't think we need a lint rule here.

I mean if we chose one or the other (mostly for style I think) we should enforce it so our codebase is consistent.
It doesn't have to be in this PR, but maybe open an issue for it? Otherwise I can see someone re-adding `interface{}`

candiduslynx:
@erezrokah 
I've just added the lint config t check for this.
if you have troubles on Mac with this, install `diffutils` via Brew

hermanschaaf (approved):

yevgenypats (approved):

PR #519, opened 2022-12-19
fix(destination): Rename `NewDestinationPlugin` to `NewPlugin`

erezrokah:
#### Summary

Follow up to https://github.com/cloudquery/plugin-sdk/pull/516

---

- [ ] Update or add tests 🧪


yevgenypats (approved):
Nice. I already did it in my Follow-up PR but I get that otherwise we will need to fix breaking compilation changes twice so better to do it once now.

PR #521, opened 2022-12-21
feat: Add managed API for destination plugins

yevgenypats:
This adds a managed API for destination plugins. 

Closes https://github.com/cloudquery/plugin-sdk/pull/518 as it got destroyed with conflicts.

This should go with that: https://github.com/cloudquery/cloudquery/pull/5805


plugins/destination/plugin.go
```diff
@@ -3,54 +3,127 @@ package destination
...
+
+const (
+	unmanaged writerType = iota
+	managed
 )
 
 type NewClientFunc func(context.Context, zerolog.Logger, specs.Destination) (Client, error)
 
+type ManagedWriter interface {
+	PreWriteTableBatch(ctx context.Context, tables schema.Tables) error
```
hermanschaaf, line 24:
```suggestion
	InitializeTable(ctx context.Context, tables schema.Tables) error
```

I think `PreWriteTableBatch` might give the impression that this function will be called before every call to `WriteTableBatch`. Maybe we should call it something like `OpenTable` (and `CloseTable` for a post-hook). For BigQuery this might instantiate the client, for files this might open the file.

plugins/destination/plugin.go
```diff
@@ -3,54 +3,127 @@ package destination
...
+	WriteTableBatch(ctx context.Context, table *schema.Table, data [][]interface{}) error
+}
+
+type UnimplementedManagedWriter struct{}
```
hermanschaaf, line 28: Why do we need an unimplemented managed writer?

plugins/destination/plugin.go
```diff
@@ -3,54 +3,127 @@ package destination
...
+
+type UnmanagedWriter interface {
+	Write(ctx context.Context, tables schema.Tables, res <-chan *ClientResource) error
+	Metrics() Metrics
```
hermanschaaf, line 32: Why is `Metrics()` part of `UnmanagedWriter`? Shouldn't it be part of the `Client` interface directly?

plugins/destination/plugin.go
```diff
@@ -3,54 +3,127 @@ package destination
...
+
 type Client interface {
 	schema.CQTypeTransformer
-	ReverseTransformValues(table *schema.Table, values []any) (schema.CQTypes, error)
+	ReverseTransformValues(table *schema.Table, values []interface{}) (schema.CQTypes, error)
 	Migrate(ctx context.Context, tables schema.Tables) error
-	Read(ctx context.Context, table *schema.Table, sourceName string, res chan<- []any) error
-	Write(ctx context.Context, tables schema.Tables, res <-chan *ClientResource) error
-	Metrics() Metrics
+	Read(ctx context.Context, table *schema.Table, sourceName string, res chan<- []interface{}) error
```
hermanschaaf, line 57:
```suggestion
	Read(ctx context.Context, table *schema.Table, sourceName string, res chan<- []any) error
```

plugins/destination/plugin.go
```diff
@@ -3,54 +3,127 @@ package destination
...
+	ManagedWriter
+	UnmanagedWriter
 	DeleteStale(ctx context.Context, tables schema.Tables, sourceName string, syncTime time.Time) error
 	Close(ctx context.Context) error
 }
 
 type ClientResource struct {
 	TableName string
-	Data      []any
+	Data      []interface{}
```
hermanschaaf, line 66:
```suggestion
	Data      []any
```

plugins/destination/plugin.go
hermanschaaf: Answered my own question by reading more of the code: looks like `Metrics()` doesn't get called for managed plugins but is implemented internally.

plugins/destination/plugin.go
hermanschaaf: Answering my own question: probably to make it easier for destination plugins to implement the interface without implementing the managed writer for now.

plugins/destination/plugin.go
```diff
@@ -3,54 +3,127 @@ package destination
...
+}
 
-func NewPlugin(name string, version string, newDestinationClient NewClientFunc) *Plugin {
+func WithMaxWorkers(maxWorkers int) Option {
+	return func(p *Plugin) {
+		p.maxWorkers = maxWorkers
+	}
+}
+
+// NewPlugin creates a new unmanaged plugin
```
hermanschaaf, line 108: I think this comment isn't strictly true (or at least complete), because the `WithManagerWriter` option can be passed in to make it use a managed writer?

plugins/destination/plugin.go
```diff
@@ -3,54 +3,127 @@ package destination
...
+
 type Plugin struct {
 	// Name of destination plugin i.e postgresql,snowflake
 	name string
 	// Version of the destination plugin
 	version string
 	// Called upon configure call to validate and init configuration
-	newDestinationClient NewClientFunc
+	newClient NewClientFunc
+	writer    writerType
```
hermanschaaf, line 78:
```suggestion
	writerType    writerType
```

Let's call this field `writerType` so it's not confused with an actual writer

plugins/destination/managed_writer.go
```diff
@@ -0,0 +1,170 @@
...
+					defer wg.Done()
+					p.worker(ctx, metrics, table, ch, flush)
+				}()
+			}
+		} else {
+			p.workers[table.Name].count++
+		}
+		// we save this locally because we don't want to access the map after that so we can
+		// keep the workersLock for as short as possible
+		workers[table.Name] = p.workers[table.Name]
```
hermanschaaf, line 124: This works, but technically I think it isn't necessary as long as we ensure all writes to the map have finished by the time any reads happen lower down.

plugins/destination/plugin.go
hermanschaaf: We should check why the linter isn't complaining

plugins/destination/managed_writer.go
```diff
@@ -0,0 +1,170 @@
...
+	if err := p.client.PreWriteTableBatch(ctx, tables); err != nil {
+		p.workersLock.Unlock()
+		return err
+	}
+	for _, table := range tables.FlattenTables() {
+		table := table
+		if p.workers[table.Name] == nil {
+			ch := make(chan schema.CQTypes)
+			flush := make(chan chan bool)
+			wg := &sync.WaitGroup{}
```
hermanschaaf, line 105: It looks like we never wait on this waitgroup?

plugins/destination/managed_writer.go
```diff
@@ -0,0 +1,170 @@
...
+			p.workers[table.Name] = &worker{
+				count: 1,
+				ch:    ch,
+				flush: flush,
+				wg:    wg,
+			}
+			for i := 0; i < p.spec.Workers; i++ {
```
hermanschaaf, line 112: Is the idea here to have multiple workers per table, or why are we ranging over the number of workers here? I'm a bit confused because the map above initializes count as 1. I may need to read through the code a bit more to understand

plugins/destination/plugin.go
hermanschaaf: Follow-up comment: It seems like the need for `UnimplementedManagedWriter` is a consequence of the destination Plugin interface now needing to support two different types of writers. It also leads us to needing to check whether we're using a managed or unmanaged writer. Rather than overload the Plugin interface, maybe we should create a new ManagedPlugin interface? (This could also have a BasePlugin interface that gets used by both `Plugin` and `ManagedPlugin`)

plugins/destination/plugin_testing.go
```diff
@@ -88,15 +85,18 @@ func (s *TestSuite) destinationPluginTestWriteOverwrite(ctx context.Context, p *
 		return fmt.Errorf("expected 2 resource, got %d", len(resourcesRead))
 	}
 
-	if resource.Data.Equal(resourcesRead[0]) {
-		return fmt.Errorf("expected data to be %v, got %v", resource.Data, resourcesRead[0])
+	if diff := resource.Data.Diff(resourcesRead[0]); diff != "" {
+		return fmt.Errorf("expected first resource diff: %s", diff)
+		// return fmt.Errorf("expected first resource to be (len:%d):\n%v\ngot(len:%d):\n%v", len(resource.Data), resource.Data, len(resourcesRead[0]), resourcesRead[0])
```
hermanschaaf, line 90:
```suggestion
```

Can be removed

plugins/destination/plugin_testing.go
```diff
@@ -88,15 +85,18 @@ func (s *TestSuite) destinationPluginTestWriteOverwrite(ctx context.Context, p *
...
 	}
 
-	if resource2.Data.Equal(resourcesRead[1]) {
-		return fmt.Errorf("expected data to be %v, got %v", resource.Data, resourcesRead[1])
+	if diff := resource2.Data.Diff(resourcesRead[1]); diff != "" {
+		return fmt.Errorf("expected second resource diff: %s", diff)
+		// return fmt.Errorf("expected second resource to be:\n%v\ngot:\n%v", resource.Data, resourcesRead[1])
```
hermanschaaf, line 95:
```suggestion
		// return fmt.Errorf("expected second resource to be:\n%v\ngot:\n%v", resource.Data, resourcesRead[1])
```

Can be removed

plugins/destination/plugin_testing.go
```diff
@@ -108,15 +108,17 @@ func (s *TestSuite) destinationPluginTestWriteOverwrite(ctx context.Context, p *
...
 	if len(resourcesRead) != 2 {
-		return fmt.Errorf("expected 2 resources, got %d", len(resourcesRead))
+		return fmt.Errorf("after overwrite expected 2 resources, got %d", len(resourcesRead))
 	}
 
-	if resource.Data.Equal(resourcesRead[0]) {
-		return fmt.Errorf("expected data to be %v, got %v", resource.Data, resourcesRead[0])
+	if diff := resource2.Data.Diff(resourcesRead[0]); diff != "" {
+		return fmt.Errorf("after overwrite expected first resource diff: %s", diff)
+		// return fmt.Errorf("expected second resource to be:\n%v\ngot:\n%v", resource.Data, resourcesRead[1])
```
hermanschaaf, line 116:
```suggestion
```

plugins/destination/plugin_testing.go
```diff
@@ -130,17 +132,18 @@ func (s *TestSuite) destinationPluginTestWriteOverwrite(ctx context.Context, p *
...
 	if len(resourcesRead) != 1 {
-		return fmt.Errorf("expected 1 resource, got %d", len(resourcesRead))
+		return fmt.Errorf("expected 1 resource after delete stale, got %d", len(resourcesRead))
 	}
 
-	if resource2.Data.Equal(resourcesRead[0]) {
-		return fmt.Errorf("expected data to be %v, got %v", resource.Data, resourcesRead[0])
+	if diff := resource.Data.Diff(resourcesRead[0]); diff != "" {
+		return fmt.Errorf("after delete stale expected second resource diff: %s", diff)
+		// return fmt.Errorf("expected data to be %v, got %v", resource.Data, resourcesRead[0])
```
hermanschaaf, line 140:
```suggestion
```

plugins/destination/plugin_testing.go
```diff
@@ -155,28 +158,32 @@ func (s *TestSuite) destinationPluginTestWriteAppend(ctx context.Context, p *Plu
...
-		Data:      testdata.TestData(),
+		Data:      testdata.GenTestData(table),
 	}
 
 	if !s.tests.SkipSecondAppend {
 		// we dont use time.now because looks like there is some strange
 		// issue on windows machine on github actions where it returns the same thing
 		// for all calls.
-		secondSyncTime := syncTime.Add(time.Second).UTC()
+		secondSyncTime := syncTime.Add(10 * time.Second).UTC()
```
hermanschaaf, line 182: Just curious why this had to be increased to 10 seconds?

plugins/destination/plugin_testing.go
```diff
@@ -195,25 +202,36 @@ func (s *TestSuite) destinationPluginTestWriteAppend(ctx context.Context, p *Plu
 		return fmt.Errorf("expected %d resources, got %d", expectedResource, len(resourcesRead))
 	}
 
-	if resource.Data.Equal(resourcesRead[0]) {
-		return fmt.Errorf("expected data to be %v, got %v", resource.Data, resourcesRead[0])
+	if diff := resource.Data.Diff(resourcesRead[0]); diff != "" {
+		return fmt.Errorf("first expected resource diff: %s", diff)
+		// return fmt.Errorf("expected first resource to be (len:%d):\n%v\ngot(len:%d):\n%v", len(resource.Data), resource.Data, len(resourcesRead[0]), resourcesRead[0])
```
hermanschaaf, line 207:
```suggestion
```

plugins/destination/plugin_testing.go
```diff
@@ -195,25 +202,36 @@ func (s *TestSuite) destinationPluginTestWriteAppend(ctx context.Context, p *Plu
...
 	}
+
 	if !s.tests.SkipSecondAppend {
-		if resource.Data.Equal(resourcesRead[1]) {
-			return fmt.Errorf("expected data to be %v, got %v", resource.Data, resourcesRead[1])
+		if diff := resource2.Data.Diff(resourcesRead[1]); diff != "" {
+			return fmt.Errorf("what second expected resource diff: %s", diff)
+			// return fmt.Errorf("expected second resource to be (len:%d):\n%v\ngot (len:%d):\n%v", len(resource2.Data), resource2.Data, len(resourcesRead[1]), resourcesRead[1])
```
hermanschaaf, line 213:
```suggestion
```

plugins/destination/plugin_testing.go
```diff
@@ -195,25 +202,36 @@ func (s *TestSuite) destinationPluginTestWriteAppend(ctx context.Context, p *Plu
...
+	if diff := resource.Data.Diff(resourcesRead[0]); diff != "" {
+		return fmt.Errorf("first expected resource diff: %s", diff)
+		// return fmt.Errorf("expected first resource to be (len:%d):\n%v\ngot(len:%d):\n%v", len(resource.Data), resource.Data, len(resourcesRead[0]), resourcesRead[0])
 	}
+
 	if !s.tests.SkipSecondAppend {
-		if resource.Data.Equal(resourcesRead[1]) {
-			return fmt.Errorf("expected data to be %v, got %v", resource.Data, resourcesRead[1])
+		if diff := resource2.Data.Diff(resourcesRead[1]); diff != "" {
+			return fmt.Errorf("what second expected resource diff: %s", diff)
```
hermanschaaf, line 212:
```suggestion
			return fmt.Errorf("second expected resource diff: %s", diff)
```

plugins/destination/plugin_testing.go
```diff
@@ -108,15 +108,17 @@ func (s *TestSuite) destinationPluginTestWriteOverwrite(ctx context.Context, p *
...
+	if diff := resource2.Data.Diff(resourcesRead[0]); diff != "" {
+		return fmt.Errorf("after overwrite expected first resource diff: %s", diff)
+		// return fmt.Errorf("expected second resource to be:\n%v\ngot:\n%v", resource.Data, resourcesRead[1])
 	}
 
-	if resource2.Data.Equal(resourcesRead[1]) {
-		return fmt.Errorf("expected data to be %v, got %v", resource.Data, resourcesRead[1])
+	if diff := resource.Data.Diff(resourcesRead[1]); diff != "" {
+		return fmt.Errorf("after overwrite expected second resource diff: %s", diff)
+		// return fmt.Errorf("expected first resource to be (len:%d):\n%v\ngot(len:%d):\n%v", len(resource.Data), resource.Data, len(resourcesRead[0]), resourcesRead[0])
```
hermanschaaf, line 121:
```suggestion
```

plugins/destination/unmanaged_writer.go
```diff
@@ -0,0 +1,47 @@
...
+		case <-gctx.Done():
+			close(ch)
+			return eg.Wait()
+		case ch <- clientResource:
+		}
+	}
+
+	close(ch)
+	if err := eg.Wait(); err != nil {
+		return fmt.Errorf("failed waiting to destination clien %w", err)
```
hermanschaaf, line 44:
```suggestion
		return fmt.Errorf("failed waiting for destination client %w", err)
```

schema/timestamptz.go
```diff
@@ -41,7 +41,14 @@ func (dst *Timestamptz) Equal(src CQType) bool {
 	}
 
 	if value, ok := src.(*Timestamptz); ok {
-		return dst.Status == value.Status && dst.Time.Equal(value.Time) && dst.InfinityModifier == value.InfinityModifier
+		if dst.Status != value.Status || dst.InfinityModifier != value.InfinityModifier {
+			return false
+		}
+
+		if dst.Time.Sub(value.Time) < 1*time.Second {
```
hermanschaaf, line 48: Oof, this might bite us later. Why is it necessary to round to 1 second?

plugins/source/plugin.go
```diff
@@ -47,7 +47,7 @@ const (
 func addInternalColumns(tables []*schema.Table) {
 	for _, table := range tables {
 		if c := table.Column("_cq_id"); c != nil {
-			continue
+			panic(fmt.Sprintf("table %s already has column _cq_id", table.Name))
```
hermanschaaf, line 50: I seem to remember a case where exactly this happened, but can't remember what the situation was now. Maybe I'm imagining it :)

specs/destination_test.go
```diff
@@ -174,10 +178,12 @@ spec:
 `,
 		"",
```
hermanschaaf, line 179: Would be nice to give these tests some descriptive names :)

hermanschaaf (changes requested):
Mostly looks good - left a couple of questions / possible points for discussion

plugins/source/plugin.go
yevgenypats: I think we will know given it will be panic :)

specs/destination_test.go
yevgenypats: This isn't the test name but rather the expected error (which is "")

plugins/destination/plugin.go
yevgenypats: I think yes, I'll do something like `InitTable` and `CloseTable`. Given that we don't use it now anywhere so I'll just drop this.

internal/memdb/memdb.go
```diff
@@ -0,0 +1,201 @@
...
+	schema.DefaultTransformer
+	spec          specs.Destination
+	memoryDB      map[string][][]any
+	errOnWrite    bool
+	blockingWrite bool
+}
+
+type Option func(*client)
+
+func WithErrOnWrite() Option {
```
disq, line 27: Intended for testing, these options should have `Test` somewhere (either godoc or in the option name, like `WithTestingErrOnWrite`, `WithTestingBlockingWrite`)

plugins/destination/managed_writer.go
```diff
@@ -0,0 +1,170 @@
...
+
+type worker struct {
+	count int
+	wg    *sync.WaitGroup
+	ch    chan schema.CQTypes
+	flush chan chan bool
+}
+
+// batchTimeout is the timeout for a batch to be sent to the destination if no resources are received
+const batchTimeout = 20 * time.Second
```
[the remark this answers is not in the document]
yevgenypats: Done

schema/timestamptz.go
yevgenypats: yeah. ok I think I work around that with rounding up our auto-generated data in the tests to seconds. I only hit this snag in csvs.

internal/memdb/memdb.go
yevgenypats: those are all under `internal` so it can't be exported anyway for outside use.

plugins/destination/managed_writer.go
```diff
@@ -0,0 +1,162 @@
...
+					p.logger.Err(err).Str("table", table.Name).Int("len", len(resources)).Dur("time", time.Since(start)).Msg("failed to write batch on timeout")
+					// we don't return as we need to continue until channel is closed otherwise there will be a deadlock
+					atomic.AddUint64(&metrics.Errors, uint64(len(resources)))
+				} else {
+					p.logger.Info().Str("table", table.Name).Int("len", len(resources)).Dur("time", time.Since(start)).Msg("batch written successfully on timeout")
+					atomic.AddUint64(&metrics.Writes, uint64(len(resources)))
+				}
+				resources = make([][]interface{}, 0)
+			}
+		case done := <-flush:
```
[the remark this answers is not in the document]
yevgenypats: because write can be called multiple times so we need to wait for a flush from a specific worker and we can't close the channel as other worker might be still using this worker, so we just wanted to do a flush.

specs/destination_test.go
hermanschaaf: Ah okay, sorry, misread it! That's why I think it's better for readability to keep the field labels though 😅

plugins/destination/plugin.go
hermanschaaf: See https://github.com/cloudquery/plugin-sdk/pull/522

plugins/destination/plugin.go
yevgenypats: I already tried that route but it didn't work as NewUnmanaged plugin will mean a new Server and this is something we don't really want as we don't know ahead of time if a plugin is managed or unmanaged.

plugins/destination/managed_writer.go
yevgenypats: we can't ensure it because otherwise we will have to have a lock hold for the whole time of the write instead of for init and teardown

plugins/destination/managed_writer.go
yevgenypats: Yes, good catch. fixed.

plugins/destination/managed_writer.go
yevgenypats: Yeah, I removed this now there can be only one worker per table for now.

plugins/destination/plugin.go
hermanschaaf: Ah I see. Though on the other hand, a plugin will never need to be both managed and unmanaged, so it seems a bit confusing to keep them in one interface and force the use of `UnimplementedManagedWriter` / `UnimplementedWriter` on all plugins? We may not know initially if a plugin will benefit from being managed or not, but we can always start with managed first and then fall back to unmanaged if that turns out to not work for some reason

plugins/destination/plugin_testing.go
yevgenypats: it can be also one second. prob just was debugging something and it was changed. will change it back.

plugins/destination/managed_writer.go
hermanschaaf: I still don't see where we're waiting 🤔

hermanschaaf: Nevermind, I see it's a property of the worker and getting waited on lower down

hermanschaaf (approved):

PR #524, opened 2022-12-21
fix: Codegen: Add Recursive Embedded Struct Unwrapping

shimonp21:
Depended on by https://github.com/cloudquery/cloudquery/pull/5901

Required, because for IncidentLogEntries, the primary-key (`id`) is twice-embedded:

https://github.com/PagerDuty/go-pagerduty/blob/b6aa0d76f875e1de499b2a246c82882f022a0f8a/log_entry.go#L43
https://github.com/PagerDuty/go-pagerduty/blob/b6aa0d76f875e1de499b2a246c82882f022a0f8a/log_entry.go#L30
https://github.com/PagerDuty/go-pagerduty/blob/b6aa0d76f875e1de499b2a246c82882f022a0f8a/client.go#L49

### DIFF
--- a/codegen/table.go
+++ b/codegen/table.go
@@ -37,7 +37,8 @@ type (
 		skipFields           []string
 		structFieldsToUnwrap []string
 
-		unwrapAllEmbeddedStructFields bool
+		unwrapAllEmbeddedStructFields            bool
+		unwrapAllEmbeddedStructFieldsRecursively bool
 
 		logger zerolog.Logger
 	}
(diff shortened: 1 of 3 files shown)


codegen/table.go
```diff
@@ -37,7 +37,8 @@ type (
 		skipFields           []string
 		structFieldsToUnwrap []string
 
-		unwrapAllEmbeddedStructFields bool
+		unwrapAllEmbeddedStructFields            bool
+		unwrapAllEmbeddedStructFieldsRecursively bool
```
disq, line 41:
```suggestion
		unwrapAllEmbeddedAnonymousStructFieldsRecursively bool
```
just kidding. or am I?

PR #529, opened 2022-12-24
fix: Don't hide errors in destination server

yevgenypats:
Closes https://github.com/cloudquery/plugin-sdk/pull/512/files
Closes https://github.com/cloudquery/plugin-sdk/pull/463/files

### DIFF
--- a/internal/servers/destinations.go
+++ b/internal/servers/destinations.go
@@ -86,17 +86,17 @@ func (s *DestinationServer) Write2(msg pb.Destination_Write2Server) error {
 
 	for {
 		r, err := msg.Recv()
-		if err != nil {
-			if err == io.EOF {
-				close(resources)
-				if err := eg.Wait(); err != nil {
-					return status.Errorf(codes.Internal, "failed to wait for plugin: %v", err)
-				}
-				return msg.SendAndClose(&pb.Write2_Response{})
+		if err == io.EOF {
+			close(resources)
+			if err := eg.Wait(); err != nil {
+				return status.Errorf(codes.Internal, "write failed: %v", err)
 			}
+			return msg.SendAndClose(&pb.Write2_Response{})
+		}
+		if err != nil {
 			close(resources)
 			if wgErr := eg.Wait(); wgErr != nil {
-				return status.Errorf(codes.Internal, "failed to receive msg: %v and wait for plugin: ", err, wgErr)
+				return status.Errorf(codes.Internal, "failed to receive msg: %v and write failed: %v ", err, wgErr)
 			}
 			return status.Errorf(codes.Internal, "failed to receive msg: %v", err)
 		}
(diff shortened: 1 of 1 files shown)


internal/servers/destinations.go
```diff
@@ -95,16 +95,16 @@ func (s *DestinationServer) Write2(msg pb.Destination_Write2Server) error {
 				return msg.SendAndClose(&pb.Write2_Response{})
 			}
 			close(resources)
-			if err := eg.Wait(); err != nil {
-				s.Logger.Error().Err(err).Msg("got error. failed to wait for plugin")
+			if wgErr := eg.Wait(); wgErr != nil {
+				return status.Errorf(codes.Internal, "failed to receive msg: %v and wait for plugin: ", err, wgErr)
```
erezrokah, line 99: I think it's not accurate to report `failed to wait for plugin`.
The error returned from the wait group is whatever returned from the wait group Go Routine, i.e. whatever returned from `s.Plugin.Write(ctx, tables, sourceName, syncTime, resources)`, so if `eg.Wait() != nil` it means the `write` failed, not that the `wait` failed.

Please let me know if that makes sense

internal/servers/destinations.go
```diff
@@ -113,9 +113,9 @@ func (s *DestinationServer) Write2(msg pb.Destination_Write2Server) error {
 		case <-ctx.Done():
 			close(resources)
 			if err := eg.Wait(); err != nil {
-				s.Logger.Error().Err(err).Msg("failed to wait")
+				return status.Errorf(codes.Internal, "Context done: %v and failed to wait for plugin: %v",ctx.Err(),  err)
```
erezrokah, line 116: I'm not sure we should report the context error if the write failed. If the write fails the context error is always `context cancelled` so it has no value. See https://github.com/cloudquery/plugin-sdk/pull/460/files#diff-1df6a94b537b58b0f652697d4a670466f1f889553d4a55b654e4a818ea16a02fR113 and https://github.com/cloudquery/plugin-sdk/pull/460#issuecomment-1339230826

internal/servers/destinations.go
```diff
@@ -113,9 +113,9 @@ func (s *DestinationServer) Write2(msg pb.Destination_Write2Server) error {
...
 			}
-			return ctx.Err()
+			return status.Errorf(codes.Internal, "Context done: %v", ctx.Err())
```
erezrokah, line 118: See https://github.com/cloudquery/plugin-sdk/pull/486/files for how to convert context errors to gRPC status codes

erezrokah (changes requested):
Thanks for the PR @yevgenypats, added a few comments. Maybe we should sync on this to ensure I'm not missing anything with our error handling logic.

Also I'm not sure how this closes https://github.com/cloudquery/plugin-sdk/pull/463/files as it doesn't reduce the nesting.

Additionally https://github.com/cloudquery/plugin-sdk/pull/487 proposes a way to report multiple errors instead of using `x and y failed`

erezrokah (approved):

internal/servers/destinations.go
```diff
@@ -86,25 +86,25 @@ func (s *DestinationServer) Write2(msg pb.Destination_Write2Server) error {
...
 			if err := eg.Wait(); err != nil {
-				s.Logger.Error().Err(err).Msg("got error. failed to wait for plugin")
+				return status.Errorf(codes.Internal, "write failed: %v", err)
+			}
+			return msg.SendAndClose(&pb.Write2_Response{})
+		}
+		if err != nil {
+			close(resources)
+			if wgErr := eg.Wait(); wgErr != nil {
+				return status.Errorf(codes.Internal, "failed to receive msg: %v and write failed: %v ", err, wgErr)
```
erezrokah, line 99:
```suggestion
				return status.Errorf(codes.Internal, "failed to receive msg: %v and write failed: %v", err, wgErr)
```

PR #531, opened 2022-12-25
draft: Refactor codegen

shimonp21:
This PR comes together with [cloudquery-5953](https://github.com/cloudquery/cloudquery/pull/5953)

- Use golang's struct-inheritance (embedded-struct) instead of writing custom methods to copy-paste properties from codegen "Resources" to codegen "TableDefinitions"
- Move the `GenerateSchema` method to the SDK - no need to copy-paste it for every new plugin.

### Pros
- Reduce code-duplication: no need to copy-paste `generateSchema`, and no need to do write "field-forwarding" methods.
   - Example of "field-forwarding": https://github.com/cloudquery/cloudquery/blob/d65d80fc24bf2ae1fcda502da5de744f86491afd/plugins/source/aws/codegen/recipes/base.go#L145-L198
   - Duplicate `generateSchema` code per plugin:
      https://github.com/cloudquery/cloudquery/blob/d65d80fc24bf2ae1fcda502da5de744f86491afd/plugins/source/aws/codegen/recipes/base.go#L212-L243
- Allow further removal of code-duplication, by moving creation of the table-lists to the SDK

### Cons
- requires recipe resource definitions to change, due to the way golang works (https://gist.github.com/shimonp21/8907b859d60ec40e932172c0f21fa636):
   ```
   Resource {
   TableDefinition {
     Service: "service",
     SubService: "subservice",
     Struct: SomeStruct{},
   },
   // plugin-specific stuff goes here (resolver-and-mock-test parameters, usually).
   }
   ```


codegen/.snapshots/TestGenerateTemplate-should_add_ignore_in_tests_to_columns
```diff
@@ -1,12 +1,24 @@
...
-		Name:         "with relations",
-		Columns: []schema.Column{
-{
+// Code generated by codegen; DO NOT EDIT.
+
+package 
+
+import (
+	"github.com/cloudquery/plugin-sdk/schema"
+	"github.com/cloudquery/cloudquery/plugins/source//client"
```
hermanschaaf, line 7: This import path has a double forward slash, it doesn't look right 🤔

codegen/.snapshots/TestGenerateTemplate-should_add_comma_between_relations
```diff
@@ -1,13 +1,25 @@
...
 
-		},
+package 
 
-		Relations: []*schema.Table{
-relation1,
-relation2,
+import (
+	"github.com/cloudquery/plugin-sdk/schema"
+	"github.com/cloudquery/cloudquery/plugins/source//client"
```
hermanschaaf, line 7: Double slash in the import path here?

hermanschaaf: Also, maybe we should gofmt the file after generation so the tests aren't sensitive to whitespace changes

hermanschaaf (commented):
Just a couple of fly-by comments from a quick look through

codegen/table.go
```diff
@@ -1,11 +1,19 @@
...
+	"log"
+	"os"
+	"path"
 	"reflect"
+	"strings"
+	"text/template"
 
 	"github.com/cloudquery/plugin-sdk/schema"
-	"github.com/rs/zerolog"
+	"github.com/iancoleman/strcase"
```
erezrokah, line 16: Is there anything preventing use from using the SDK's caser https://github.com/cloudquery/plugin-sdk/tree/b47c8eda53e7a4351b2901754d5b232d04fc7967/caser?

yevgenypats:
Closing as we moved to transformation https://github.com/cloudquery/plugin-sdk/pull/564 and codegen is deprecated now.

PR #532, opened 2022-12-26
test: Add null-in-text-field test

shimonp21:
Closes https://github.com/cloudquery/cloudquery/issues/5255 after we upgrade the sdk's in the dest-plugins

#### Summary

---

- [ ] Update or add tests 🧪

### DIFF
--- a/schema/text.go
+++ b/schema/text.go
@@ -4,6 +4,7 @@ package schema
 import (
 	"database/sql/driver"
 	"fmt"
+	"strings"
 )
 
 type TextTransformer interface {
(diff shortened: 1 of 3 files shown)


erezrokah (approved):
Looks good but is there anything else we should string? Not sure 🙃

schema/text.go
```diff
@@ -19,6 +20,8 @@ func (*Text) Type() ValueType {
 	return TypeString
 }
 
+// Used only in tests - we allow one of the strings to be missing null byttes (since some destinations like)
```
disq, line 23:
```suggestion
// Used only in tests - we allow one of the strings to be missing null bytes (since some destinations like)
```

yevgenypats (changes requested):
Maybe worth live discussion as don't understand the whole context but

 `return dst.Status == s.Status && strings.ReplaceAll(dst.Str, "\x00", "") == strings.ReplaceAll(s.Str, "\x00", "")` seems a line that can cause issues in the future.

yevgenypats (approved):

Issue #413, opened 2022-11-17
feature: Support cursor-based syncing


hermanschaaf:
It has come up a couple of times that it would be good for the SDK to support cursor-based syncing. For APIs that allow it, this would allow resources to be fetched once, with only updates being fetched after the first sync. For some resource types, this optimization would be a significant performance boost.

This would be especially useful in cases where the APIs allow for time-based querying, and it is known that resources never change once they are created (like logs, reports, audit trails).

One concrete example where this would be beneficial is the Heroku audit trail event: https://devcenter.heroku.com/articles/platform-api-reference#audit-trail-event . It allows querying by day, so an optimized implementation would only fetch historical days the first time, then in subsequent runs fetch the days that have been added since the previous sync.

hermanschaaf:
> This will require persisting the last sync timestamp (we already do that) and then executing query beforehand.

Yes, though I think we should generalize it by allowing each table to specify which column should be used. In most cases this would be a strictly increasing ID column or a timestamp. This column should then be queried for the highest value, and subsequent queries started from there onwards.

candiduslynx:
I suppose, a tuple with both timestamp & some increasing ID would be the most generic.
We can use [snowflakes](https://en.wikipedia.org/wiki/Snowflake_ID), too.

PR #542, opened 2022-12-28
feat: Improve destination Read API

hermanschaaf:
This adds new options to the Read API to:
 - limit the number of records returned
 - order the returned records by a set of columns
 - select only rows that match a given value
 - select only a subset of columns

I have chosen not to support filtering using operators (like greater than, less than) as an option for now, as it's a complicated and we don't have a strong need for it. We may want to add it later to make it feature-complete.

This update allows us to fix https://github.com/cloudquery/cloudquery/issues/6069 (will be done in another PR to the CloudQuery repo) and should also allow for cursor-based syncing (https://github.com/cloudquery/plugin-sdk/issues/413). This is not implemented as part of this PR though.


hermanschaaf:
Here is a draft PR for the cloudquery repo to show what usage would look like: https://github.com/cloudquery/cloudquery/pull/6107

hermanschaaf:
I see the unit tests are failing: will mark this as a draft for now and look into it tomorrow

yevgenypats:
I like the interface but have a few questions/thought:

Do we really need the `LessThan` for all types? also for some types Im not sure there is such a think, like for `json` or for `bytea`. can we maybe have additional api for destinations called `ReadOne` which gets and ID? This way it will be consistent? and can also serve for the cursor based thing?

hermanschaaf:
@yevgenypats Indeed, the meaning of `LessThan` is unclear for some types, but I don't foresee us using those any time soon, because just like with a JSON database column, sorting on it would be ill-advised. I'm not sure I want to necessarily panic there though. In these cases any kind of ordering is okay, as long as its consistent.

A single `ReadOne` function, though simpler, will not be enough. I see your point and it could work for the testing case in isolation, because we already know the IDs we want to read (because we just wrote them). But then the test wouldn't be able to assert that the correct number of records were inserted. Maybe we can live with this. 

However, when we get to cursor-based syncing, we don't know the ID we're looking for up-front. Let's take a few examples:
 - message history for the Slack plugin: here we need to fetch the latest `ts` (timestamp) that was inserted into the database in a previous sync. The only information we have before reading is the source name. There is no one ID that we can look up by, but we can get the latest timestamp by issuing an `order by ts desc limit 1` read query.  
 - service stats for the Fastly plugin: here we need to know the latest `start_time` for a given service that was inserted in a previous sync. So we'll need to issue an `where service_id = $1 order by start_time limit 1` query (aside: this made me realize that we will need some filtering capabilities soon after all)
 - stories for a hypothetical Hacker News plugin: here we need to know the latest story ID that was inserted into the database. We'd want to issue a `order by id desc limit 1` read query.

The important thing is that we have `OrderBy`. To implement `OrderBy` for the memory db and file types like CSV and JSON, we need `LessThan` on the types. For databases we can rely on the built-in `ORDER BY` functionality.

Makefile
```diff
@@ -4,7 +4,7 @@ test:
 
 .PHONY: lint
 lint:
-	golangci-lint run
+	golangci-lint run --config .golangci.yml
```
erezrokah, line 7: [non blocking] I don't think we need to specify the configuration here as `golangci-lint` looks for it by default https://golangci-lint.run/usage/configuration/#config-file

internal/memdb/memdb.go
```diff
@@ -105,24 +106,67 @@ func (c *client) Migrate(_ context.Context, tables schema.Tables) error {
...
-			sortedRes = append(sortedRes, row)
+		if row[sourceColIndex].(*schema.Text).Str != source {
+			continue
+		}
+		if !match(opts.Where, table, row) {
+			continue
+		}
+		sortedRes = append(sortedRes, row)
+	}
+	sort.Slice(sortedRes, func(i, j int) bool {
```
erezrokah, line 128:
```suggestion
	sort.SliceStable(sortedRes, func(i, j int) bool {
```

plugins/destination/plugin_testing.go
```diff
@@ -154,37 +150,32 @@ func (s *PluginTestSuite) destinationPluginTestWriteAppend(ctx context.Context,
...
 		// write second time
-		if err := p.writeOne(ctx, tables, sourceName, secondSyncTime, resource2); err != nil {
+		if err := p.writeOne(ctx, tables, sourceName, secondSyncTime, resources[1]); err != nil {
 			return fmt.Errorf("failed to write one second time: %w", err)
 		}
 	}
 
-	resourcesRead, err := p.readAll(ctx, tables[0], sourceName)
+	cqIDIndex := table.Columns.Index(schema.CqIDColumn.Name)
+	sort.Slice(resources, func(i, j int) bool {
```
erezrokah, line 170:
```suggestion
	sort.SliceStable(resources, func(i, j int) bool {
```

erezrokah (approved):
Added a few comments, but I think I'm missing something regarding the reason this is required for testing purposes.
Could we update the equality check in the tests to ignore the order of the results instead?

I get that we might need this for a future feature (curser based fetching), but since the read is only used in tests, maybe we could find a solution that doesn't require adding functionality to non test related code?

As for cursor based fetching, maybe we could add the order by functionality once we implement that feature, as we might discover more requirements once we try to implement cursor based fetching.

Approving the code change regardless of ⬆️

hermanschaaf:
@erezrokah Yeah, I was working on this with the eye on it being the first step towards cursor-based syncing (rather than doing it all in one or two massive PRs). However after seeing how it works out, I think we'll probably not go with this Read API approach and use a different interface to achieve cursor-based syncing.

I'll simplify this PR to focus on fixing the tests and mostly try and leave the Read API untouched.

PR #544, opened 2022-12-30
chore: Add benchmark with rate limiting

hermanschaaf:
This refactors the scheduling benchmark a little to accommodate a new "rate-limited" benchmark. The idea of this benchmark is to apply back-pressure on a per-client-table basis if too many calls are made in a given window. A good scheduler for this scenario would spread the calls across all table-client combinations so as to not saturate any one table-client.

Because this makes no changes to the DFS scheduling code, the benchmark currently takes about 20 seconds to run on my machine. With appropriate changes to the scheduling, I believe we will be able to make it run in 2-3 seconds.


hermanschaaf:
Closing in favor of https://github.com/cloudquery/plugin-sdk/pull/545

PR #545, opened 2022-12-30
feat: Add scheduler option and introduce Round Robin scheduler

hermanschaaf:
This adds an optional Scheduler option to source configs. It still defaults to DFS, our current scheduler. However, for cases like GCP, alternative scheduling algorithms may be beneficial for large deployments (as explained in https://github.com/cloudquery/cloudquery/issues/5389). For this reason, this change also introduces a new (currently opt-in only) scheduler called `round-robin`. 

A new benchmark scenario has been added to showcase when round robin really shines. In this scenario, strict global rate limits mean that waiting for each table to finish all of its clients is not the most efficient schedule. Instead, a schedule that interleaves the clients for all tables (as round robin does) performs near optimally, and the benchmarks show this difference:

```
BenchmarkDefaultConcurrencyDFS
BenchmarkDefaultConcurrencyDFS-8              	       1	     10835 resources/s	73788592 B/op	 1057929 allocs/op
BenchmarkDefaultConcurrencyRoundRobin
BenchmarkDefaultConcurrencyRoundRobin-8       	       2	     11276 resources/s	72121956 B/op	 1054439 allocs/op
BenchmarkTablesWithChildrenDFS
BenchmarkTablesWithChildrenDFS-8              	       1	     31900 resources/s	465282072 B/op	 6859524 allocs/op
BenchmarkTablesWithChildrenRoundRobin
BenchmarkTablesWithChildrenRoundRobin-8       	       1	     33276 resources/s	463475144 B/op	 6856392 allocs/op
BenchmarkTablesWithRateLimitingDFS
BenchmarkTablesWithRateLimitingDFS-8          	       1	        28.30 resources/s	 9098456 B/op	  132129 allocs/op
BenchmarkTablesWithRateLimitingRoundRobin
BenchmarkTablesWithRateLimitingRoundRobin-8   	       1	       821.5 resources/s	 8213856 B/op	  112748 allocs/op
PASS
```

In the `BenchmarkTablesWithRateLimiting` case, Round Robin completes in just over 1s, DFS completes in 35s.

## Notes for Reviewers

- No functional changes to the DFS scheduler were made in this PR, but some functions were moved to a different file so they can be re-used by Round Robin.
-  I'm also removing `targetResources/s` from the benchmarks in this PR. They were useful once, but with rate limiting scenarios it's becoming too difficult to maintain accuracy.


plugins/source/benchmark_test.go
```diff
@@ -270,46 +246,180 @@ func nMultiplexer(n int) schema.Multiplexer {
...
+		mean:   mean,
+		stdDev: stdDev,
+	}
+}
+
+func (c *DefaultClient) Call(_, _ string) error {
+	sample := int(rand.NormFloat64()*float64(c.stdDev) + float64(c.mean))
+	duration := time.Duration(sample)
+	if duration < c.min {
+		time.Sleep(c.min)
```
disq, line 331: `duration = c.min` and remove return?

plugins/source/scheduler_round_robin.go
```diff
@@ -0,0 +1,197 @@
...
+	go p.periodicMetricLogger(logCtx, &logWg)
+
+	var wg sync.WaitGroup
+	type tableClient struct {
+		table  *schema.Table
+		client schema.ClientMeta
+	}
+
+	// interleave table-clients so that we get:
+	// table1-client1, table2-client1, table3-client, table1-client2, table2-client2, table3-client2, ...
```
disq, line 58:
```suggestion
	// table1-client1, table2-client1, table3-client1, table1-client2, table2-client2, table3-client2, ...
```

plugins/source/scheduler_round_robin.go
```diff
@@ -0,0 +1,197 @@
...
+	}()
+
+	var wg sync.WaitGroup
+	for resource := range resourcesChan {
+		resource := resource
+		resolvedResources <- resource
+		for _, relation := range resource.Table.Relations {
+			relation := relation
+			if err := p.tableSems[depth].Acquire(ctx, 1); err != nil {
+				// This means context was cancelled
```
disq, line 184: An opportunity to use `goto` (add a label in L196 where it `wg.Wait()`s and the fn ends) was missed here...

plugins/source/scheduler_round_robin.go
hermanschaaf: Updated, thanks!

plugins/source/benchmark_test.go
hermanschaaf: Good suggestion 👍

plugins/source/scheduler_round_robin.go
```diff
@@ -87,9 +82,11 @@ func (p *Plugin) syncRoundRobin(ctx context.Context, spec specs.Source, client s
 		go func() {
 			defer wg.Done()
 			defer p.tableSems[0].Release(1)
-			// not checking for error here as nothing much todo.
+			// not checking for error here as nothing much to do.
 			// the error is logged and this happens when context is cancelled
-			p.resolveTableRoundRobin(ctx, table, cl, nil, resolvedResources, 1)
+			// Round Robin currently uses the DFS algorithm to resolve the tables, but this
+			// may change in the future.
+			p.resolveTableDfs(ctx, table, cl, nil, resolvedResources, 1)
```
disq, line 89: Nice!

plugins/source/scheduler.go
```diff
@@ -0,0 +1,120 @@
+package source
```
hermanschaaf, line 1: Everything in this file was copied from `scheduler_dfs.go` with no changes

erezrokah (approved):
Looks great 🚀 To confirm, this only does round robin at the top level correct?
(Not that we need to do it for relations, just wondering)

plugins/source/plugin.go
```diff
@@ -159,7 +169,14 @@ func (p *Plugin) Sync(ctx context.Context, spec specs.Source, res chan<- *schema
...
 	}
 	startTime := time.Now()
-	p.syncDfs(ctx, spec, c, tables, res)
+	switch spec.Scheduler {
+	case SchedulerDFS:
+		p.syncDfs(ctx, spec, c, tables, res)
+	case SchedulerRoundRobin:
+		p.syncRoundRobin(ctx, spec, c, tables, res)
+	default:
+		return fmt.Errorf("unknown scheduler %s. Options are: %v", spec.Scheduler, allSchedulers)
```
candiduslynx, line 178: default to dfs

plugins/source/plugin.go
```diff
@@ -12,6 +12,16 @@ import (
 	"golang.org/x/sync/semaphore"
 )
 
+const (
+	SchedulerDFS        = "dfs"
```
candiduslynx, line 16: [nit]: use a typed consts instead of strings?

plugins/source/scheduler_round_robin.go
```diff
@@ -0,0 +1,99 @@
...
+			wg.Wait()
+			// gracefully shut down the logger goroutine
+			logCancel()
+			logWg.Wait()
+			return
+		}
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			defer p.tableSems[0].Release(1)
```
candiduslynx, line 84: [nit]: Is lookup for `p.tableSems[0]` necessary, especially, in a loop? Shouldn't it be extracted?

specs/source.go
```diff
@@ -57,6 +60,9 @@ func (s *Source) SetDefaults() {
 	if s.Concurrency == 0 {
 		s.Concurrency = defaultConcurrency
 	}
+	if s.Scheduler == "" {
+		s.Scheduler = "dfs"
```
candiduslynx, line 64: move this check to the `plugins/source/plugin.go`, see comment there

hermanschaaf:
> @erezrokah To confirm, this only does round robin at the top level correct?

Yes, correct. I'm working on the assumption that the same rate limits will probably be applied to a table and its children.

specs/source.go
```diff
@@ -94,6 +100,10 @@ func (s *Source) Validate() error {
 		}
 	}
 
+	if s.Scheduler != "dfs" && s.Scheduler != "round-robin" {
```
candiduslynx, line 103: nit: if you make scheduler option a typed one, there's a quick check pattern:
```go
type SchedulerFlavor byte
const (
    SchedulerDFS SchedulerFlavor = iota
    SchedulerRR
    _badSchedulerOption
)

...
if s.Scheduler >= _badSchedulerOption { ...
```

candiduslynx (commented):
several comments to be considered+ a couple of nits

plugins/source/plugin.go
hermanschaaf: The default is handled elsewhere (in SetDefaults). We want the user to know if they've entered an invalid scheduler.

plugins/source/scheduler_round_robin.go
hermanschaaf: Not sure, but it seems like a micro-optimization and not relevant to this PR, as this is not unique to the round robin scheduler. Feel free to try it and see if it makes a difference in the benchmarks.

specs/source.go
hermanschaaf: Disagree, we should set a default if empty, but tell the user if they entered something invalid.

specs/source.go
hermanschaaf: An import cycle prevents us from using the constants from the other package here. We'd have to put the constants somewhere else first, maybe to this package

plugins/source/plugin.go
hermanschaaf: Updated 👍

specs/source.go
hermanschaaf: Updated by moving the constants to this package

specs/scheduler.go
```diff
@@ -0,0 +1,62 @@
...
+		if i > 0 {
+			buffer.WriteString(", ")
+		}
+		buffer.WriteString(scheduler.String())
+	}
+	return buffer.String()
+}
+
+func (s Scheduler) String() string {
+	return AllSchedulerNames[s]
```
candiduslynx, line 33: To be more confident, here's a little trick:
```go
var AllSchedulerNames = [...]string{
	SchedulerDFS:        "dfs",
	SchedulerRoundRobin: "round-robin",
}
```

https://go.dev/play/p/IDDk0f9sBpw

specs/scheduler.go
```diff
@@ -0,0 +1,62 @@
...
+}
+func (s Scheduler) MarshalJSON() ([]byte, error) {
+	buffer := bytes.NewBufferString(`"`)
```
candiduslynx, line 36: why not simple strings concat?

candiduslynx (approved):
only several more-or-less-nit comments

specs/scheduler.go
hermanschaaf: There are many ways to skin a concat

specs/scheduler.go
hermanschaaf: Interesting, I hadn't seen that syntax before!

candiduslynx (approved):

plugins/source/scheduler_round_robin.go
```diff
@@ -0,0 +1,99 @@
...
+		addedNew := false
+		for i, table := range tables {
+			if c < len(preInitialisedClients[i]) {
+				tableClients = append(tableClients, tableClient{table: table, client: preInitialisedClients[i][c]})
+				addedNew = true
+			}
+		}
+		c++
+		if !addedNew {
+			break
```
yevgenypats, line 66: Im not sure this will work. what if one table has less clients then the others? then it wont go to add others.

yevgenypats (changes requested):
Looks good but I think there is a bug in the interleave logic which skips some clients in some cases.

plugins/source/scheduler_round_robin.go
hermanschaaf: The code was written to take that case into account, so I think it should work. The algorithm is roughly like this:

```
set counter to 1
add first client for all tables with >= 1 clients
set counter to 2
add second client for all tables with >= 2 clients
...
set counter to N
add Nth client for all tables with >= N clients
set counter to N+1
stop because no tables are found to have >= N+1 clients
```

I believe it handles all cases, but I've extracted this into a function now and added a test for it, so we can be sure.

yevgenypats (approved):

PR #561, opened 2023-01-02
fix(testing): Sort results before comparison in tests

hermanschaaf:
Extracted from https://github.com/cloudquery/plugin-sdk/pull/542 to allow us to fix https://github.com/cloudquery/cloudquery/issues/6069

Before this change, running cockroachDB tests:

```
Jan  2 11:03:08.760000 INF Query args=["testAppendSource474bca19-69ba-4a5a-8af0-3d890198a243"] commandTag="SELECT 2" module=pgx pid=1570743 sql="SELECT * FROM \"cq_test_write_append\" WHERE _cq_source_name = $1 order by _cq_sync_time asc"
--- FAIL: TestPgPlugin (0.36s)
    --- FAIL: TestPgPlugin/TestWriteOverwrite (0.22s)
        client_test.go:20: expected first resource diff: value at 2 is(d1ebd2a374f14556852c4d5b23608201) TypeUUID while other is(3da5b2eafc814ecb973bcee770bb70e1) TypeUUID
            value at 3 is(d03dd7f8781a4dbb93aaebefb6aaba87) TypeUUID while other is(f16b7d3a0f85441daf30da5d2d8afb26) TypeUUID
            value at 7 is(758c419b5a1b4a46ba45499b402c6aaf) TypeUUID while other is(0f252e92137b4f55a856cbf97d43d098) TypeUUID

FAIL
FAIL	github.com/cloudquery/cloudquery/plugins/destination/postgresql/client	0.640s
```

After this change:

```
ok  	github.com/cloudquery/cloudquery/plugins/destination/postgresql/client	0.662s
```

### DIFF
--- a/plugins/destination/plugin_testing.go
+++ b/plugins/destination/plugin_testing.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"sort"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


plugins/destination/plugin_testing.go
```diff
@@ -253,3 +230,31 @@ func PluginTestSuiteRunner(t *testing.T, p *Plugin, spec any, tests PluginTestSu
...
+		_ = resource.Data[0].Set(sourceName)
+		_ = resource.Data[1].Set(syncTime)
+		resources[i] = resource
+	}
+	return resources
+}
+
+func sortResources(table *schema.Table, resources []schema.DestinationResource) {
+	cqIDIndex := table.Columns.Index(schema.CqIDColumn.Name)
+	sort.Slice(resources, func(i, j int) bool {
```
erezrokah, line 250:
```suggestion
	sort.SliceStable(resources, func(i, j int) bool {
```

plugins/destination/plugin_testing.go
```diff
@@ -253,3 +230,31 @@ func PluginTestSuiteRunner(t *testing.T, p *Plugin, spec any, tests PluginTestSu
...
+		return resources[i].Data[cqIDIndex].String() < resources[j].Data[cqIDIndex].String()
+	})
+}
+
+func sortCQTypes(table *schema.Table, resources []schema.CQTypes) {
+	cqIDIndex := table.Columns.Index(schema.CqIDColumn.Name)
+	sort.Slice(resources, func(i, j int) bool {
```
erezrokah, line 257:
```suggestion
	sort.SliceStable(resources, func(i, j int) bool {
```

erezrokah (approved):

yevgenypats (approved):

yevgenypats:
Nice!

plugins/destination/plugin_testing.go
hermanschaaf: It's a two-element slice being sorted on UUIDs, I don't think it really maters 😄

PR #563, opened 2023-01-02
feat: Add primary key validation

yevgenypats:


schema/resource.go
```diff
@@ -101,6 +101,19 @@ func (r *Resource) Columns() []string {
 	return r.Table.Columns.Names()
 }
 
+// Validates various constraints. for exmaple
+// does all primary keys has values
```
hermanschaaf:
```suggestion
// Validates various constraints. For example,
// checks that all primary keys have values.
```

plugins/source/scheduler_dfs.go
```diff
@@ -169,6 +169,12 @@ func (p *Plugin) resolveResourcesDfs(ctx context.Context, table *schema.Table, c
 				if resolvedResource == nil {
 					return
 				}
+				tableMetrics := p.metrics.TableClient[table.Name][client.ID()]
```
hermanschaaf, line 172: Maybe this lookup should happen inside the `if`, because that's the only place it's used?

hermanschaaf (approved):

schema/resource.go
```diff
@@ -101,6 +101,18 @@ func (r *Resource) Columns() []string {
 	return r.Table.Columns.Names()
 }
 
+// Validates various constraints. For example: Does all primary keys has values.
```
erezrokah, line 104:
```suggestion
// Validates that all primary keys have values.
```

I believe the comment should reflect what the function does now. We can change the comment if/when we add more functionality

schema/resource.go
```diff
@@ -101,6 +101,18 @@ func (r *Resource) Columns() []string {
...
+func (r *Resource) Validate() error {
+	for i, c := range r.Table.Columns {
```
erezrokah, line 106: How about we don't return early and report in case multiple primary keys are missing instead of only the first one?

erezrokah (approved):
Looks good, nothing blocking

plugins/source/scheduler_dfs.go
yevgenypats: Good catch.

schema/resource.go
yevgenypats: Good idea. Added.

erezrokah:
FYI, this validation caught an issue in our tests, see https://github.com/cloudquery/cloudquery/pull/6256/commits/5fc8b1f9739a0894625566767b6467f36afb5b7d and https://github.com/cloudquery/cloudquery/actions/runs/3828313454/jobs/6513763728#step:7:115
🎉

PR #564, opened 2023-01-02
feat: Add Transformer for tables (codegen replacement)

yevgenypats:


hermanschaaf (approved):
Looks good, would definitely be nice to not need codegen! I guess developers will still have to rely on docs generation to see what the output is though, so there is still some codegen in the loop.

I'm also curious to see what the performance of this will be like with hundreds of tables, and whether it will be good enough to run without some kind of caching (*cough* codegen *cough*) step :)

erezrokah (approved):
Looks good 🚀 

Same concerns as @hermanschaaf as we're moving the generation to runtime and also lose some visibility on what's generated.

We should probably have some test that runs on all tables and ensures the transformation works (we don't have mock tests coverage for all plugins). That test could also serve as a benchmark to know how much time it takes to start a plugin

> Also to confirm - this removes only the table codegen, but we still need resolvers codegen in each plugin correct?

transformers/struct.go
```diff
@@ -0,0 +1,151 @@
+package transformers
```
disq, line 1: 🤖

disq:
Apart from the missing modifiers to add `ColumnOptions`/PKs, or Resolvers without redefining the whole column, this method of operation removes transparency as to what is actually being generated. There's no way to debug what's happening or no way to use it as a one-time generator (to then edit the result to fit) which in turn makes things more closed and harder to understand.

yevgenypats:
> Apart from the missing modifiers to add `ColumnOptions`/PKs, or Resolvers without redefining the whole column, this method of operation removes transparency as to what is actually being generated. There's no way to debug what's happening or no way to use it as a one-time generator (to then edit the result to fit) which in turn makes things more closed and harder to understand.

Indeed this is the downside but to understand what the transformation does it generate the markdown which shows exactly what columns are created after transformation. The otherway around creates an unmaintainable codegen code.

PR #569, opened 2023-01-03
feat: Add local backend for storing cursor state

hermanschaaf:
A state store allows us to store cursors for https://github.com/cloudquery/plugin-sdk/issues/413. This first iteration introduces the interface and implements it for a local JSON storage backend. After some back and forth I decided to call it `backend`, similar to Terraform. Alternatives considered were `state` (confusing because it's not a single state) and `storage` (confusing because it's similar to `destination`). 

Since the interface is a generic key-value store, we lose the ability to know whether a table is incremental or not, and which columns are used as cursor keys if it is. I'm proposing that we add some fields to the `Table` definition to be able to at least document these aspects, but it's not enforced.

See https://github.com/cloudquery/cloudquery/pull/6336 for an example implementation using this new functionality.


internal/backends/local/local.go
```diff
@@ -0,0 +1,157 @@
...
+		tables[table] = kv
+	}
+	return tables, nil
+}
+
+func (l *Local) readFile(name string) (table string, kv entries, err error) {
+	p := path.Join(l.spec.Path, name)
+	f, err := os.Open(p)
+	if err != nil {
+		return "", nil, fmt.Errorf("failed to open state file %v: %w", name, err)
```
disq, line 75: stdlib functions used in this method all wrap their errors with `os.PathError` (99% of the time) so the extra wrapping is unnecessary.

schema/column.go
```diff
@@ -26,10 +30,11 @@ type Column struct {
 	Type ValueType `json:"type,omitempty"`
 	// Description about column, this description is added as a comment in the database
 	Description string `json:"-"`
-	// Column Resolver allows to set you own data based on resolving this can be an API call or setting multiple embedded values etc'
+	// Column Resolver allows to set your own data for a column; this can be an API call, setting multiple embedded values, etc
 	Resolver ColumnResolver `json:"-"`
 	// Creation options allow modifying how column is defined when table is created
-	CreationOptions ColumnCreationOptions `json:"creation_options,omitempty"`
+	CreationOptions  ColumnCreationOptions `json:"creation_options,omitempty"`
+	IsIncrementalKey bool                  `json:"is_incremental_key,omitempty"`
```
disq, line 37: Missing godoc (all the other fields have explanations)

hermanschaaf: Nice catch, I think this field got added here by accident actually

disq (approved):

internal/backends/local/local.go
hermanschaaf: I'll remove the name of the file but keep some wrapping so we can identify where the error happened

codegen/templates/column.go.tpl
```diff
@@ -12,6 +12,9 @@
     {{- if .Options.PrimaryKey}}
       PrimaryKey: true,
     {{- end }}
+    {{- if .Options.IncrementalKey}}
```
yevgenypats, line 15: Do we need this in codegen given we are deprecating this?

plugins/source/plugin.go
```diff
@@ -5,14 +5,16 @@ import (
...
 	"github.com/cloudquery/plugin-sdk/caser"
+	"github.com/cloudquery/plugin-sdk/internal/backends/local"
 	"github.com/cloudquery/plugin-sdk/schema"
 	"github.com/cloudquery/plugin-sdk/specs"
 	"github.com/rs/zerolog"
 	"golang.org/x/sync/semaphore"
 )
 
-type NewExecutionClientFunc func(context.Context, zerolog.Logger, specs.Source) (schema.ClientMeta, error)
+type NewExecutionClientFunc func(context.Context, zerolog.Logger, specs.Source, backend.Backend) (schema.ClientMeta, error)
```
yevgenypats, line 17: Let's introduce an optional struct here so we can have a way of both adding required parameters that will break the API and once that are not required like this one (i.e not all plugins have to implement/use cursor)

internal/backends/local/local.go
```diff
@@ -0,0 +1,157 @@
...
+
+	if _, ok := l.tables[table]; !ok {
+		return "", nil
+	}
+	return l.tables[table][key], nil
+}
+
+func (l *Local) Set(_ context.Context, table, key, value string) error {
+	l.tablesLock.Lock()
+	defer l.tablesLock.Unlock()
```
yevgenypats, line 105: I think if you pre-alloc the tables you can avoid locking the whole map and just locking on a per table.

yevgenypats (changes requested):
Looks really good. Don't have a lot of comments apart from some nits.

I do worry a bit on the global lock and Im curious to run some benchmarks with global and non global locks. If there is no difference then a global lock will be simple for sure if there is maybe we can change the api only to lock per table per client.

backend/backend.go
```diff
@@ -0,0 +1,12 @@
+package backend
+
+import "context"
+
+type Backend interface {
```
erezrokah, line 5: Should we make these table methods? So the API will be `t.SetCursor(value string)`?

That would make it more restrictive, and avoid potential collisions. Also what's the reason we need a key? Can a single table have more than 1 key?

erezrokah (approved):
Looks good, but maybe we should consider making the API more restrictive

codegen/templates/column.go.tpl
hermanschaaf: I think we may as well keep it here until the codegen gets removed (and then this entire file will get removed), don't see any harm in it

backend/backend.go
hermanschaaf: Yeah, a single table can have more than 1 key. For example, if the table is multiplexed on account or project, we may need to pass in different IDs for each of those accounts/projects to resume from the last position. Generally I think only very simple cases will have a single key, most cases will have many.

hermanschaaf: I like the table method idea! It might be hard to do in practice though. Right now resolvers don't have access to their own table, and even if they did, the table object wouldn't have a backend attached to it. I think we'll just have to accept that this will be the first pass, and we can iterate on the interface as we see how it's used in practice.

internal/backends/local/local.go
hermanschaaf: Maybe I'm misunderstanding, but I think we can't pre-alloc tables, because we don't know which tables will use this before `Set` is called. And if no tables use it, we don't want to write any file when the backend gets closed.

plugins/source/plugin.go
hermanschaaf: Looking into this now

hermanschaaf:
@yevgenypats Addressed all the comments and went through another iteration, this should be good for another look now.

backend/backend.go
erezrokah: >Yeah, a single table can have more than 1 key. For example, if the table is multiplexed on account or project, we may need to pass in different IDs for each of those accounts/projects to resume from the last position. Generally I think only very simple cases will have a single key, most cases will have many.

Multiplexed clients have an `ID()` method (I think tables can access it). Would that make a good key? I think that would make it easier to trace which cursor belongs to which client/table.

> I think we'll just have to accept that this will be the first pass

Good point. If we see the API is used in an inconsistent way we can change it (unless we want it that way :) )

hermanschaaf: Yup, I haven't changed it to be a method on Table (which might be hard), but I have updated the Backend interface to be like this:

```
Set(ctx context.Context, table, clientID, value string) error
```

So basically we're relying on the client id to give uniqueness (which should be true if you use multiplexers correctly)

yevgenypats (approved):
Looks great! I might have some suggestions on improving the implementation or performance but I think The external API looks good so it should easy to adjust as we go.

Prob for next feature / improvement around that area I'd like to work on it to get more hands-on experience with that feature to see if I've any additional ideas around that.

PR #570, opened 2023-01-04
feat: Support conversion of Unix timestamps in timestamptz

hermanschaaf:
This adds support for Unix timestamps to the Timestamptz type. I think it's reasonable to expect that an integer type is to be interpreted as unix time, and a custom resolver should be used if this is not the case.


disq:
What about UnixMillis? How can you be sure the given timestamp is unix (ie. seconds)? (Auto-detection might work)

candiduslynx (approved):

hermanschaaf:
> What about UnixMillis? How can you be sure the given timestamp is unix (ie. seconds)? (Auto-detection might work)

@disq Yeah, there are many variations (also unix time + milliseconds as float), but I think it's about providing a reasonable default in the SDK (which we can document). If an int field contains something other than Unix seconds, it will need to be converted first, but this should cover most cases without the need for additional code.

As for unix millis specifically, I don't think we can reliably tell which one is which (imagine the time is supposed to represent 1 second after 1 January 1970, for example).

yevgenypats (approved):

PR #573, opened 2023-01-04
feat: Add unwrap option to transformations

yevgenypats:
Follow-up on transformation. Adding unwrap option as needed in some cases where we create our own wrapper struct.

This also moves the tests from codegen.

### DIFF
--- a/transformers/struct.go
+++ b/transformers/struct.go
@@ -10,10 +10,12 @@ import (
 )
 
 type structTransformer struct {
-	table           *schema.Table
-	skipFields      []string
-	nameTransformer NameTransformer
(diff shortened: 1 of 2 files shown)


erezrokah (approved):
Looks good, maybe we can do another PR to add the `WithPrimaryKeys` option?

yevgenypats:
> Looks good, maybe we can do another PR to add the `WithPrimaryKeys` option?

I want to avoid it for now. The way to override in the paradaigm is just to enter it manually in the column otherwise we will end up `WithPKNewType`. `WithOverrideType` and so on...

PR #574, opened 2023-01-05
feat(testing): Add test for migrations

hermanschaaf:
This adds a test to the test suite specifically for migrations. It validates that:
 - new columns can be added
 - columns added by users get ignored
 - column ordering changes don't affect the outcome


erezrokah (approved):
Looks good, do we want to repeat the write/read test after each migration? For example if there's an extra column in the table, we expect it to have a null value after write

yevgenypats (commented):
🪄

yevgenypats (approved):

PR #581, opened 2023-01-06
fix(destination-testing): Make destination tests opt-in rather than opt-out

hermanschaaf:
I'd like to propose changing the destination tests to opt-in, rather than opt-out, so that we can add new tests without immediately needing to update all destination plugins to either make the test pass or skip the test. Instead, we can then gradually add the test case to plugins where it makes sense, and ignore it in other plugins.

Within each destination, this also makes it more explicit what tests will be run. E.g. this is what it would look like for BigQuery:

```
destination.PluginTestSuiteTests{
	TestOverwrite:     false,
	TestDeleteStale:   true,
	TestAppend:        true,
	TestSecondAppend:  true,
	TestMigrateAppend: true,
})
```

This is a breaking change for the SDK interface, but since it doesn't affect the gPRC interface this won't be a major bump.

### DIFF
--- a/plugins/destination/plugin_testing.go
+++ b/plugins/destination/plugin_testing.go
@@ -20,27 +20,27 @@ type PluginTestSuite struct {
 }
 
 type PluginTestSuiteTests struct {
-	// SkipOverwrite skips testing for "overwrite" mode. Use if the destination
-	//	// plugin doesn't support this feature.
-	SkipOverwrite bool
+	// TestOverwrite adds testing for "overwrite" mode. Use if the destination
+	// plugin supports this feature.
+	TestOverwrite bool
 
-	// SkipDeleteStale skips testing "delete-stale" mode. Use if the destination
-	// plugin doesn't support this feature.
-	SkipDeleteStale bool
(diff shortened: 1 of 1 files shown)


disq (approved):

yevgenypats (changes requested):
I actually changed it the otherway around exactly for that reason. let's say we add a new test because we found a bug then new destination will automatically get it and fail if they don't support it which they should because otherwise they will just upgrade the SDK silently.

hermanschaaf:
@yevgenypats That would be fine if we expected all destinations to pass all tests, but as things stand they each have a different mix of what they support (append / second append / stale / migrations / overwrite), and this will probably always be the case. So when we introduce a new test now, we need to immediately go update a large percentage of plugins when we upgrade the SDK version, otherwise CI fails. I'd rather have it be the other way round, so adding new tests to the SDK doesn't become a blocker for other features.

erezrokah:
Maybe we should have a set of tests we expect to pass on all plugins, and another opt-in? Or we could make it so the SDK only adds utility methods for tests (creating data, adding data, validating that it exists, etc.).

PR #582, opened 2023-01-06
feat: Add batch size bytes as additional option

hermanschaaf:
This adds `batchSizeBytes` as an additional option to limit batches not only by number of records, but also by number of bytes.

### DIFF
--- a/plugins/destination/managed_writer.go
+++ b/plugins/destination/managed_writer.go
@@ -23,20 +23,19 @@ func (p *Plugin) worker(ctx context.Context, metrics *Metrics, table *schema.Tab
 	for {
 		select {
 		case r, ok := <-ch:
-			if ok {
-				if len(resources) == p.spec.BatchSize || sizeBytes+r.Size() > p.spec.BatchSizeBytes {
-					p.flush(ctx, metrics, table, resources)
-					resources = make([][]any, 0)
-					sizeBytes = 0
-				}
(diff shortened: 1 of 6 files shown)


plugins/destination/managed_writer.go
```diff
@@ -19,68 +19,50 @@ type worker struct {
...
-						atomic.AddUint64(&metrics.Errors, uint64(len(resources)))
-					} else {
-						p.logger.Info().Str("table", table.Name).Int("len", len(resources)).Dur("duration", time.Since(start)).Msg("last batch written successfully")
-						atomic.AddUint64(&metrics.Writes, uint64(len(resources)))
-					}
+					p.flush(ctx, metrics, table, resources)
 				}
 				return
 			}
+			if len(resources) == p.spec.BatchSize || sizeBytes+r.Size() > p.spec.BatchSizeBytes {
```
disq, line 32: Using zero as the default value shortcut breaks my heart every time. Maybe allow negative values (erm, `-1`) to mean "unlimited"?

yevgenypats (approved):
LGTM!

plugins/destination/managed_writer.go
hermanschaaf: Sorry, not sure I understood the comment... Was it meant to be a comment on this line? If the batch size limit is 0, the SetDefaults call will ensure it's updated to the default batch size. So there will always be some limit, but you can of course set it to 1PB if you don't want it to affect anything in practice

PR #583, opened 2023-01-07
fix(metrics): Fix metrics propagation for destination plugins with managed write mode

candiduslynx:


erezrokah (changes requested):
Thanks for the PR @candiduslynx, can we separate the refactor from the bug fix?
It's hard to understand the fix itself.

Also I think with this fix, we're losing the granularity of the error messages (e.g. which flush failed)

candiduslynx:
refactor part was similarly implemented in #582.

PR #585, opened 2023-01-09
fix(metrics): Fix metrics propagation for destination plugins with managed write mode

candiduslynx:
Refactored destination plugin metrics:
1. incorporated metrics as well as table info into workers
2. properly collecting metrics from workers for managed writers
3. left only a single mutex in the destination plugin
4. refactored map handling a bit

### DIFF
--- a/plugins/destination/plugin.go
+++ b/plugins/destination/plugin.go
@@ -3,6 +3,7 @@ package destination
 import (
 	"context"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/cloudquery/plugin-sdk/schema"
@@ -55,10 +56,14 @@ type Client interface {
 	ReverseTransformValues(table *schema.Table, values []any) (schema.CQTypes, error)
 	Migrate(ctx context.Context, tables schema.Tables) error
 	Read(ctx context.Context, table *schema.Table, sourceName string, res chan<- []any) error
-	ManagedWriter
-	UnmanagedWriter
 	DeleteStale(ctx context.Context, tables schema.Tables, sourceName string, syncTime time.Time) error
 	Close(ctx context.Context) error
+
+	// ManagedWriter interface should be implemented if you use WithManagedWriter option.
+	ManagedWriter
+
+	// UnmanagedWriter interface should be implemented if you don't use WithManagedWriter option.
+	UnmanagedWriter
 }
 
 type ClientResource struct {
@@ -83,12 +88,12 @@ type Plugin struct {
 	// Logger to call, this logger is passed to the serve.Serve Client, if not define Serve will create one instead.
 	logger zerolog.Logger
 
-	// This is in use if the user passed a managed client
-	metrics     map[string]*Metrics
-	metricsLock *sync.RWMutex
-
 	workers     map[string]*worker
-	workersLock *sync.Mutex
+	workersLock sync.RWMutex // implies usage by pointer
+
+	// the errors and writes imply usage by pointer only
+	errors atomic.Uint64
+	writes atomic.Uint64
 
 	batchTimeout          time.Duration
 	defaultBatchSize      int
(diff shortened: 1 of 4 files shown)


candiduslynx:
this PR is following in the steps of #583

plugins/destination/plugin.go
```diff
@@ -157,14 +150,13 @@ func (p *Plugin) Metrics() Metrics {
 	case unmanaged:
 		return p.client.Metrics()
 	case managed:
-		metrics := Metrics{}
-		p.metricsLock.RLock()
-		for _, m := range p.metrics {
-			metrics.Errors += m.Errors
-			metrics.Writes += m.Writes
+		metrics := new(Metrics)
+		p.RLock()
```
hermanschaaf, line 154: Shouldn't this be a write lock, because we're updating the metric values?

plugins/destination/plugin.go
```diff
@@ -83,12 +83,8 @@ type Plugin struct {
 	// Logger to call, this logger is passed to the serve.Serve Client, if not define Serve will create one instead.
 	logger zerolog.Logger
 
-	// This is in use if the user passed a managed client
-	metrics     map[string]*Metrics
-	metricsLock *sync.RWMutex
-
-	workers     map[string]*worker
-	workersLock *sync.Mutex
+	sync.RWMutex // implies usage by pointer
```
hermanschaaf, line 86: Could we rather not use an embedding here? It's arguably nicer to read `p.Lock()`, but it makes it harder for a reader to figure out where the `[R]Lock`, `[R]Unlock` functions are defined, and it doesn't tell them what the lock is supposed to be for. I believe it also now makes `Lock` / `Unlock` public functions on `Plugin`; probably not something we want as a public API.

hermanschaaf (changes requested):
Looks good, just left a few comments. It would also be good to add a test that ensures the metrics behavior stays fixed.

plugins/destination/plugin.go
candiduslynx: No, as we're only reading them. The only one we're updating is the local `metrics` variable

plugins/destination/plugin.go
candiduslynx: I'll move it to `workersMux` then

plugins/destination/plugin.go
hermanschaaf: Ah okay, I see, thanks!

hermanschaaf (approved):

yevgenypats (changes requested):
Quite a. big change so just reviewed partially.

plugins/destination/metrics.go
```diff
@@ -1,8 +1,27 @@
...
 }
+
+func (mx *Metrics) Get() Metrics {
+	return Metrics{
+		Errors: atomic.LoadUint64(&mx.Errors),
+		Writes: atomic.LoadUint64(&mx.Writes),
+	}
+}
+
+func (mx *Metrics) Failed(amount int) {
```
yevgenypats, line 21: seems strange why `errors` -> `failed` and `writes` -> `Wrote`. aslo what is the value of having those operations instead of just returning the struct. I don't see any race possible here in the copy

candiduslynx: I wanted to expose atomic methods for the public fields, didn't find better names.
As for the atomic for the load part: same logic that was used for atomic store applies.

yevgenypats:
Prob worth a live chat as I don't think I quite understand what it solves

candiduslynx:
@yevgenypats I've updated the metrics code. Now we use [sync.Uint64](https://pkg.go.dev/sync/atomic#Uint64) to store locally & return the `destination.Metrics` as read value.

candiduslynx:
related to https://github.com/cloudquery/cloudquery/issues/6761

yevgenypats:
Closing as I didn't understand the fix and I'll prob just fix the metrics later.

PR #587, opened 2023-01-09
fix: Array types

disq:
untested


codegen/transformers.go
```diff
@@ -54,11 +54,24 @@ func defaultGoTypeToSchemaType(v reflect.Type) (schema.ValueType, error) {
 		}
 		return schema.TypeJSON, nil
 	case reflect.Slice:
+		switch v.Elem().Kind() {
+		case reflect.Uint8:
```
erezrokah, line 58: Shouldn't this be in `defaultGoTypeToSchemaType`?

candiduslynx: `defaultGoTypeToSchemaType` will return `TypeInt` in this case, effectively resulting in `TypeIntArray` returned

disq: there's no `schema.TypeByte`, so no.

disq:
> Can we add a regression test for that?

@yevgenypats added

PR #590, opened 2023-01-09
feat: Add mockgen package for generating service interfaces

hermanschaaf:
This adds a new package to the SDK, `mockgen`, that generates interfaces for SDK clients for use with mockgen. 

Here is an example of it now being used in the Fastly plugin: https://github.com/cloudquery/cloudquery/pull/6546

Most of this isn't really new code: it's just being moved from the AWS, Fastly, Slack and other plugins where service interface generation is/was used. Once moved into the SDK, it could also be used by plugins like GitHub and DigitalOcean.


erezrokah (approved):

yevgenypats (commented):
Question: Do we want to move this to a seperate `codegen` package as not this doesn't really use or relate to the SDK? We can then also move https://github.com/cloudquery/cloudquery/tree/main/plugins/source/azure/codegen0 to the same package as utilities for code travesal (can be useful for oracle plugin for example)

yevgenypats:
I think let's do it as looks like codegen helpers still exist like this one, and the azure one which can potentially be used for stripe and oracle. this will enable avoiding updating the SDK for repositories that don't need the codegen helpers.

disq:
Shouldn't this be called `mockgengen` (or `mockgen0`?)

hermanschaaf:
> Shouldn't this be called mockgengen (or mockgen0?)

@disq Maybe, but can we please not call it either of those? :)

PR #594, opened 2023-01-10
fix: Incremental tables should not delete stale

hermanschaaf:
When `overwrite-delete-stale` write mode is used, the delete stale should exclude incremental tables. We still want `delete-stale` behavior for non-incremental tables, but incremental tables have the expectation that previously synced rows should remain there. 

This also fixes a small bug where we were deleting stale entries twice for managed tables.

### DIFF
--- a/plugins/destination/plugin.go
+++ b/plugins/destination/plugin.go
@@ -254,7 +254,13 @@ func (p *Plugin) Write(ctx context.Context, tables schema.Tables, sourceName str
 		panic("unknown client type")
 	}
 	if p.spec.WriteMode == specs.WriteModeOverwriteDeleteStale {
-		if err := p.DeleteStale(ctx, tables, sourceName, syncTime); err != nil {
+		include := func(t *schema.Table) bool { return true }
+		exclude := func(t *schema.Table) bool { return t.IsIncremental }
+		nonIncrementalTables, err := tables.FilterDfsFunc(include, exclude)
+		if err != nil {
+			return err
+		}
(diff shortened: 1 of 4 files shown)


disq (approved):

plugins/destination/plugin.go
```diff
@@ -254,7 +254,13 @@ func (p *Plugin) Write(ctx context.Context, tables schema.Tables, sourceName str
 		panic("unknown client type")
 	}
 	if p.spec.WriteMode == specs.WriteModeOverwriteDeleteStale {
-		if err := p.DeleteStale(ctx, tables, sourceName, syncTime); err != nil {
+		include := func(t *schema.Table) bool { return true }
+		exclude := func(t *schema.Table) bool { return t.IsIncremental }
+		nonIncrementalTables, err := tables.FilterDfsFunc(include, exclude)
```
erezrokah, line 259: Are `tables` already filtered by the glob patterns in the spec?
Making sure we don't return all tables here since we include everything.

Also since `include` is always `true`, maybe add only the exclude part for now?

erezrokah (approved):
Looks good, added one comment for clarification

plugins/destination/plugin.go
hermanschaaf: Right, `tables` are already filtered by glob patterns at this point (previously we were passing all these tables in to delete stale, now we're filtering out the incremental ones from that list first). 

`FilterDfsFunc` is now also used by `FilterDfs` internally, and that needs both `include` and `exclude`.

yevgenypats (approved):
🪄

PR #596, opened 2023-01-10
fix(types-timestamptz): Handle `MarshalText` and `String` for pointer receivers

erezrokah:
#### Summary

We need to also try calling `MarshalText` and `String` when those methods are declared on a pointer receiver

---

- [ ] Update or add tests 🧪


yevgenypats:
@erezrokah can you link to where this is used/current failing in the plugins ?

erezrokah:
I thought I needed it for GCP, but actually the fix there was to override the resolver. I still think it makes sense as it's common to declare methods on pointer receivers.

I don't think it's super important at the moment so we can close this

PR #598, opened 2023-01-11
feat(transformers): Add WithPrimaryKeys option

erezrokah:
#### Summary

Adds `WithPrimaryKeys` so we don't need to define extra columns just for the sake of primary keys.
The issue with extra columns is that you need to manually configure the column name resolver and type. It's easy to miss those especially when plugins have custom logic for resolvers and types, for example:
https://github.com/cloudquery/cloudquery/blob/cb9b4f85ccb82f850e6ded6c19a39d06c4f2c94a/plugins/source/gcp/client/transformers.go#L17
https://github.com/cloudquery/cloudquery/blob/cb9b4f85ccb82f850e6ded6c19a39d06c4f2c94a/plugins/source/gcp/client/transformers.go#L39
https://github.com/cloudquery/cloudquery/blob/cb9b4f85ccb82f850e6ded6c19a39d06c4f2c94a/plugins/source/gcp/client/transformers.go#L54

---

- [ ] Update or add tests 🧪

### DIFF
--- a/transformers/struct.go
+++ b/transformers/struct.go
@@ -20,6 +20,7 @@ type structTransformer struct {
 	ignoreInTestsTransformer      IgnoreInTestsTransformer
 	unwrapAllEmbeddedStructFields bool
 	structFieldsToUnwrap          []string
+	pkFields                      []string
 }
 
 type NameTransformer func(reflect.StructField) (string, error)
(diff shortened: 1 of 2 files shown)


yevgenypats (approved):
Nice. cc @disq (I know you were also waiting for this) :)

PR #599, opened 2023-01-11
feat(docs): Sort tables

erezrokah:
#### Summary

Our docs show tables based on definition order which can be inconsistent and makes it harder to locate a table, see https://www.cloudquery.io/docs/plugins/sources/azure/tables:
image

This PR sorts the tables and relations before generating the docs

---

- [ ] Update or add tests 🧪

### DIFF
--- a/plugins/source/docs.go
+++ b/plugins/source/docs.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"sort"
 	"strings"
 	"text/template"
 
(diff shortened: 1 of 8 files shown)


plugins/source/docs.go
```diff
@@ -21,13 +22,24 @@ var templatesFS embed.FS
...
+}
+
 // GeneratePluginDocs creates table documentation for the source plugin based on its list of tables
 func (p *Plugin) GeneratePluginDocs(dir, format string) error {
 	if err := os.MkdirAll(dir, os.ModePerm); err != nil {
 		return err
 	}
 
 	destination.SetDestinationManagedCqColumns(p.Tables())
+	sortTables(p.Tables())
```
hermanschaaf, line 42: I think it's a bit unexpected that the `GeneratePluginDocs` function changes the ordering of the tables for the plugin. Maybe `p.Tables()` should rather always return a sorted list?

disq (approved):

plugins/source/docs.go
hermanschaaf: Either that, or we should make a copy of the tables slice, sort it, and pass it in to the render functions

erezrokah: > Maybe p.Tables() should rather always return a sorted list?

I think that would be a bigger change as it could impact our scheduling (if we rely on that order to decide which tables go first).

A copy works 👍

hermanschaaf (approved):

yevgenypats (approved):

PR #601, opened 2023-01-11
fix: Send resource validation errors to Sentry

hermanschaaf:
Resource validation errors, like a row missing a PK value, are non-recoverable and need to be fixed by the plugin author. As such, we should know about them when they happen, even though they do not cause a panic in the code.

### DIFF
--- a/plugins/source/scheduler_dfs.go
+++ b/plugins/source/scheduler_dfs.go
@@ -161,6 +161,10 @@ func (p *Plugin) resolveResourcesDfs(ctx context.Context, table *schema.Table, c
 				if err := resolvedResource.Validate(); err != nil {
 					tableMetrics := p.metrics.TableClient[table.Name][client.ID()]
 					p.logger.Error().Err(err).Str("table", table.Name).Str("client", client.ID()).Msg("resource resolver finished with validation error")
+					sentry.WithScope(func(scope *sentry.Scope) {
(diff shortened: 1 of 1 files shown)


erezrokah (approved):

plugins/source/scheduler_dfs.go
```diff
@@ -161,6 +162,14 @@ func (p *Plugin) resolveResourcesDfs(ctx context.Context, table *schema.Table, c
 				if err := resolvedResource.Validate(); err != nil {
 					tableMetrics := p.metrics.TableClient[table.Name][client.ID()]
 					p.logger.Error().Err(err).Str("table", table.Name).Str("client", client.ID()).Msg("resource resolver finished with validation error")
+					if _, found := sentValidationErrors.LoadOrStore(table.Name, struct{}{}); !found {
```
yevgenypats, line 165: do we we want only to send once per table? I guess it makes sense even though we can miss some things but yeah let's start with that.

yevgenypats (approved):

plugins/source/scheduler_dfs.go
hermanschaaf: Yeah, we may miss some things, but I hope not. And we'll catch it once the first issue is fixed anyway. I figure it's better than using our entire Sentry quota in one sync (unless I'm misunderstanding how Sentry billing works)

PR #603, opened 2023-01-11
fix: Change options for new client

yevgenypats:
I think I missed it in the review but the api should include an option struct rather then the `...option` as otherwise it's just introduce un-necessary boilerplate for each client.

Also, I've added BackendNone as otherwise there is no way to specify that you don't want incremental syncing.


yevgenypats:
Now that Im thinking more about it I think we might have a problem in our incremental syncing approach.

If we specify incremental, only incremental tables should be synced as otherwise it can cause issues with destinations such as gcs.

also, on the destination side we always don't delete incremental tables which basically now we disabled an option of `overwrite-delete-stale` for a lot of tables (i.e let's say we just want `overwrite-delete-stale` without any incremental syncing).

hermanschaaf (approved):

hermanschaaf:
> If we specify incremental, only incremental tables should be synced as otherwise it can cause issues with destinations such as gcs.

@yevgenypats what kind of issues are you thinking about here? Maybe let's sync up on this if you have time later :)

disq:
@yevgenypats backend could've been `nil` before (not specifying the option would make it `nil`) so it worked for non-incrementals.

disq:
I don't think incremental-only should be automatically decided based on backend. If the user wants to sync incrementals on short periods but don't spam the full resources, they can have separate configs and use `skip_tables`. (Could also add an `only_incremental bool` option for convenience)

specs/incremental.go
```diff
@@ -0,0 +1,49 @@
...
+func (r *Incremental) UnmarshalJSON(data []byte) (err error) {
+	var registry string
+	if err := json.Unmarshal(data, &registry); err != nil {
+		return err
+	}
+	if *r, err = IncrementalFromString(registry); err != nil {
+		return err
+	}
+	return nil
+}
```
hermanschaaf, line 36:
```suggestion
func (r *Incremental) UnmarshalJSON(data []byte) (err error) {
	var incremental string
	if err := json.Unmarshal(data, &incremental); err != nil {
		return err
	}
	if *r, err = IncrementalFromString(incremental); err != nil {
		return err
	}
	return nil
}
```

specs/incremental.go
```diff
@@ -0,0 +1,49 @@
...
+func IncrementalFromString(s string) (Incremental, error) {
+	switch s {
+	case "none":
+		return IncrementalNone, nil
+	case "only":
+		return IncrementalTablesOnly, nil
+	case "both":
+		return IncrementalBoth, nil
+	default:
+		return IncrementalNone, fmt.Errorf("unknown registry %s", s)
```
hermanschaaf, line 47:
```suggestion
		return IncrementalNone, fmt.Errorf("unknown incremental value %s", s)
```

specs/incremental.go
```diff
@@ -0,0 +1,49 @@
...
+	"encoding/json"
+	"fmt"
+)
+
+type Incremental int
+
+const (
+	IncrementalNone Incremental = iota
+	IncrementalTablesOnly
+	IncrementalBoth
```
hermanschaaf, line 14: What is the behavior expected to be for all these settings? What is the difference between setting `IncrementalTablesOnly` here and using the `OnlyIncrementalTables` boolean at the source spec level?

hermanschaaf (commented):
Maybe let's catch up about the new changes? I think we should consider some alternatives

specs/source.go
```diff
@@ -39,6 +39,10 @@ type Source struct {
 
 	// Backend is the name of the state backend to use
 	Backend Backend `json:"backend,omitempty"`
+	// Enable incremental syncing for tables that supports that
+	Incremental Incremental `json:"incremental,omitempty"`
+	// Sync on
+	OnlyIncrementalTables bool `json:"only_incremental_tables,omitempty"`
```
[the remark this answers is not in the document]
disq: It would be hard to determine which tables are incremental and generate a config and keep it up to date. This way you can have one config with all tables, fetched daily, and another one, fetched hourly, updating only incrementals (taking shorter bursts, not wasting API limits and other resources)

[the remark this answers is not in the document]
hermanschaaf: True, but I think keeping your list of tables up-to-date (which to include, which to skip) and how regularly you want to sync each is a problem separate from incremental-ness, adding this one boolean is not going to solve it except maybe for a narrow use case

yevgenypats:
Removed Incremental per discussion and using only `backend` config.

internal/servers/destinations.go
```diff
@@ -76,12 +76,26 @@ func (s *DestinationServer) Write2(msg pb.Destination_Write2Server) error {
...
+	} else {
+		if err := json.Unmarshal(r.SourceSpec, &sourceSpec); err != nil {
+			return status.Errorf(codes.InvalidArgument, "failed to unmarshal source spec: %v", err)
+		}
+	}
 	sourceName := r.Source
 	syncTime := r.Timestamp.AsTime()
+	sourceSpec = specs.Source{
+		Name: sourceName,
+	}
```
hermanschaaf, line 94: Why are we overwriting `sourceSpec`, shortly after unmarshaling from `r.SourceSpec`?

hermanschaaf (approved):
Looks good, just had the one question about overwriting the variable shortly after unmarshalling it

internal/servers/destinations.go
yevgenypats: good catch. a copy paste from the previous if

PR #604, opened 2023-01-12
feat(destinations): Add `migrate_mode`

erezrokah:
#### Summary

Related to https://github.com/cloudquery/cloudquery/issues/6600, needed for https://github.com/cloudquery/cloudquery/pull/6763

---

- [ ] Update or add tests 🧪

### DIFF
--- a/specs/destination.go
+++ b/specs/destination.go
@@ -9,10 +9,6 @@ import (
 	"github.com/thoas/go-funk"
 )
 
-type WriteMode int
-
-type MigrateMode int
-
 type Destination struct {
 	Name           string      `json:"name,omitempty"`
 	Version        string      `json:"version,omitempty"`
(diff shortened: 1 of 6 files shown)


specs/destination.go
```diff
@@ -11,15 +11,18 @@ import (
...
-	WriteMode      WriteMode `json:"write_mode,omitempty"`
-	BatchSize      int       `json:"batch_size,omitempty"`
-	BatchSizeBytes int       `json:"batch_size_bytes,omitempty"`
-	Spec           any       `json:"spec,omitempty"`
+	Name           string      `json:"name,omitempty"`
+	Version        string      `json:"version,omitempty"`
+	Path           string      `json:"path,omitempty"`
+	Registry       Registry    `json:"registry,omitempty"`
+	WriteMode      WriteMode   `json:"write_mode,omitempty"`
+	MigrateMode    MigrateMode `json:"migrate_mode,omitempty"`
```
erezrokah, line 22: This can also be a boolean, just found it easier to communicate the purpose using strings

hermanschaaf (approved):
This LGTM; are we also planning on supporting a `--migrate forced` (or similar) flag for the CLI? Might make it a bit easier for users doing things interactively

hermanschaaf:
We're also not using the setting anywhere yet - is that because it needs to be implemented in each destination? I think we'll need to pass the mode in to the `Migrate` call so that destinations can act on it?

disq (approved):

erezrokah:
> a `--migrate forced` (or similar) flag for the CLI? Might make it a bit easier for users doing things interactively

So a CLI will we have to "take" for all specs, where a spec config only "takes" on the specific spec. We can consider adding it but I don't think we can similar flags (that override spec configuration(s))

yevgenypats (changes requested):
Looks good. One nit:

Can we move migrate enum code to `migrate.go` as looks like our enums are growing and maybe also add a small sanity test for marshal/unmarsahl.

erezrokah:
> Can we move migrate enum code to `migrate.go` as looks like our enums are growing and maybe also add a small sanity test for marshal/unmarsahl.

Sure, good point. Let me know if https://github.com/cloudquery/plugin-sdk/pull/604/commits/1ff6062f3821a20e6a7e67f1aee6e026288b2cbb is what you had in mind

yevgenypats (approved):

PR #609, opened 2023-01-16
feat(spec): Add source, destination String methods

erezrokah:
#### Summary

A user recently ran into some issues since they thought they used a newer version of a plugin, but actually used an old one.
This PR adds `String()` methods to specs so we can use them in logs/errors.

CLI PR https://github.com/cloudquery/cloudquery/pull/6842

---

- [ ] Update or add tests 🧪

### DIFF
--- a/specs/destination.go
+++ b/specs/destination.go
@@ -71,3 +71,11 @@ func (d *Destination) Validate() error {
 	}
 	return nil
 }
+
+func (d Destination) String() string {
+	pathParts := strings.Split(d.Path, "/")
+	if d.Name == pathParts[1] {
+		return fmt.Sprintf("%s (%s)", d.Name, d.Version)
+	}
+	return fmt.Sprintf("%s (%s@%s)", d.Name, pathParts[1], d.Version)
+}
--- a/specs/destination.go
+++ b/specs/destination.go
@@ -73,7 +73,13 @@ func (d *Destination) Validate() error {
 }
 
 func (d Destination) String() string {
+	if d.Registry != RegistryGithub {
+		return fmt.Sprintf("%s (%s@%s)", d.Name, d.Registry, d.Path)
+	}
 	pathParts := strings.Split(d.Path, "/")
+	if len(pathParts) != 2 {
+		return fmt.Sprintf("%s (%s@%s)", d.Name, d.Path, d.Version)
+	}
 	if d.Name == pathParts[1] {
 		return fmt.Sprintf("%s (%s)", d.Name, d.Version)
 	}
--- a/specs/destination.go
+++ b/specs/destination.go
@@ -72,7 +72,7 @@ func (d *Destination) Validate() error {
 	return nil
 }
 
-func (d Destination) String() string {
+func (d Destination) VersionString() string {
 	if d.Registry != RegistryGithub {
 		return fmt.Sprintf("%s (%s@%s)", d.Name, d.Registry, d.Path)
 	}
--- a/specs/source.go
+++ b/specs/source.go
@@ -129,7 +129,7 @@ func (s *Source) Validate() error {
 	return nil
 }
 
-func (s Source) String() string {
+func (s Source) VersionString() string {
 	if s.Registry != RegistryGithub {
 		return fmt.Sprintf("%s (%s@%s)", s.Name, s.Registry, s.Path)
 	}
--- a/specs/source.go
+++ b/specs/source.go
@@ -130,7 +130,13 @@ func (s *Source) Validate() error {
 }
 
 func (s Source) String() string {
+	if s.Registry != RegistryGithub {
+		return fmt.Sprintf("%s (%s@%s)", s.Name, s.Registry, s.Path)
+	}
 	pathParts := strings.Split(s.Path, "/")
+	if len(pathParts) != 2 {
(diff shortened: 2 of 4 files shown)


specs/source.go
```diff
@@ -128,3 +128,11 @@ func (s *Source) Validate() error {
 	}
 	return nil
 }
+
+func (s Source) String() string {
+	pathParts := strings.Split(s.Path, "/")
```
erezrokah, line 133: We can drop this bit for simplicity too (but we'll get `aws (aws@v1.0.0)` sometimes).
I'm also removing the org part of the path to make this shorter

specs/destination.go
```diff
@@ -71,3 +71,11 @@ func (d *Destination) Validate() error {
 	}
 	return nil
 }
+
+func (d Destination) String() string {
+	pathParts := strings.Split(d.Path, "/")
+	if d.Name == pathParts[1] {
```
disq, line 77: Do we validate `d.Path` contains at least one `/` beforehand? otherwise we'll get an index out of range panic here.

specs/source.go
disq: Same here, should check for `len(pathParts)` before going for `pathParts[1]` if it's not verified elsewhere beforehand

specs/destination.go
erezrokah: Nice catch! We do, but not soon enough for some log prints, I added https://github.com/cloudquery/plugin-sdk/pull/609/commits/83d728c697195291cb84d2053c5ac7581ab9879c to also handle non GitHub registries

disq (approved):

yevgenypats (changes requested):
`String` is a saved method in Go and usually means that it stringify the whole object. so let's say if we use `fmt.Printf(spec)` this will now print the version instead of the string.

Can we just use `VersionString` method for this?

erezrokah:
> `String` is a saved method in Go and usually means that it stringify the whole object. so let's say if we use `fmt.Printf(spec)` this will now print the version instead of the string.

Isn't that what overriding `String()` is for? To provide your own representation of how a struct will be printed?
It doesn't print only the version, it prints the `name + version` or `name + path + version`.

> Can we just use VersionString method for this?

Using `String()` is useful for that exact reason. Formatting functions will use that implicitly. Also our logger has a `Stringer` function:
https://github.com/cloudquery/cloudquery/pull/6842/files#diff-ca3888a7ad618b67068e6b20ecb264fe87bb19554e6e8b62a13f8b5ce540f431R24

We can have a dedicated method, but that means we have to call it before passing the spec to any method that does formatting. That's especially clunky for arrays as you'll need to iterate over the array and call the method per item

yevgenypats:
> > `String` is a saved method in Go and usually means that it stringify the whole object. so let's say if we use `fmt.Printf(spec)` this will now print the version instead of the string.
> 
> Isn't that what overriding `String()` is for? To provide your own representation of how a struct will be printed? It doesn't print only the version, it prints the `name + version` or `name + path + version`.
> 
> > Can we just use VersionString method for this?
> 
> Using `String()` is useful for that exact reason. Formatting functions will use that implicitly. Also our logger has a `Stringer` function: https://github.com/cloudquery/cloudquery/pull/6842/files#diff-ca3888a7ad618b67068e6b20ecb264fe87bb19554e6e8b62a13f8b5ce540f431R24
> 
> We can have a dedicated method, but that means we have to call it before passing the spec to any method that does formatting. That's especially clunky for arrays as you'll need to iterate over the array and call the method per item



> > `String` is a saved method in Go and usually means that it stringify the whole object. so let's say if we use `fmt.Printf(spec)` this will now print the version instead of the string.
> 
> Isn't that what overriding `String()` is for? To provide your own representation of how a struct will be printed? It doesn't print only the version, it prints the `name + version` or `name + path + version`.
> 
> > Can we just use VersionString method for this?
> 
> Using `String()` is useful for that exact reason. Formatting functions will use that implicitly. Also our logger has a `Stringer` function: https://github.com/cloudquery/cloudquery/pull/6842/files#diff-ca3888a7ad618b67068e6b20ecb264fe87bb19554e6e8b62a13f8b5ce540f431R24
> 
> We can have a dedicated method, but that means we have to call it before passing the spec to any method that does formatting. That's especially clunky for arrays as you'll need to iterate over the array and call the method per item

I don't think this is what we want or at least for me it was quite confusing as I would assume `String` function will return some kind of `json` or `yaml` representation of the whole spec and not the `pat@version`. I would prefer to have an explicit `Version` or `PathVersionString` then something explicit that we will have hard time changing in case we do want `String` and then we will have to have a function name `FullString` (so prefer it the other way around)

erezrokah:
> I don't think this is what we want or at least for me it was quite confusing as I would assume `String` function will return some kind of `json` or `yaml` representation of the whole spec and not the `pat@version`. I would prefer to have an explicit `Version` or `PathVersionString` then something implicit that we will have hard time changing in case we do want `String` and then we will have to have a function name `FullString` (so prefer it the other way around)

Cool, got it. I'll make the function explicit

erezrokah:
Done in https://github.com/cloudquery/plugin-sdk/pull/609/commits/734f49c14ba824a16b8110c1f93eb2492828cb78 cc @yevgenypats

yevgenypats (approved):

PR #610, opened 2023-01-16
feat: Dynamic tables and introduce proto versioning

yevgenypats:
This introduce WithDynamicTables for source plugins to give an option for plugins to generate tables on the go, this can be useful in GA analytics, Salesforce and others where you can only get the schema and tables in runtime.

Give WithDynamicTables is actually a breaking change and I didn't want to turn our code into spagetti that tries to understand in runtime which version is the client I took the opportunity to fix our protobuf structures and introduce a way for creating new versions for server in a breaking way while also serving older versions of the server so it wont affect user experience but will give us a better way to write and maintain different versions and eventually deprecate older versions.

cc @erezrokah as I know you suggested exactly that a while back but it was too early for me after the initial refactor :) and wasn't sure what are the requirements yet but here we go!


yevgenypats:
> Looks great! I had a look through the code. @yevgenypats Have you been able to use it successfully for a source plugin? It would be nice to see an example of it in use to better judge the interface and make sure it can be used to do what we need, before committing to it :)

Yes, definitely. What I'd like to do is the following:
1) to merge this after initial review
2) Put the SDK on freeze (not push any release and just keep this commit in main).
3) open a CLI PR and open any small PRs on top if any changes to v1 so it will be easier to review.

internal/servers/source/v0/source.go
```diff
@@ -101,15 +109,23 @@ func (s *SourceServer) Sync2(req *pb.Sync2_Request, stream pb.Source_Sync2Server
 	if err := dec.Decode(&spec); err != nil {
 		return status.Errorf(codes.InvalidArgument, "failed to decode spec: %v", err)
 	}
+	ctx := stream.Context()
+	if err := s.Plugin.Init(ctx, spec); err != nil {
```
erezrokah, line 113: Was this missing or moved from someplace else?

plugins/source/plugin.go
```diff
@@ -164,58 +195,107 @@ func (p *Plugin) Metrics() *Metrics {
 	return p.metrics
 }
 
-// Sync is syncing data from the requested tables in spec to the given channel
-func (p *Plugin) Sync(ctx context.Context, spec specs.Source, res chan<- *schema.Resource) error {
+func (p *Plugin) Init(ctx context.Context, spec specs.Source) error {
```
erezrokah, line 198: Ah, got it we need a separate `Init` to sync dynamic tables

erezrokah (approved):
Looks good but it's quite a big diff to review.
To confirm, other than the new package structure the changes are:
1. Add the new option
2. Sync the dynamic tables during init
3. Continue with sync based on dynamic tables

Is that mostly correct?

yevgenypats:
> Looks good but it's quite a big diff to review. To confirm, other than the new package structure the changes are:
> 
> 1. Add the new option
> 2. Sync the dynamic tables during init
> 3. Continue with sync based on dynamic tables
> 
> Is that mostly correct?
correct.

I can split to PRs if that's helpful with one only introducing v0 moving things around and then one introducing v1.

erezrokah:
> I can split to PRs if that's helpful with one only introducing v0 moving things around and then one introducing v1.

No need to split, just wanted to confirm

hermanschaaf (approved):
LGTM, not sure why the build is failing on the post-step

internal/servers/source/v0/source.go
yevgenypats: Yeah now we need to call it in `v0` as we don't want to break the v0 protocol

PR #611, opened 2023-01-17
fix(clients): Update `log line too long` message

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/5665, by making this log message look less like an error.
I could go with `Warn` too and maybe update the level of the other errors here (but won't do it unless someone finds those confusing).

The current message can cause users to think that there is data loss or something wrong with the sync

---

- [ ] Update or add tests 🧪

### DIFF
--- a/clients/destination/v0/destination.go
+++ b/clients/destination/v0/destination.go
@@ -153,16 +153,16 @@ func (c *Client) newManagedClient(ctx context.Context, path string) error {
 				break
 			}
 			if errors.Is(err, logging.ErrLogLineToLong) {
-				c.logger.Err(err).Str("line", string(line)).Msg("skipping too long log line")
+				c.logger.Info().Str("line", string(line)).Msg("truncated destination plugin log line")
 				continue
 			}
 			if err != nil {
-				c.logger.Err(err).Msg("failed to read log line from plugin")
+				c.logger.Err(err).Msg("failed to read log line from destination plugin")
 				break
 			}
 			var structuredLogLine map[string]any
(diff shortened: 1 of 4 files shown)


hermanschaaf (approved):

clients/source.go
```diff
@@ -146,16 +146,16 @@ func (c *SourceClient) newManagedClient(ctx context.Context, path string) error
 				break
 			}
 			if errors.Is(err, errLogLineToLong) {
-				c.logger.Err(err).Str("line", string(line)).Msg("skipping too long log line")
+				c.logger.Info().Str("line", string(line)).Msg("truncated source plugin log line")
```
yevgenypats, line 149: I wonder if the word `source plugin` is needed. our logger has context so it is contexted with the source plugin name and all other properties.

yevgenypats (commented):
Small nit on logs.

clients/source.go
erezrokah: This logging happens in the context of the CLI logger so `module=cli`.
It comes from https://github.com/cloudquery/cloudquery/blob/9f33362ebc9989ca0f6f74eac9e5700657a18116/cli/cmd/sync.go#L79

You might be referring to the plugin logger that is created here https://github.com/cloudquery/plugin-sdk/blob/448232c8789350c8fb071902d33a5c5f07d2b82c/plugins/source/plugin.go#L153 from the serve command

yevgenypats (approved):

Issue #613, opened 2023-01-17
Bug(env-expanding): Commented env variables should not be expanded


erezrokah:
### Describe the bug

Given the following configuration:

```yaml
kind: source
spec:
  name: "azure"
  path: "cloudquery/azure"
  version: "v3.3.0"
  tables: ["*"]
  destinations: ["postgresql"]
  concurrency: 1000
  spec:
    subscriptions:
      - 21e0d134-5da8-4754-8587-c58425d93de5
---
kind: destination
spec:
  name: "postgresql"
  path: "cloudquery/postgresql"
  version: "v2.0.5"
  write_mode: "overwrite-delete-stale"
  spec:
    # connection_string: ${TEST_STRING}
    connection_string: postgresql://postgres:pass@localhost:5432/postgres
```

Running `cloudquery sync <path-to-config>` you get an error if the `TEST_STRING` variable is not set

### Expected Behavior

Don't fail on missing env vars in comments

### Steps to Reproduce

Run `cloudquery sync` with the provider configuration

### Possible Solution

Ignore comments when expanding env variables

### Provider and CloudQuery version

`cloudquery version 2.1.0`

hermanschaaf:
Related to https://github.com/cloudquery/plugin-sdk/issues/480 (this problem was also mentioned in a comment there)

yevgenypats:
I'll reply the same as https://github.com/cloudquery/plugin-sdk/issues/480. I don't think this is something we need to solve as it's a product on it's own. i.e: "smart context based variable expansion".

I think a documentaiton on how variable expansion works:
1) plain search and replace
2) specify what are the limitation

should solve both those issues.

Issue #480, opened 2022-12-09
Environment & File Variable replacement should escape their contents


hermanschaaf:
Right now environment & file variable replacement in yaml config files do a simple string replacement, which can lead to hard-to-debug issues. For example, if the contents of a file contains JSON, the following config:

```
  spec:
    project_id: "cq-playground"
    dataset_id: "cloudquery"
    service_account_key_json: ${file:key.json}
```

will throw an error about unexpected keys in the JSON. That's because the JSON gets interpreted as an additional yaml object. The workaround is to place the variable in single quotes:

```
  spec:
    project_id: "cq-playground"
    dataset_id: "cloudquery"
    service_account_key_json: '${file:key.json}'
```

Similarly, if you were to use double quotes instead of single quotes there, you'd also get a hard-to-debug error:

```
  spec:
    project_id: "cq-playground"
    dataset_id: "cloudquery"
    service_account_key_json: "${file:key.json}"
```

That's because the JSON file contents may include double quotes, which won't get escaped. So you'll get `    service_account_key_json: "{"name":"value"}"` which is invalid yaml syntax.

Or if the file contains newlines, you can also run into similar issues. I don't think this is how variables are intended to be used, and we should escape their contents. I'm not sure what the right way will be to do this; we'd still want number fields to be interpreted as numbers, so we can't (for example) add quotes around interpolated variables.

I'm just writing this issue now so I don't forget about it, but won't go into too much detail now.

hermanschaaf:
Here's another case that can cause issues:

```
  spec:
    project_id: "cq-playground"
    dataset_id: "cloudquery"
    # service_account_key_json: "${file:key.json}"
```

Even though the line is commented out, it expands into multiple lines (most of which are not commented out) and so you get an error related to an unexpected yaml key.

yevgenypats:
I wonder how this is done in other configuration and variable expansion places. Im not sure we can escape the content and I think it should actually be part of the user responsibility. For example I can think of some edge cases that wont make the escape work:

```
spec:
    project_id: "something${file:key.json}"
```

so potentially escaping here wont work.

I think the variable substitution should be just simple search and replace. maybe we can have some flag to print the expanded config to help debug it but I wouldn't go into the rabbit whole of understanding where the variable is and trying to understand if and how we need to quote it.

hermanschaaf:
@yevgenypats I'm not sure I understand why escaping won't work in this case:

```
spec:
    project_id: "something${file:key.json}"
[...]

PR #614, opened 2023-01-17
feat(spec): Report all expand errors

erezrokah:
#### Summary

Related to https://github.com/cloudquery/plugin-sdk/issues/613 and https://github.com/cloudquery/plugin-sdk/issues/480.
This PR refactors the code to use built in `os.Expand` and avoids using multiple regular expressions.
It also reports all relevant errors instead of just the last one

---

- [ ] Update or add tests 🧪

### DIFF
--- a/specs/spec_reader.go
+++ b/specs/spec_reader.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"regexp"
 	"strings"
 )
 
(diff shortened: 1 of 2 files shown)


hermanschaaf (approved):
LGTM, nice one! Maybe just worth noting that there is a slight change in behavior: I believe previously we didn't expand `$VAR` but now we do, because `os.Expand` accepts both `$VAR` and `${VAR}`

yevgenypats (commented):
Nice one! Per @hermanschaaf comment maybe we do want to keep the regex to expand stuff as we have more control over it as I feel we might go back and forth with this code as before it was `os.Expand` and then I needed some more control so I falled back to regex. It's not that `os.Expand` saves alot of code in that case?

erezrokah:
> Nice one! Per @hermanschaaf comment maybe we do want to keep the regex to expand stuff as we have more control over it as I feel we might go back and forth with this code as before it was `os.Expand` and then I needed some more control so I falled back to regex. It's not that `os.Expand` saves alot of code in that case?

I think we used to support both `$` and `${}` as we used `os.ExpandEnv`. I think maybe you're referring to `os.ExpandEnv` lack of control? See https://github.com/cloudquery/plugin-sdk/pull/499/files#diff-12fc83214927c910b85cf4b5e8ec7e26de428fd7293767e4a807bfdcab8f042bL41

I think this mostly saves the usage of RegExp to re-implement code that's already available in the standard lib

yevgenypats:
> 

I would actually want to rollback to regex as it was already implemented so it doesn't really save us from implementing it again. The reason is because if we ever want to add additional find/replace capability the code is already there and we wont need to go back and forth. basically the version of the code right now is a more advanced so I don't see a reason to move us back and then move here again if we will need it.

erezrokah:
> basically the version of the code right now is a more advanced so I don't see a reason to move us back

Can you explain how the current code is more advanced and how this is a step back?
To add more functionality with the current code one would need write a new Regex (which is error prone) + mostly copy paste a new `expand*` function (and if we want multiple errors report those separately). 
With the new version the expand logic is the same for all cases, and adding new functionality will be in a single place under the same function. Seems like `os.Expand` exists for this use case of providing custom logic when expanding $ placeholders.

If we ever want to do more transformations on the config. it will probably best not to use regular expressions, and process the YAML object to avoid things like https://github.com/cloudquery/plugin-sdk/issues/613.

Regardless I don't think it's high priority so I'll close this PR and we can re-visit if we ever add new functionality

PR #618, opened 2023-01-19
chore(test): Add overwrite + migrate destination plugin test

candiduslynx:


hermanschaaf (changes requested):
It seems like the test runs the code in overwrite mode, but doesn't actually do any overwrite-specific assertions, because all the items being written have unique PKs. Shouldn't we test / expect some difference between append and overwrite modes?

candiduslynx:
> It seems like the test runs the code in overwrite mode, but doesn't actually do any overwrite-specific assertions, because all the items being written have unique PKs. Shouldn't we test / expect some difference between append and overwrite modes?

We want to test that overwrite works if the column order changed. The actual overwrite with check regarding overwritten data is done in a separate test.

hermanschaaf (approved):

yevgenypats (approved):

PR #619, opened 2023-01-21
feat: Add version discovery service

yevgenypats:
This adds discovery service to support multiple versions. Follow-up to this: https://github.com/cloudquery/plugin-sdk/pull/610


clients/discovery/v0/discovery.go
```diff
@@ -0,0 +1,203 @@
...
+			return nil, err
+		}
+	default:
+		return nil, fmt.Errorf("unsupported registry %s", registrySpec)
+	}
+
+	return c, nil
+}
+
+// newManagedClient starts a new destination plugin process from local file, connects to it via gRPC server
```
erezrokah, line 105:
```suggestion
// newManagedClient starts a new discovery plugin process from local file, connects to it via gRPC server
```

erezrokah (approved):
Looks great 🚀 I wish we could avoid spawning a new process only to get the versions, but I don't see a way around it at the moment

yevgenypats:
> Looks great 🚀 I wish we could avoid spawning a new process only to get the versions, but I don't see a way around it at the moment

Yeah I thought about this as well but yeah don't see an easy way around it, at least not without some refactoring :(

PR #625, opened 2023-01-25
fix: Ignore env variables in comments

hermanschaaf:
- Fixes https://github.com/cloudquery/plugin-sdk/issues/613

This changes the yaml config parsing code to ignore comments entirely. The solution used here is relatively short and clean, but it required some unfortunate hacking because the curly brackets we use for env variable placeholders are not valid yaml, and so any attempt to parse the yaml with them still in place fails (unless we were to write our own parser). 

Simple regex replacement of comments would also be prone to issues, because it wouldn't be aware of the context (maybe the `#` is inside a string, for example). This solution replaces the variables with temporary placeholders, converts each yaml section to json and back to yaml to strip comments, and then replaces the temporary placeholders again before continuing as before.

### DIFF
--- a/specs/spec_reader.go
+++ b/specs/spec_reader.go
@@ -1,11 +1,14 @@
 package specs
 
 import (
+	"bytes"
 	"fmt"
 	"os"
 	"path/filepath"
 	"regexp"
 	"strings"
+
+	"github.com/ghodss/yaml"
 )
 
 type SpecReader struct {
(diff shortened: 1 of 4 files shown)


specs/spec_reader.go
```diff
@@ -173,3 +182,34 @@ func NewSpecReader(paths []string) (*SpecReader, error) {
...
+	}
+
+	// replace placeholder variables with valid yaml, otherwise it cannot be parsed
+	// in some cases. Short of writing our own yaml parser to remove comments,
+	// this seems like the best we can do.
+	b = envRegex.ReplaceAllFunc(b, func(match []byte) []byte {
+		content := envRegex.FindSubmatch(match)[1]
+		return []byte(openPlaceholder + string(content) + closePlaceholder)
+	})
+	j, err := yaml.YAMLToJSON(b)
```
candiduslynx, line 203: already done in `SpecUnmarshalYamlStrict`, maybe we could do yaml-> json and then work with json only?

specs/spec_reader.go
```diff
@@ -173,3 +182,34 @@ func NewSpecReader(paths []string) (*SpecReader, error) {
...
+func stripYamlComments(b []byte) ([]byte, error) {
+	const openPlaceholder = "$$$OPEN$$$"
+	const closePlaceholder = "$$$CLOSE$$$"
+
+	// return an error if the yaml already contains our temporary placeholder for env variables by some unlucky coincidence
+	if bytes.Contains(b, []byte(openPlaceholder)) || bytes.Contains(b, []byte(closePlaceholder)) {
+		return nil, fmt.Errorf("%s and %s are reserved words in CloudQuery config", openPlaceholder, closePlaceholder)
+	}
+
+	// replace placeholder variables with valid yaml, otherwise it cannot be parsed
```
erezrokah, line 196: GitHub wrote their own YAML parser. We should not do it https://github.com/cloudquery/plugin-sdk/issues/480#issuecomment-1404694684

erezrokah (approved):
Approving and 
giphy (1)

specs/spec_reader.go
hermanschaaf: We would then be breaking some functionality that we know some users are relying on (and requested several times in the past when we didn't have it, as I was informed by @bbernays): the ability to store actual yaml inside environment variables and having it expand. I am of the opinion that doing this is a terrible idea, but for now I just want to fix the thing we can all agree needs to be fixed, which is that environment variables in comments should be ignored.

yevgenypats (approved):

PR #626, opened 2023-01-26
feat: Validate PK Creation

bbernays:
#### Summary

Currently there is no warning or error if you add a PK field  (`WithPrimaryKey()` that doesn't correspond to an actual field

---

- [ ] Update or add tests 🧪

### DIFF
--- a/transformers/struct.go
+++ b/transformers/struct.go
@@ -151,7 +151,11 @@ func TransformWithStruct(st any, opts ...StructTransformerOption) schema.Transfo
 				}
 			}
 		}
-
+		// PKs can be defined in the struct tags or via the WithPrimaryKeys option, so the only heuristic we can apply
+		// is to validate that the total number of created PKs is not less than the total number of created PKs
+		if len(t.table.PrimaryKeys()) < len(t.pkFields) {
+			return fmt.Errorf("failed to create all of the desired primary keys: %v", t.pkFields)
+		}
 		return nil
 	}
 }


transformers/struct.go
```diff
@@ -151,7 +151,11 @@ func TransformWithStruct(st any, opts ...StructTransformerOption) schema.Transfo
 				}
 			}
 		}
-
+		// PKs can be defined in the struct tags or via the WithPrimaryKeys option, so the only heuristic we can apply
+		// is to validate that the total number of created PKs is not less than the total number of created PKs
+		if len(t.table.PrimaryKeys()) < len(t.pkFields) {
```
erezrokah, line 156: Nice 🚀 I get the check is a bit tricky as we define PKS on fields, but validate on columns

erezrokah (approved):
Great addition, can we add a test for it?

transformers/struct.go
hermanschaaf: What if one of the primary keys isn't a field on the struct - i.e. it's a custom column - will this check still work? I had a case like that not too long ago, can try and find it if you need an example

hermanschaaf (changes requested):
Really needed feature and clever way of achieving it! Just need to check it's not going to prevent some valid cases from going through

transformers/struct.go
erezrokah: Ah, right it won't work

bbernays: @hermanschaaf - if you can find the example that would be great!

hermanschaaf: @bbernays Sure, here's one recent example from the Alicloud plugin that came to mind: https://github.com/cloudquery/cloudquery/blob/main/plugins/source/alicloud/resources/services/oss/buckets.go#L10

But I think it will happen any time we put the multiplexer value(s) as separate columns and want to also have them be part of the PK. I believe there are lots of examples of this in AWS as well

erezrokah: I think we will need to keep track which PKs matches a column here https://github.com/cloudquery/plugin-sdk/blob/08bace89c708ca7f20490ce9756f8276b7e5d6f2/transformers/struct.go#L271 and fail at the end if we didn't have a match for all of them

bbernays: Updated the code to do this

bbernays: (Still working on the tests)...

transformers/struct.go
```diff
@@ -151,7 +153,10 @@ func TransformWithStruct(st any, opts ...StructTransformerOption) schema.Transfo
 				}
 			}
 		}
-
+		// Validate that all expected PK fields were found
+		if diff := funk.SubtractString(t.pkFields, t.pkFieldsFound); len(diff) > 0 {
```
yevgenypats, line 157: Is the order important here. can we just use set comparision?

bbernays: The diff is used to be able to provide an accurate message to the user

disq: Order doesn't seem important https://github.com/thoas/go-funk#funksubtract

yevgenypats (approved):

PR #630, opened 2023-01-26
fix: Only call `newExecutionClient` if needed

erezrokah:
#### Summary

This is more of hot fix/patch to https://github.com/cloudquery/plugin-sdk/pull/610.
https://github.com/cloudquery/plugin-sdk/pull/610 had an unwanted side effect on the migrate command, as now we init the plugin before running it.
initializing the plugin is expensive, requires credentials and does a whole lot of validations, for example:
https://github.com/cloudquery/cloudquery/blob/a06139efad9a2e8de14f186a223222c9d52ce58a/plugins/source/aws/client/client.go#L383
https://github.com/cloudquery/cloudquery/blob/a06139efad9a2e8de14f186a223222c9d52ce58a/plugins/source/azure/client/client.go#L124

Those seem unnecessary for tables migration (unless one uses dynamic tables).

---

- [ ] Update or add tests 🧪

### DIFF
--- a/plugins/source/plugin.go
+++ b/plugins/source/plugin.go
@@ -221,12 +221,12 @@ func (p *Plugin) Init(ctx context.Context, spec specs.Source) error {
 		return fmt.Errorf("unknown backend: %s", spec.Backend)
 	}
 
-	p.client, err = p.newExecutionClient(ctx, p.logger, spec, Options{Backend: p.backend})
-	if err != nil {
-		return fmt.Errorf("failed to create execution client for source plugin %s: %w", p.name, err)
-	}
 	tables := p.tables
 	if p.getDynamicTables != nil {
+		p.client, err = p.newExecutionClient(ctx, p.logger, spec, Options{Backend: p.backend})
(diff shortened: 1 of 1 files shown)


yevgenypats (changes requested):
Looks good. Seems we need to add then the same workaround for `GetDynamicTables`.  and use `GetTablesForSpec` in `GetDynamicTables`. Im also not a fan of this fix but I guess that's fine for now. basically it doesn make sense to authenticate to migrate in a way but yeah if `dynamic_tables` is `nil` then in theory this shouldn't be needed so that's the only workaround possible actually

erezrokah:
> Seems we need to add then the same workaround for `GetDynamicTables`. and use `GetTablesForSpec` in `GetDynamicTables`

Can you explain this a bit? Not sure I got it, `sessionTables` is still initialized by `Init` even if you don't have dynamic tables

yevgenypats (approved):
I see. I think it's good then. I missed it in the GitHub UI, sorry.

PR #632, opened 2023-01-27
feat: Support downloading plugins from other cloudquery repos

hermanschaaf:
Right now, if a plugin is hosted in anywhere on the `cloudquery` org, it can only be downloaded if it is part of the `cloudquery/cloudquery` monorepo. This makes it impossible to temporarily host `cq-source-*` type plugins as other `cloudquery` repos, as can be done with community plugins on other orgs.

This changes the download logic to first check the monorepo, but fall back to the community standard of `cq-source-*` if it's A) `cloudquery` org and B) the monorepo download attempt resulted in a 404. If both fail with 404, the attempt is aborted. Other status codes are retried.

It would have been better if we could know, somehow, when a plugin is hosted in the monorepo or outside it, but right now this is not possible unless we change the spec in some way. 

This change allows us to use the CLI to download https://github.com/cloudquery/cq-source-simple-analytics

### DIFF
--- a/registry/download.go
+++ b/registry/download.go
@@ -26,7 +26,7 @@ const (
 	RetryWaitTime                    = 1 * time.Second
 )
 
-type pluginUrl struct {
+type pluginURL struct {
 	url      string
 	monorepo bool
 }
@@ -35,14 +35,14 @@ func DownloadPluginFromGithub(ctx context.Context, localPath string, org string,
 	downloadDir := filepath.Dir(localPath)
 	pluginZipPath := localPath + ".zip"
 	// https://github.com/cloudquery/cloudquery/releases/download/plugins-source-test-v1.1.5/test_darwin_amd64.zip
-	urls := []pluginUrl{
+	urls := []pluginURL{
 		// community plugin format
-		pluginUrl{url: fmt.Sprintf("https://github.com/%s/cq-%s-%s/releases/download/%s/cq-%s-%s_%s_%s.zip", org, typ, name, version, typ, name, runtime.GOOS, runtime.GOARCH)},
+		{url: fmt.Sprintf("https://github.com/%s/cq-%s-%s/releases/download/%s/cq-%s-%s_%s_%s.zip", org, typ, name, version, typ, name, runtime.GOOS, runtime.GOARCH)},
 	}
 	if org == "cloudquery" {
 		urls = append(
 			// CloudQuery monorepo plugin
-			[]pluginUrl{{url: fmt.Sprintf("https://github.com/cloudquery/cloudquery/releases/download/plugins-%s-%s-%s/%s_%s_%s.zip", typ, name, version, name, runtime.GOOS, runtime.GOARCH), monorepo: true}},
+			[]pluginURL{{url: fmt.Sprintf("https://github.com/cloudquery/cloudquery/releases/download/plugins-%s-%s-%s/%s_%s_%s.zip", typ, name, version, name, runtime.GOOS, runtime.GOARCH), monorepo: true}},
 			// fall back to community plugin format if the plugin is not found in the monorepo
 			urls...,
 		)
@@ -94,11 +94,11 @@ func DownloadPluginFromGithub(ctx context.Context, localPath string, org string,
 	return nil
 }
 
-func downloadFile(ctx context.Context, localPath string, urls ...pluginUrl) (used pluginUrl, err error) {
+func downloadFile(ctx context.Context, localPath string, urls ...pluginURL) (used pluginURL, err error) {
 	// Create the file
 	out, err := os.Create(localPath)
 	if err != nil {
-		return pluginUrl{}, fmt.Errorf("failed to create file %s: %w", localPath, err)
+		return pluginURL{}, fmt.Errorf("failed to create file %s: %w", localPath, err)
 	}
 	defer out.Close()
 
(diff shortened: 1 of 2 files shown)


registry/download.go
```diff
@@ -30,10 +29,20 @@ func DownloadPluginFromGithub(ctx context.Context, localPath string, org string,
...
+			fmt.Sprintf("https://github.com/cloudquery/cloudquery/releases/download/plugins-%s-%s-%s/%s_%s_%s.zip", typ, name, version, name, runtime.GOOS, runtime.GOARCH),
+			// community plugin under CloudQuery org
+			fmt.Sprintf("https://github.com/cloudquery/cq-%s-%s/releases/download/%s/cq-%s-%s_%s_%s.zip", typ, name, version, typ, name, runtime.GOOS, runtime.GOARCH),
+		}
+	} else {
 		// https://github.com/yevgenypats/cq-source-test/releases/download/v1.0.1/cq-source-test_darwin_amd64.zip
-		downloadURL = fmt.Sprintf("https://github.com/%s/cq-%s-%s/releases/download/%s/cq-%s-%s_%s_%s.zip", org, typ, name, version, typ, name, runtime.GOOS, runtime.GOARCH)
+		urls = []string{
+			// community plugin under user org
+			fmt.Sprintf("https://github.com/%s/cq-%s-%s/releases/download/%s/cq-%s-%s_%s_%s.zip", org, typ, name, version, typ, name, runtime.GOOS, runtime.GOARCH),
```
disq, line 44: Why not just have this as the initial `urls` value (as an item in the slice) and `if org == cloudquery` we just prepend our own monorepo to beginning? Would look simpler and less of a sea of string formatting.

hermanschaaf: Sure, that can work

registry/download.go
```diff
@@ -80,58 +91,53 @@ func DownloadPluginFromGithub(ctx context.Context, localPath string, org string,
...
-		retry.Attempts(RetryAttempts),
-		retry.Delay(RetryWaitTime),
-	)
-
-	if err != nil {
-		return fmt.Errorf("failed downloading: %s", url)
+			resp.Body.Close()
+			return i, nil
+		}
+		time.Sleep(RetryWaitTime)
```
disq, line 137: Why not keep retry logic and just don't retry on 404? Also with the logic above, looks like for each retry URLs would start from index 0?

hermanschaaf: Until now we have retried on 404s, so I wanted to keep that behavior the same. But maybe we shouldn't 🤷 Though 404s can happen for other reasons, like the plugin version still being in the process of being published. So with this updated logic things are exactly the same as before, but if it's the cloudquery repo we also check the secondary url on 404 before sleeping for the next try.

erezrokah: The retry logic was added due to https://github.com/cloudquery/plugin-sdk/issues/291 (503 GitHub errors).
I don't this we've see other error codes so I think we don't need to retry on 404

hermanschaaf: Thanks for the context @erezrokah, I'll update this to not retry on 404s, but fall back if there are more urls to try.

hermanschaaf: Updated 👍 I also added an integration test that tries out the various cases

erezrokah (approved):

registry/download.go
```diff
@@ -30,10 +30,17 @@ func DownloadPluginFromGithub(ctx context.Context, localPath string, org string,
...
+	urls := []string{
+		// community plugin format
+		fmt.Sprintf("https://github.com/%s/cq-%s-%s/releases/download/%s/cq-%s-%s_%s_%s.zip", org, typ, name, version, typ, name, runtime.GOOS, runtime.GOARCH),
+	}
+	if org == "cloudquery" {
+		urls = []string{
+			// CloudQuery monorepo plugin
+			fmt.Sprintf("https://github.com/cloudquery/cloudquery/releases/download/plugins-%s-%s-%s/%s_%s_%s.zip", typ, name, version, name, runtime.GOOS, runtime.GOARCH),
+			// fall back to community plugin format if the plugin is not found in the monorepo
+			urls[0],
```
disq, line 42: why not just `urls = append([]string{...}, urls...)`? urls[0] would fail if we introduce something that leaves the community plugin empty, or would miss things if we come up with more than one format in the future etc. Using `append` would make it resilient to those changes.

registry/download.go
```diff
@@ -55,8 +62,10 @@ func DownloadPluginFromGithub(ctx context.Context, localPath string, org string,
 	}
 	defer archive.Close()
 
-	pathInArchive := fmt.Sprintf("plugins/%s/%s", typ, name)
-	if org != "cloudquery" {
+	var pathInArchive string
+	if org == "cloudquery" && urlIndex == 0 {
```
disq, line 66: This indirect comparison looks a bit smelly but for now it's ok I think, as other methods would make it more complicated

registry/download.go
hermanschaaf: I felt like this was more readable and explicit, and I would hope that someone changing the code in the future would read the one or two lines of surrounding context before doing so, and we can't really know what those future changes will be. But I don't think it matters, I will change it.

disq: append reads more natural to me than a specific index of a pre-set slice etc. 👍

registry/download.go
hermanschaaf: Yeah, agreed... I think I improved it a little bit now by using a struct around the urls instead, PTAL

disq: Nice

disq (approved):

yevgenypats (approved):

erezrokah:
Opened https://github.com/cloudquery/plugin-sdk/issues/655 as a follow up as I'm not sure the current URL logic for community plugins is what we want

PR #638, opened 2023-01-31
feat(sentry): Report source plugin & its version on PK duplicates

candiduslynx:
Also moved some parts to worker altogether.


plugins/destination/managed_writer.go
```diff
@@ -15,39 +15,44 @@ import (
...
+
 	ch    chan schema.CQTypes
 	flush chan chan bool
+
+	table   *schema.Table
+	srcSpec *specs.Source
 }
 
-func (p *Plugin) worker(ctx context.Context, metrics *Metrics, table *schema.Table, ch <-chan schema.CQTypes, flush <-chan chan bool) {
+func (p *Plugin) worker(ctx context.Context, metrics *Metrics, w *worker) {
```
hermanschaaf, line 27: Maybe let's call it `runWorker`, because it's already passing a `worker` in as an argument

candiduslynx:
close for now

PR #651, opened 2023-02-01
fix: Handle numbers in env variables

hermanschaaf:
This fixes an edge case in the yaml comment removal code, where environment variables containing only the numbers lost any quotes around them during the yaml -> json -> yaml conversion process. We can force quotes being maintained by making sure the placeholders are replaced by numbers. If they were quoted before, the yaml conversion process will then ensure the quotes remain. In this implementation we use random numbers with a constant seed, so it's spread out but predictable. We also check if the number already occurs in the config (which is incredibly unlikely, but anyway) and choose a different one if so. 

Fixes https://github.com/cloudquery/plugin-sdk/issues/652


disq (approved):

erezrokah (approved):
🙈

specs/testdata/number.txt
```diff
@@ -0,0 +1 @@
+12345
```
bbernays, line 1: Can we try this with an accountID that begins with a `0`? It is a common problem that tools mishandle accountIDs that begin with 0s... I think it was EKS when it was first released the first 6 hours you couldn't use it any accounts that begin with a `0`

hermanschaaf: Changed the case, but also added a test where we specifically assert that the name is preserved correctly (otherwise maybe it could be a string, but with the first 0 trimmed off) :)

yevgenypats:
we might want to change the approach instead of yaml->json->yaml  just do a small regex removing comments from the yaml . I've a feeling yaml->json->yaml  is a place for a lot of such edge cases. but we can do that in the next edge case.

PR #654, opened 2023-02-01
feat: Enable Custom Validators

bbernays:
#### Summary

Enable each individual plugin to pass in any and all additional validator functions they want...

---

- [ ] Update or add tests 🧪

### DIFF
--- a/plugins/source/testing.go
+++ b/plugins/source/testing.go
@@ -8,14 +8,14 @@ import (
 	"github.com/cloudquery/plugin-sdk/specs"
 )
 
-type Validator func(t *testing.T, tables schema.Tables, resources []*schema.Resource)
+type Validator func(t *testing.T, plugin *Plugin, resources []*schema.Resource)
 
 func TestPluginSync(t *testing.T, plugin *Plugin, spec specs.Source, opts ...TestPluginOption) {
 	t.Helper()
 
 	o := &testPluginOptions{
 		parallel:   true,
-		validators: []Validator{validateTables},
+		validators: []Validator{validatePlugin},
 	}
 	for _, opt := range opts {
 		opt(o)
@@ -44,7 +44,7 @@ func TestPluginSync(t *testing.T, plugin *Plugin, spec specs.Source, opts ...Tes
 		t.Fatal(syncErr)
 	}
 	for _, validator := range o.validators {
-		validator(t, plugin.Tables(), syncedResources)
+		validator(t, plugin, syncedResources)
 	}
 }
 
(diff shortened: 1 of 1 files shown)


bbernays:
> Nice, maybe you can share a PR that uses this?

I am working on one now in the AWS plugin to check for the Tags issue we just solved... https://github.com/cloudquery/cloudquery/pull/7621

candiduslynx (approved):

disq (approved):

bbernays:
@erezrokah The initial pr for utilizing this [functionality is up](https://github.com/cloudquery/cloudquery/pull/7651)... It already found 5 issues...

hermanschaaf (approved):
I'd probably define the Validator function signature as a type, but it's a minor nit 👍

yevgenypats (changes requested):
Looks good. Few thoughts on the validator `interface`

`func(t *testing.T, table *schema.Table, resources *schema.Resource)`

This way the SDK can do the same loop instead of the validator doing the loop again. Unless there is a need for the validator to have access to all tables but I feel it has those when it is calling the test anyway so feel a bit redundant.

Nit: per Herman comment type for the func would be nice.

bbernays:
> Looks good. Few thoughts on the validator `interface`
> 
> `func(t *testing.T, table *schema.Table, resources *schema.Resource)`
> 
> This way the SDK can do the same loop instead of the validator doing the loop again. Unless there is a need for the validator to have access to all tables but I feel it has those when it is calling the test anyway so feel a bit redundant.

I wanted to give the validator access to  all of the tables so that it could validate/ensure consistency across multiple tables... I thought that the hit for performance is negligible...

 
> Nit: per Herman comment type for the func would be nice.

 will look into this

yevgenypats:
> > Looks good. Few thoughts on the validator `interface`
> > `func(t *testing.T, table *schema.Table, resources *schema.Resource)`
> > This way the SDK can do the same loop instead of the validator doing the loop again. Unless there is a need for the validator to have access to all tables but I feel it has those when it is calling the test anyway so feel a bit redundant.
> 
> I wanted to give the validator access to all of the tables so that it could validate/ensure consistency across multiple tables... I thought that the hit for performance is negligible...
> 
> > Nit: per Herman comment type for the func would be nice.
> 
> will look into this

But the validator already have access to all the tables because 

> > Looks good. Few thoughts on the validator `interface`
> > `func(t *testing.T, table *schema.Table, resources *schema.Resource)`
> > This way the SDK can do the same loop instead of the validator doing the loop again. Unless there is a need for the validator to have access to all tables but I feel it has those when it is calling the test anyway so feel a bit redundant.
> 
> I wanted to give the validator access to all of the tables so that it could validate/ensure consistency across multiple tables... I thought that the hit for performance is negligible...
> 
> > Nit: per Herman comment type for the func would be nice.
> 
> will look into this

But the validator already has in some sense access to all the tables. Anyway, if we want to pass the whole context to the validator, I think it makes sense but then let's just use the `Plugin` as a whole?

candiduslynx (approved):

yevgenypats (approved):

PR #662, opened 2023-02-06
feat: Add skip_dependent_tables option

hermanschaaf:
This adds a flag to the source config that allows users to opt out of syncing dependent tables by default. This has been raised as an issue by users a few times already, most recently [here](https://discord.com/channels/872925471417962546/873606591335759872/1071081863105495178). 

The issue with the current default behavior is that new dependencies can be added with minor versions, which can 1) drastically slow down a sync and 2) create unexpected migrations or new tables. It can also be surprising for new users, when selecting e.g. only the `aws_s3_buckets` table, that they get ~6 tables synced.

This should ideally be the default behavior, but that would be a breaking change. For now, I think we should keep backwards-compatibility but add this flag (and recommend it as best-practice).

Note about naming: `dependent` is the standard American spelling. `dependant` is used in the UK. I went with the American spelling here.

### DIFF
--- a/plugins/destination/plugin.go
+++ b/plugins/destination/plugin.go
@@ -262,7 +262,7 @@ func (p *Plugin) Write(ctx context.Context, sourceSpec specs.Source, tables sche
 			exclude := func(t *schema.Table) bool {
 				return t.IsIncremental
 			}
-			tablesToDelete = tables.FilterDfsFunc(include, exclude)
+			tablesToDelete = tables.FilterDfsFunc(include, exclude, sourceSpec.SkipDependentTables)
 		}
 		if err := p.DeleteStale(ctx, tablesToDelete, sourceSpec.Name, syncTime); err != nil {
 			return err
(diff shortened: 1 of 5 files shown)


hermanschaaf:
@erezrokah Yes, that's the current workaround. But to be safe, you'd need an entry like that for every table you want to sync. So your config may end up looking like:

```
tables:
    - aws_ec2_instances
    - aws_s3_buckets
    - aws_organizations_accounts
 skip_tables:
    - aws_ec2_instance_*
    - aws_s3_bucket_*
    - aws_organizations_account_*
```

etc, and that's what we've recommended users do up to now. But that's not always going to work. There are cases [like this one](https://github.com/cloudquery/cloudquery/pull/7688/files#diff-5069862e0827dc6c3b3502644f7a44eff1963607efab94f5916a09119949d2d7R176) where if you were syncing `aws_ecs_cluster_services` and skipping `aws_ecs_cluster_*` before, you'd now also suddenly be syncing `aws_ecs_cluster_task_sets`. I think this default behavior is surprising and users should have a way to disable it.

erezrokah (approved):

disq (approved):

yevgenypats (approved):

PR #664, opened 2023-02-07
feat: Use Setpgid=true on Unix systems so that signals are not sent to the child process

hermanschaaf:
By default, commands in Go are started in the same process group. When a signal is sent (e.g. ctrl+c), this gets sent to the entire group, causing plugins to shut down before the CLI has a chance to do closing operations and then relay the signal. By setting `Setpgid: true` on UNIX systems, we can get better control over the shutdown process of child processes.


disq (approved):

hermanschaaf:
Oh okay, this field is not even available on Windows :) I will refactor a bit

PR #667, opened 2023-02-08
feat: Add Validation to Ensure All table names are plural

bbernays:
#### Summary

In Codegen we had functionality to ensure that the table name was plural and consistent... This adds a new validator to ensure no regressions

---

- [ ] Update or add tests 🧪


plugins/source/unit_test_validators.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+	}
+	return nil
+}
+
+func validateTableNamePlural(t *testing.T, plugin *Plugin, resources []*schema.Resource) error {
+	t.Helper()
+	tables := extractTables(plugin.tables)
+	for _, table := range tables {
+		if !strings.HasSuffix(table.Name, "s") {
```
erezrokah, line 29: Should we use https://github.com/gertd/go-pluralize (or any other similar lib) to check this?
Otherwise we're only testing that tables end with `s`

erezrokah (commented):
Nice @bbernays added a comment

plugins/source/unit_test_validators.go
bbernays: We can definitely do something more substantial, I was just pulling the same functionality that we had in codegen here...

hermanschaaf (approved):

erezrokah (approved):

plugins/source/unit_test_validators.go
```diff
@@ -0,0 +1,99 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/cloudquery/plugin-sdk/schema"
+	pluralize "github.com/gertd/go-pluralize"
+)
+
+type Validator func(t *testing.T, plugin *Plugin, resources []*schema.Resource) error
```
disq, line 11: I think it should either have `t` or `error`, not both? You can check if it errored by calling `t.Failed()`

disq (changes requested):

disq (approved):

plugins/source/testing.go
```diff
@@ -8,14 +8,12 @@ import (
...
 
-type Validator func(t *testing.T, plugin *Plugin, resources []*schema.Resource)
-
 func TestPluginSync(t *testing.T, plugin *Plugin, spec specs.Source, opts ...TestPluginOption) {
 	t.Helper()
 
 	o := &testPluginOptions{
 		parallel:   true,
-		validators: []Validator{validatePlugin},
+		validators: []Validator{validateColumnsHaveData, validateTableNamePlural},
```
yevgenypats, line 16: `validateColumnsHaveData` make sense to have as a default but `validateTableNamePlural` seems like something to be defined by the plugin itself.

yevgenypats (changes requested):
I think we should start by defining the EnforcePlural on the plugin side and not making it as default on the sdk side.

PR #668, opened 2023-02-08
feat: Add table diff methods

yevgenypats:
This adds diff methods for `Table`. This will help simplify migrations so we can transform the destination/database into `Table` struct and then understand what was changed/added in the same way across all destination plugins.

PostgreSQL PR example - https://github.com/cloudquery/cloudquery/pull/7819


plugins/source/plugin.go
```diff
@@ -33,6 +33,11 @@ type Plugin struct {
 	newExecutionClient NewExecutionClientFunc
 	// dynamic table function if specified
 	getDynamicTables GetTables
+	// NoInternalColumns if set to true will not add internal columns to tables such as _cq_id and _cq_parent_id
+	// useful for sources such as PostgreSQL and other databases
+	internalColumns bool
```
disq, line 38:
```suggestion
	addInternalColumns bool
```

hermanschaaf (approved):
LGTM, just a few minor suggestions

erezrokah:
Didn't manage to comment but we already have a `write_mode: forced` in the spec:
https://github.com/cloudquery/plugin-sdk/blob/f6baa82d7d1db6d28a47bf3d206306f98aa84bd4/specs/migrate_mode.go

It's used in the SQLite destination with similar logic:
https://github.com/cloudquery/cloudquery/blob/4efbf32f77146b61b62956ca9ef41c925d644d4f/plugins/destination/sqlite/client/migrate.go#L242

yevgenypats:
> Didn't manage to comment but we already have a `migrate_mode: forced` in the spec: https://github.com/cloudquery/plugin-sdk/blob/f6baa82d7d1db6d28a47bf3d206306f98aa84bd4/specs/migrate_mode.go
> 
> It's used in the SQLite destination with similar logic: https://github.com/cloudquery/cloudquery/blob/4efbf32f77146b61b62956ca9ef41c925d644d4f/plugins/destination/sqlite/client/migrate.go#L242

Thanks , good catch. removed the options thing - https://github.com/cloudquery/plugin-sdk/pull/670

PR #671, opened 2023-02-08
fix: Destination testing add force tests

yevgenypats:
Add force tests

### DIFF
--- a/plugins/destination/plugin_testing.go
+++ b/plugins/destination/plugin_testing.go
@@ -2,18 +2,14 @@ package destination
 
 import (
 	"context"
-	"fmt"
 	"os"
 	"sort"
-	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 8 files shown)


hermanschaaf:
@yevgenypats Is it wise to only test forced migrations? I think if we need a default to test, it should be non-forced, which has been our default behavior until now

yevgenypats:
> @yevgenypats Is it wise to only test forced migrations? I think if we need a default to test, it should be non-forced, which has been our default behavior until now

We don't really use those / use it anywhere apart from the latest postgres destination.

I can add it though those tests in this PR

yevgenypats:
@hermanschaaf added proper tests for force migrations.

I can also add another type of tests where in non-force it returns an error on column type change but prefer to do it in a follow-up PR after we update all destination to this sdk.

erezrokah (approved):

PR #672, opened 2023-02-08
feat: Retain Order of Columns from struct

bbernays:
#### Summary

When we update a column resolver it will always change the order of the fields... This PR only changes the order if it is specifically requested (Adds an attribute to Columns `RetainOrder`) or if it is a primary key.

---

- [ ] Update or add tests 🧪


erezrokah:
What is the status of this PR? Can we close it, or link to an issue it solves? I don't think we consider changing columns order breaking unless one of these columns is a PK and even then we can control the order by manually configuring the columns.

erezrokah:
Closing as stale

PR #675, opened 2023-02-09
fix: Add test for NULLS in JSONs

shimonp21:
See https://github.com/cloudquery/plugin-sdk/pull/532/files for similar PR.

[the PR is labeled as 'fix' for CI reasons]

#### Summary

---

- [ ] Update or add tests 🧪

### DIFF
--- a/schema/json.go
+++ b/schema/json.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"reflect"
+	"strings"
 )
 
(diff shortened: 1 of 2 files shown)


disq (approved):

erezrokah (commented):
Hi @shimonp21, in which scenario are we getting null bytes in JSONs? Where is the data coming from? If it's only in our tests, maybe we should remove those use case

shimonp21:
stale

PR #677, opened 2023-02-12
feat: Add unmanaged sources

yevgenypats:
Used in this postgres source - https://github.com/cloudquery/cloudquery/pull/7835


plugins/source/plugin.go
```diff
@@ -22,6 +22,11 @@ type Options struct {
 
 type NewExecutionClientFunc func(context.Context, zerolog.Logger, specs.Source, Options) (schema.ClientMeta, error)
 
+type SourceUnmanagedClient interface {
```
erezrokah, line 25:
```suggestion
type UnmanagedClient interface {
```

erezrokah (approved):

erezrokah (approved):

PR #688, opened 2023-02-15
feat: Improve migration detection APIs

yevgenypats:
Migration detection APIs take 2.

- This creates **one** API that returns all the changes instead of multiple growing APIs for different set of changes.
- Improves Migration tests for different use-cases

This will come up also with a Postgres PR and documentation PR:

Basically each destination plugin would have the following table:

Migrations Logic

|  Change  | Behaviour                                        |
|----------|----------------------------------|
|  Add Column | Auto Migrate/Drop Table         | 
|  Add Column with Not Null | Auto Migrate/Drop Table  |  
|  Remove Column | Auto Migrate/Drop Table  |
|  Remove Column with Not Null | Auto Migrate/Drop Table  |
|  Change Column |  Auto Migrate/Drop Table |

and we can easily expand it in the code, tests and documentations.

Example of usage in https://github.com/cloudquery/cloudquery/pull/7819

### DIFF
--- a/internal/memdb/memdb_test.go
+++ b/internal/memdb/memdb_test.go
@@ -64,7 +64,7 @@ func TestPluginManagedClientWithSmallBatchSize(t *testing.T) {
 		}, nil,
 		destination.PluginTestSuiteTests{
 			MigrateStrategyOverwrite: migrateStrategyOverwrite,
-			MigrateStrategyAppend:    migrateStrategyOverwrite,
+			MigrateStrategyAppend:    migrateStrategyAppend,
 		})
 }
 
--- a/internal/memdb/memdb_test.go
+++ b/internal/memdb/memdb_test.go
@@ -13,6 +13,22 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+var migrateStrategyOverwrite = destination.MigrateStrategy{
+	AddColumn:           specs.MigrateModeForced,
+	AddColumnNotNull:    specs.MigrateModeForced,
+	RemoveColumn:        specs.MigrateModeForced,
+	RemoveColumnNotNull: specs.MigrateModeForced,
+	ChangeColumn:        specs.MigrateModeForced,
+}
+
+var migrateStrategyAppend = destination.MigrateStrategy{
+	AddColumn:           specs.MigrateModeForced,
+	AddColumnNotNull:    specs.MigrateModeForced,
+	RemoveColumn:        specs.MigrateModeForced,
+	RemoveColumnNotNull: specs.MigrateModeForced,
+	ChangeColumn:        specs.MigrateModeForced,
+}
+
 func TestPluginUnmanagedClient(t *testing.T) {
 	destination.PluginTestSuiteRunner(
 		t,
@@ -20,7 +36,11 @@ func TestPluginUnmanagedClient(t *testing.T) {
 			return destination.NewPlugin("test", "development", NewClient)
 		},
 		nil,
-		destination.PluginTestSuiteTests{})
+		destination.PluginTestSuiteTests{
+			MigrateStrategyOverwrite: migrateStrategyOverwrite,
+			MigrateStrategyAppend:    migrateStrategyOverwrite,
+		},
+	)
 }
 
 func TestPluginManagedClient(t *testing.T) {
@@ -29,7 +49,10 @@ func TestPluginManagedClient(t *testing.T) {
 			return destination.NewPlugin("test", "development", NewClient, destination.WithManagedWriter())
 		},
 		nil,
-		destination.PluginTestSuiteTests{})
+		destination.PluginTestSuiteTests{
+			MigrateStrategyOverwrite: migrateStrategyOverwrite,
+			MigrateStrategyAppend:    migrateStrategyOverwrite,
+		})
 }
 
 func TestPluginManagedClientWithSmallBatchSize(t *testing.T) {
@@ -39,7 +62,10 @@ func TestPluginManagedClientWithSmallBatchSize(t *testing.T) {
 				destination.WithDefaultBatchSize(1),
 				destination.WithDefaultBatchSizeBytes(1))
 		}, nil,
-		destination.PluginTestSuiteTests{})
+		destination.PluginTestSuiteTests{
+			MigrateStrategyOverwrite: migrateStrategyOverwrite,
+			MigrateStrategyAppend:    migrateStrategyOverwrite,
+		})
 }
 
 func TestPluginManagedClientWithLargeBatchSize(t *testing.T) {
@@ -50,7 +76,10 @@ func TestPluginManagedClientWithLargeBatchSize(t *testing.T) {
 				destination.WithDefaultBatchSizeBytes(100000000))
 		},
 		nil,
-		destination.PluginTestSuiteTests{})
+		destination.PluginTestSuiteTests{
+			MigrateStrategyOverwrite: migrateStrategyOverwrite,
+			MigrateStrategyAppend:    migrateStrategyOverwrite,
+		})
 }
 
 func TestPluginOnNewError(t *testing.T) {
--- a/internal/memdb/memdb_test.go
+++ b/internal/memdb/memdb_test.go
@@ -51,7 +51,7 @@ func TestPluginManagedClient(t *testing.T) {
 		nil,
 		destination.PluginTestSuiteTests{
 			MigrateStrategyOverwrite: migrateStrategyOverwrite,
-			MigrateStrategyAppend:    migrateStrategyOverwrite,
+			MigrateStrategyAppend:    migrateStrategyAppend,
 		})
 }
 
--- a/internal/memdb/memdb_test.go
+++ b/internal/memdb/memdb_test.go
@@ -78,7 +78,7 @@ func TestPluginManagedClientWithLargeBatchSize(t *testing.T) {
 		nil,
 		destination.PluginTestSuiteTests{
 			MigrateStrategyOverwrite: migrateStrategyOverwrite,
-			MigrateStrategyAppend:    migrateStrategyOverwrite,
+			MigrateStrategyAppend:    migrateStrategyAppend,
 		})
 }
 
--- a/internal/memdb/memdb_test.go
+++ b/internal/memdb/memdb_test.go
@@ -38,7 +38,7 @@ func TestPluginUnmanagedClient(t *testing.T) {
 		nil,
 		destination.PluginTestSuiteTests{
 			MigrateStrategyOverwrite: migrateStrategyOverwrite,
-			MigrateStrategyAppend:    migrateStrategyOverwrite,
+			MigrateStrategyAppend:    migrateStrategyAppend,
 		},
 	)
 }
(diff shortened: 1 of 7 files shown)


schema/table.go
```diff
@@ -248,55 +281,50 @@ func (t *Table) ValidateName() error {
...
+// changed columns: change type - drop table or drop/add column? // table.GetChangedColumns(other)
+// changed columns: added constraint not null or unique - drop table // table.GetChangedColumnsWithAddedConstraints
+// changed columns: removed constraint: not null, removed unique - drop constraint? do nothing? what if a user is adding a unique constraint? we will drop it? // table.GetChangedColumnsWithRemovedConstraints
+
+// removed columns: with no constraints - do nothing?
+// removed columns: with constraints not null or unique - drop column? drop table ? // table.RemovedColumnsWithConstraints(other)
+
+// changed pks: drop table // table.IsPrimaryKeyEqual(other)
+
+func (t *Table) GetChanges(other *Table) []TableColumnChange {
```
erezrokah, line 297:
```suggestion
func (new *Table) GetChanges(old *Table) []TableColumnChange {
```

Maybe this will make it easier to follow? Non blocking

schema/table.go
```diff
@@ -248,55 +281,50 @@ func (t *Table) ValidateName() error {
...
-func (t *Table) GetAddedColumns(other *Table) []Column {
-	var added []Column
-	for _, c := range t.Columns {
-		if other.Columns.Get(c.Name) == nil {
-			added = append(added, c)
-		}
-	}
-	return added
-}
+// added columns with no constraints - migrate // table.GetAddedColumns(other)
```
erezrokah, line 284: These comments seem a bit outdated, but I think:

```
Destrucitve changes are done only in force mode

added columns:
 - no constraints -> migrate
 - not null -> drop table
  - unique -> drop table

changed columns:
 - only type change -> drop/add column
 - added constraint -> drop table
 - removed constraint -> remove constraint (e.g. we remove a not null from a column that can be null). OK to drop a user constraint in forced mode (we can document it)

removed columns:
  - no constraints -> do nothing, unless forced, then drop
  - with constraints -> drop column, see https://github.com/cloudquery/cloudquery/blob/main/plugins/destination/postgresql/client/migrate.go#L103
```

internal/memdb/memdb_test.go
```diff
@@ -13,14 +13,34 @@ import (
...
 	destination.PluginTestSuiteRunner(
 		t,
 		func() *destination.Plugin {
 			return destination.NewPlugin("test", "development", NewClient)
 		},
 		nil,
-		destination.PluginTestSuiteTests{})
+		destination.PluginTestSuiteTests{
+			MigrateStrategyOverwrite: migrateStrategyOverwrite,
+			MigrateStrategyAppend:    migrateStrategyOverwrite,
```
erezrokah, line 41:
```suggestion
			MigrateStrategyAppend:    migrateStrategyAppend,
```

internal/memdb/memdb_test.go
```diff
@@ -29,7 +49,10 @@ func TestPluginManagedClient(t *testing.T) {
 			return destination.NewPlugin("test", "development", NewClient, destination.WithManagedWriter())
 		},
 		nil,
-		destination.PluginTestSuiteTests{})
+		destination.PluginTestSuiteTests{
+			MigrateStrategyOverwrite: migrateStrategyOverwrite,
+			MigrateStrategyAppend:    migrateStrategyOverwrite,
```
erezrokah, line 54:
```suggestion
			MigrateStrategyAppend:    migrateStrategyAppend,
```

internal/memdb/memdb_test.go
```diff
@@ -39,7 +62,10 @@ func TestPluginManagedClientWithSmallBatchSize(t *testing.T) {
 				destination.WithDefaultBatchSize(1),
 				destination.WithDefaultBatchSizeBytes(1))
 		}, nil,
-		destination.PluginTestSuiteTests{})
+		destination.PluginTestSuiteTests{
+			MigrateStrategyOverwrite: migrateStrategyOverwrite,
+			MigrateStrategyAppend:    migrateStrategyOverwrite,
```
erezrokah, line 67:
```suggestion
			MigrateStrategyAppend:    migrateStrategyAppend,
```

internal/memdb/memdb_test.go
```diff
@@ -50,7 +76,10 @@ func TestPluginManagedClientWithLargeBatchSize(t *testing.T) {
 				destination.WithDefaultBatchSizeBytes(100000000))
 		},
 		nil,
-		destination.PluginTestSuiteTests{})
+		destination.PluginTestSuiteTests{
+			MigrateStrategyOverwrite: migrateStrategyOverwrite,
+			MigrateStrategyAppend:    migrateStrategyOverwrite,
```
erezrokah, line 81:
```suggestion
			MigrateStrategyAppend:    migrateStrategyAppend,
```

plugins/destination/plugin_testing_migrate.go
```diff
@@ -3,124 +3,281 @@ package destination
...
+	t *testing.T,
+	newPlugin NewPluginFunc,
+	logger zerolog.Logger,
+	spec specs.Destination,
+	strategy MigrateStrategy,
+) {
+	spec.BatchSize = 1
+
+	t.Run("add_column", func(t *testing.T) {
+		if strategy.AddColumn == specs.MigrateModeForced && spec.MigrateMode == specs.MigrateModeSafe {
```
erezrokah, line 90: I'm a bit confused by this, does this mean "Only run the test if it's expected to pass?". I mean `safe migrate mode declared in test && safe migrate mode declared in spec` or `forced migrate mode declared in test && forced migrate mode declared in spec`?

Don't we also want to test that there's an error when we have `spec safe mode` and can't migrate?

erezrokah (approved):
Added a few comments, I don't think any are blocking

schema/table.go
disq: `new` being a reserved word probably can't be used, but the `old` change is 👍

schema/table.go
disq: `Destructive` 😉

plugins/destination/plugin_testing_migrate.go
yevgenypats: I think for not we don't want to check for error yet as there are a lot of tests already (this will be `add_column_error` test varient). so I would give it some time to see if this approach is at all valid and then we can improve further.

schema/table.go
yevgenypats: Fixed.

schema/table.go
yevgenypats: This is a non-relevant comment - removed. I was just taking some notes and forgot to remove it.

plugins/destination/plugin_testing.go
```diff
@@ -17,9 +17,26 @@ type PluginTestSuite struct {
 	tests PluginTestSuiteTests
 }
 
+// type MigrateResult int
+
+// const (
+// 	MigrateResultSuccess MigrateResult = iota
+// 	MigrateResultDropTable
+// )
```
hermanschaaf, line 25: Looks like this can be removed?

plugins/destination/plugin_testing.go
```diff
@@ -17,9 +17,26 @@ type PluginTestSuite struct {
...
+
+// MigrateStrategy defines which tests we should include
+// true means test should succeed for either automigrate
```
hermanschaaf, line 28:
```suggestion
```

plugins/destination/plugin_testing_migrate.go
```diff
@@ -3,124 +3,281 @@ package destination
...
+	if mode == specs.MigrateModeSafe {
+		if len(resourcesRead) != 2 {
+			return fmt.Errorf("expected 2 resources after write, got %d", len(resourcesRead))
+		}
+		if diff := resourcesRead[1].Diff(resource2.Data); diff != "" {
+			return fmt.Errorf("resource2 diff: %s", diff)
+		}
+	} else {
+		if len(resourcesRead) != 1 {
+			return fmt.Errorf("expected 1 resources after write, got %d", len(resourcesRead))
```
hermanschaaf, line 69:
```suggestion
			return fmt.Errorf("expected 1 resource after write, got %d", len(resourcesRead))
```

plugins/destination/plugin_testing.go
```diff
@@ -17,9 +17,26 @@ type PluginTestSuite struct {
...
+type MigrateStrategy struct {
+	AddColumn           specs.MigrateMode
+	AddColumnNotNull    specs.MigrateMode
+	RemoveColumn        specs.MigrateMode
+	RemoveColumnNotNull specs.MigrateMode
+	ChangeColumn        specs.MigrateMode
+}
```
hermanschaaf, line 27: It might be just semantics, but I'm a bit confused by the options here 🤔 

The table in the PR description has two options for each entry: `Auto Migrate / Drop Table`. I think it would be clearer if we used that terminology here, rather than `MigrateMode = forced` or `MigrateMode = safe`. `forced/safe` doesn't convey the same meaning to me `drop/auto`, but I think that is the intention of this configuration, right?

plugins/destination/plugin_testing_migrate.go
```diff
@@ -3,124 +3,281 @@ package destination
...
+				},
+				{
+					Name: "bool",
+					Type: schema.TypeString,
+				},
+			},
+		}
+		p := newPlugin()
+		if err := testMigration(ctx, p, logger, spec, target, source, strategy.ChangeColumn); err != nil {
+			t.Fatalf("failed to migrate add_column: %v", err)
```
hermanschaaf, line 277:
```suggestion
			t.Fatalf("failed to migrate change_column: %v", err)
```

plugins/destination/plugin_testing_migrate.go
```diff
@@ -3,124 +3,281 @@ package destination
...
+			Columns: []schema.Column{
+				{
+					Name: "id",
+					Type: schema.TypeUUID,
+				},
+			},
+		}
+		p := newPlugin()
+		if err := testMigration(ctx, p, logger, spec, target, source, strategy.RemoveColumnNotNull); err != nil {
+			t.Fatalf("failed to migrate add_column: %v", err)
```
hermanschaaf, line 236:
```suggestion
			t.Fatalf("failed to migrate remove_column_not_null: %v", err)
```

hermanschaaf (approved):
Nice, looks good 👍 If we want to split hairs I'd probably change the naming of the options a bit to make them clearer (IMO), and I spotted a few error messages we may want to fix, but other than that all good :)

plugins/destination/plugin_testing.go
yevgenypats: Yeah I had the same dilemma if to introduce another type, but then it would be quite annoying as we will have `OverwriteForceStrategy` `OverwriteAutoMigrateStrategy` and the same for `append` so I preferred for now just to do one for each mode and then we can understand if it is being dropped or not.

But I think that yes, we potentially will need to change it, I just didn't want to make it even more complex right now as I think we will need to learn a bit more about this API and usage and then do another iteration on the testing API at least.

hermanschaaf: Maybe a comment would help for now. Something like `forced == drop table`, `safe == do not drop table`

PR #694, opened 2023-02-20
fix(test): Migrate testing fixes - WIP DON'T REVIEW

erezrokah:
#### Summary

Don't review, opening for reference. Will break into 3 PRs

---

Use the following steps to ensure your PR is ready to be reviewed

- [ ] Read the [contribution guidelines](../blob/main/CONTRIBUTING.md) 🧑‍🎓
- [ ] Run `go fmt` to format your code 🖊
- [ ] Lint your changes via `golangci-lint run` 🚨 (install golangci-lint [here](https://golangci-lint.run/usage/install/#local-installation))
- [ ] Update or add tests 🧪
- [ ] Ensure the status checks below are successful ✅

PR #696, opened 2023-02-20
fix(test-migrate): Ignore order when comparing resources read

erezrokah:
#### Summary

Extracted from https://github.com/cloudquery/plugin-sdk/pull/694.

`readAll` does not guarantee resources are returned in order of insertion. In other tests we sort the resources read by sync time, then by CqId (though not sure how sorting by CqId helps as it's random).
Since in the migrate tests we use the same sync time for both insertions using the same method as other tests doesn't work.

Instead this PR uses `testify` contains method to check that the new resource exists in the resources read (we already check the count a bit before that

---

- [ ] Update or add tests 🧪

### DIFF
--- a/plugins/destination/plugin_testing_migrate.go
+++ b/plugins/destination/plugin_testing_migrate.go
@@ -11,13 +11,14 @@ import (
 	"github.com/cloudquery/plugin-sdk/specs"
 	"github.com/google/uuid"
 	"github.com/rs/zerolog"
+	"github.com/stretchr/testify/require"
 )
 
 func tableUUIDSuffix() string {
 	return strings.ReplaceAll(uuid.NewString(), "-", "_")
 }
 
-func testMigration(ctx context.Context, p *Plugin, logger zerolog.Logger, spec specs.Destination, target *schema.Table, source *schema.Table, mode specs.MigrateMode) error {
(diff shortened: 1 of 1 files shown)


plugins/destination/plugin_testing_migrate.go
```diff
@@ -11,13 +11,14 @@ import (
...
 	"github.com/rs/zerolog"
+	"github.com/stretchr/testify/require"
 )
 
 func tableUUIDSuffix() string {
 	return strings.ReplaceAll(uuid.NewString(), "-", "_")
 }
 
-func testMigration(ctx context.Context, p *Plugin, logger zerolog.Logger, spec specs.Destination, target *schema.Table, source *schema.Table, mode specs.MigrateMode) error {
+func testMigration(ctx context.Context, t *testing.T, p *Plugin, logger zerolog.Logger, spec specs.Destination, target *schema.Table, source *schema.Table, mode specs.MigrateMode) error {
```
disq, line 21: If we're getting `*testing.T` as an input param, we don't need to return `error`. Could use `assert.Contains` and a "fake" testing.T (assert gets an interface with Errorf and nothing else) or just use a separate helper function below, or just remove the returned error and fail the test instead of returning error.

erezrokah: Ah, makes sense. How about I do it in a follow up PR to keep the diff on this one small?

disq (approved):

yevgenypats (approved):

PR #697, opened 2023-02-20
feat(test-migrate): Add `DataLossMode` enum

erezrokah:
#### Summary

Extracted from https://github.com/cloudquery/plugin-sdk/pull/694.

The current `MigrateStrategy` struct uses `MigrateMode` enum, however that only allows the following:
- Safe mode without any data loss
- Forced mode with table drop

If we only change a column type (without any other changes like adding not null), there's no reason to drop the table (in some destinations).
This PR adds an enum that lets destination plugins decide what kind of data loss is expected when doing a forced migrate.

See example https://github.com/cloudquery/cloudquery/pull/8255/files#diff-16b35c7fa05984f0e08cef54e2b7f6a4079474b4e6eb0a5a93e333ca54ed57e1R17

---

- [ ] Update or add tests 🧪

### DIFF
--- a/internal/memdb/memdb_test.go
+++ b/internal/memdb/memdb_test.go
@@ -14,19 +14,19 @@ import (
 )
 
 var migrateStrategyOverwrite = destination.MigrateStrategy{
-	AddColumn:           specs.MigrateModeForced,
-	AddColumnNotNull:    specs.MigrateModeForced,
-	RemoveColumn:        specs.MigrateModeForced,
-	RemoveColumnNotNull: specs.MigrateModeForced,
-	ChangeColumn:        specs.MigrateModeForced,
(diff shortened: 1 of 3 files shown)


yevgenypats:
Yes, I think we discussed it in the migration PR as well but in this case it is basically going to add a matrix as we now need to add a strategy for force and a strategy for safe (for both overwrite and append mode). we can do that but then we need to add those structs.

yevgenypats (changes requested):
Yes, I think we discussed it in the migration PR as well but in this case it is basically going to add a matrix as we now need to add a strategy for force and a strategy for safe (for both overwrite and append mode). we can do that but then we need to add those structs.

erezrokah:
Going to close this for now. We can add if we would like to support column drops as a part of a forced upgrade

PR #699, opened 2023-02-20
chore(test): Make destination plugin test options be hierarchical

candiduslynx:
When we define `SkipAppend` we want all `append` mode tests to be skipped. Same goes for `SkipOverwrite`


yevgenypats (changes requested):
I think it's just moving the problem from one place to another. The underlying problem is that the abstraction is leaky and it is not clear yet how to build it well. SkipAppend  so skip all append tests will be confusing in a similar way but no way to skip specific cases, so no need to do it right now as there is a work-around to achieve anything

PR #707, opened 2023-02-22
fix: Remove sentry sending from destinations

yevgenypats:
Right now the logic is not quite correct as if there are duplicate pks in the destination we should just filter but no reason to send to sentry. multiple sources can sync the same resources and it should be all good. We also check duplicates on the source level (and I think send to sentry) so no need to do it twice.

### DIFF
--- a/plugins/destination/managed_writer.go
+++ b/plugins/destination/managed_writer.go
@@ -69,7 +69,7 @@ func (p *Plugin) flush(ctx context.Context, metrics *Metrics, table *schema.Tabl
 	}
 }
 
-func (p *Plugin) removeDuplicatesByPK(table *schema.Table, resources [][]any) [][]any {
+func (*Plugin) removeDuplicatesByPK(table *schema.Table, resources [][]any) [][]any {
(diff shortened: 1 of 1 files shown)


candiduslynx:
I remember that we created this report when we didn't check duplicates on the source side.
Could you link a corresponding PR that implemented this change?

yevgenypats:
> I remember that we created this report when we didn't check duplicates on the source side. Could you link a corresponding PR that implemented this change?

If we don't have then we should only send it on the source side. either way this have to go.

hermanschaaf:
Agree it would be better on the source side, but I'm not sure how this can be implemented there? The only way I can think of would be to keep a map of seen PKs per table in the `Sync` server function, but this would add quite a lot of overhead. If it's already implemented, I'd like to see where this is done :)

candiduslynx (changes requested):
We don't send from the source side currently.
I don't think we should disable this on the destination side until source reporting is implemented.

candiduslynx:
Closing in favor of #760

PR #712, opened 2023-02-24
feat: Deterministic _cq_id

bbernays:
#### Summary

When making cq_id deterministic we cannot use any of the existing resolver types because we cannot guarantee that the value has been fetched by that time so in this PR I call a new function after the resource has been fully resolved in `resolveResource`

---

- [ ] Update or add tests 🧪


schema/resource.go
```diff
@@ -91,6 +95,32 @@ func (r *Resource) Columns() []string {
...
 
+func (r *Resource) CalculateUniqueValue() error {
+	names := r.Table.PrimaryKeys()
+	if len(names) == 0 {
+		names = r.Table.Columns.Names()
+	}
+	slices.Sort(names)
+	value := make([]any, len(names))
+	for i, name := range names {
+		value[i] = r.Get(name)
```
disq, line 106: It's also adding an empty `_cq_id` here I think. (also `_cq_parent_id` is still here, but I now think it should be included in the hash, so no worries there)

bbernays: But as long as it always adds the empty object it is fine

disq (approved):

candiduslynx (approved):
This will require a follow-up PR to overwrite `_cq_id` and `_cq_parent_id` where they've been skipped.

yevgenypats:
Looks good. Can we make it an opt-int option given we don't know the impact of this on compute at scale/real-world scenarios?

schema/resource.go
```diff
@@ -95,7 +95,15 @@ func (r *Resource) Columns() []string {
 	return r.Table.Columns.Names()
 }
 
-func (r *Resource) CalculateUniqueValue() error {
+func (r *Resource) CalculateUniqueValue(consistentID bool) error {
+	if !consistentID {
```
erezrokah, line 99: I think this means you'll need to create a separate function for it

erezrokah: I think it's a bit confusing to have a function named `CalculateUniqueValue` with a param to generate a non unique value

bbernays: Fair! I have updated the name to be `CalculateCQID`

schema/resource.go
```diff
@@ -91,6 +95,41 @@ func (r *Resource) Columns() []string {
...
+		if err != nil {
+			return err
+		}
+		return r.Set(CqIDColumn.Name, b)
+	}
+	names := r.Table.PrimaryKeys()
+	if len(names) == 0 {
+		names = r.Table.Columns.Names()
+	}
+	slices.Sort(names)
```
erezrokah, line 112: What's the reason we need to sort the PKs? It doesn't guarantee stability

schema/resource.go
```diff
@@ -91,6 +95,41 @@ func (r *Resource) Columns() []string {
 	return r.Table.Columns.Names()
 }
 
+//nolint:revive
+func (r *Resource) CalculateUniqueValue(consistentID bool) error {
```
erezrokah: Is there a reason this couldn't be implemented as a resolver?

bbernays: When making cq_id deterministic we cannot use any of the existing resolver types because we cannot guarantee that the value has been fetched by that time so in this PR I call a new function after the resource has been fully resolved in resolveResource

specs/source.go
```diff
@@ -48,6 +48,9 @@ type Source struct {
 	// Spec defines plugin specific configuration
 	// This is different in every source plugin.
 	Spec any `json:"spec,omitempty"`
+
+	// ConsistentID is a flag that indicates whether the source plugin should generate consistent IDs for resources.
+	ConsistentID bool `json:"consistent_id,omitempty"`
```
erezrokah, line 53:
```suggestion
	DeterministicCQID bool `json:"deterministic_cq_id,omitempty"`
```

Naming is hard, but maybe be more explicit of what this means

schema/resource.go
bbernays: It is an aspect of stability... If we rearrange columns (or support users rearranging) then we should not end up with a different hashed value...

specs/source.go
bbernays: Fixed. and changed all of the other places

schema/resource.go
```diff
@@ -91,6 +95,41 @@ func (r *Resource) Columns() []string {
...
 
+//nolint:revive
+func (r *Resource) CalculateCQID(deterministicCQID bool) error {
+	if !deterministicCQID {
+		uuidGen := uuid.New()
+		b, err := uuidGen.MarshalBinary()
+		if err != nil {
+			return err
+		}
+		return r.Set(CqIDColumn.Name, b)
```
candiduslynx, line 106: [nit]
We do the same `MarshalBinary` -> check `err` -> `return r.Set(CqIDColumn.Name, b)` at the end of the func.
Maybe we could extract the uuid func to return either `uuid.New()` or deterministic SHA1-based one and then just call `MarshalBinary` on the returned value?

candiduslynx (changes requested):
I'm OK with the changes. However, the problems with deterministic CQ ID for append mode in PG, MS SQL & other destinations should be addressed before this could be merged, hence `requesting changes` just to block the PR merge.

bbernays:
@candiduslynx- I don't think we need to fix those issues prior to releasing this because the default behavior hasn't changed. Users have to opt in to change the behavior.

yevgenypats (changes requested):
Sorry I missed it in prior reviews but we shouldn't use here hashstructure library. We already have a type system in place so no need for reflection. Either adding `Hash` function for each of them using `string` and hashing this with `io.writer` and `hasher` otherwise we don't really control the hashing mechanism also hashstructure will be much slower.

yevgenypats (approved):
Looks good.

Few non-blocking comments:
1) bytes.Buffer (instead of append)
2) the current slack thread - do we want to have random uuid for things without pk in any case.

yevgenypats (approved):

Issue #622, opened 2023-01-24
Destination plugins receive escaped data


candiduslynx:
### Describe the bug

JSON fields (and possibly others) receive escaped data, as was seen in https://github.com/cloudquery/cloudquery/issues/7001.
The reason for this is that source plugins use `json.Marshal` that does escaping, but the destination plugins can't perform unmarshaling due to lack of original structure.
So, the data received has unicode characters escaped, and it might be written to the destination in the same way.

### Expected Behavior

All values received by destination plugins should be unescaped.

### Steps to Reproduce

1. Create a tagged resource in Azure with key `MyKey` and value `A&B`.
2. Run the destination plugin in debugger
3. See that the value is received as `{"MyKey":"A\u0026B"}`, while it should've been presented as `{"MyKey":"A&B"}`

### Possible Solution

Add preliminary unescape for the data in the destination.

### Provider and CloudQuery version

latest

candiduslynx:
I suggest adding the Unicode characters to the generated input in the plugin testing for every applicable type.
This way we'll be able to see the data types that require unescaping.

candiduslynx:
Related to https://github.com/cloudquery/cloudquery/pull/7815

PR #714, opened 2023-02-27
fix(types-json): Disable HTML escaping during JSON marshalling

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/plugin-sdk/issues/622

See docs about the default behavior of `json.Marshal`:
https://pkg.go.dev/encoding/json#Marshal
image
See also https://pkg.go.dev/encoding/json#Encoder.SetEscapeHTML:
image

---

- [ ] Update or add tests 🧪

### DIFF
--- a/schema/json.go
+++ b/schema/json.go
@@ -124,13 +124,17 @@ func (dst *JSON) Set(src any) error {
 		return &ValidationError{Type: TypeJSON, Msg: "use pointer to JSON instead of value", Value: value}
 
 	default:
-		buf, err := json.Marshal(value)
+		buffer := &bytes.Buffer{}
+		encoder := json.NewEncoder(buffer)
+		encoder.SetEscapeHTML(false)
+		err := encoder.Encode(value)
 		if err != nil {
 			return err
 		}
 
+		buf := buffer.Bytes()
 		// For map and slice jsons, it is easier for users to work with '[]' or '{}' instead of JSON's 'null'.
-		if bytes.Equal(buf, []byte(`null`)) {
+		if bytes.Equal(buf, []byte("null\n")) {
 			if isEmptyStringMap(value) {
 				*dst = JSON{Bytes: []byte("{}"), Status: Present}
 				return nil
--- a/schema/json.go
+++ b/schema/json.go
@@ -132,9 +132,10 @@ func (dst *JSON) Set(src any) error {
 			return err
 		}
 
-		buf := buffer.Bytes()
+		// JSON encoder adds a newline to the end of the output that we don't want.
+		buf := bytes.TrimSuffix(buffer.Bytes(), []byte("\n"))
 		// For map and slice jsons, it is easier for users to work with '[]' or '{}' instead of JSON's 'null'.
-		if bytes.Equal(buf, []byte("null\n")) {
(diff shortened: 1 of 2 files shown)


schema/json_test.go
```diff
@@ -43,6 +43,8 @@ func TestJSONSet(t *testing.T) {
 		{source: map[string]Foo{}, result: JSON{Bytes: []byte(`{}`), Status: Present}},
 
 		{source: nil, result: JSON{Status: Null}},
+
+		{source: map[string]any{"test1": "a&b", "test2": "😀"}, result: JSON{Bytes: []byte(`{"test1": "a&b", "test2": "😀"}`), Status: Present}},
```
hermanschaaf, line 47: 😀

hermanschaaf: I assume this test failed before the fix? 🤔

schema/json.go
```diff
@@ -124,13 +124,17 @@ func (dst *JSON) Set(src any) error {
...
+		encoder.SetEscapeHTML(false)
+		err := encoder.Encode(value)
 		if err != nil {
 			return err
 		}
 
+		buf := buffer.Bytes()
 		// For map and slice jsons, it is easier for users to work with '[]' or '{}' instead of JSON's 'null'.
-		if bytes.Equal(buf, []byte(`null`)) {
+		if bytes.Equal(buf, []byte("null\n")) {
```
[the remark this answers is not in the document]
erezrokah: Existing tests are failing without it. I think the encoder handles `nil` values a bit different

schema/json_test.go
erezrokah: Yes, it failed due to the `&` character that was escaped. The emoji is somehow redundant for the purpose of the failure but I wanted to verify it doesn't get escaped too

schema/json.go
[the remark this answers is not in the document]
erezrokah: Here's why https://github.com/golang/go/blob/5e4e2acc8e53022f57398c4b0f3892f8113687f6/src/encoding/json/stream.go#L220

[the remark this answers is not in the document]
candiduslynx: I think that `null` should be preserved as is

schema/json_test.go
candiduslynx: Kindly add the test from https://github.com/cloudquery/cloudquery/pull/7815 as well

erezrokah: Can you please be specific on which test to add? I think with this fix:
1. Unescaping `{"BillTo":"P\u0026T"}` is not needed at the destination, as the cause was the redundant escaping fixed by this PR
2. `{"BillTo":"P&T"}` should be covered by the test here
3. Not sure why we need to unquote, is there a bug report I can use to reproduce?

candiduslynx: The unquote one. Basically, it was produced by parsing an AWS output, as the original field was a string representing the JSON contents, and it's returned from the API/SDK escaped/quoted

candiduslynx (approved):

hermanschaaf (approved):

yevgenypats (approved):

candiduslynx (approved):

PR #724, opened 2023-03-06
feat: Add arrow support for timestamp and bytea

yevgenypats:
This should unblock https://github.com/cloudquery/filetypes/pull/87

Add supports for timestamp to decode arrow string and for byte to decode base64 (backward compat is saved)


disq (approved):

testdata/testdata.go
```diff
@@ -183,7 +183,7 @@ func GenTestData(table *schema.Table) schema.CQTypes {
 			data[i] = uuidArrayColumn
 		case schema.TypeInet:
 			inetColumn := &schema.Inet{}
-			if err := inetColumn.Set("192.0.2.1/24"); err != nil {
+			if err := inetColumn.Set("192.0.2.0/24"); err != nil {
```
erezrokah, line 186: What's the reason for this change?

schema/resource_test.go
```diff
@@ -438,7 +438,7 @@ var calculateCQIDPrimaryKeyTestCases = []struct {
 			"MacAddressValue":      "aa:bb:cc:dd:ee:ff",
 			"MacAddressArrayValue": []string{"aa:bb:cc:dd:ee:ff", "11:22:33:44:55:66"},
 		},
-		ExpectedValue:     &UUID{Bytes: [16]uint8{0xa6, 0x76, 0x76, 0xfb, 0x4c, 0x95, 0x51, 0x2d, 0x9e, 0xcd, 0xf4, 0xc4, 0xae, 0xc1, 0x2a, 0xf5}, Status: 0x2},
+		ExpectedValue:     &UUID{Bytes: [16]uint8{0x82, 0xaa, 0xa2, 0xc7, 0x75, 0x10, 0x5c, 0x6d, 0xb8, 0xef, 0x18, 0x49, 0x33, 0x99, 0x4a, 0x9d}, Status: 0x2},
```
erezrokah, line 441: If this PR is backwards compatible I would expect existing tests to pass. Is this related to the change in `testdata/testdata.go`?

schema/bytea.go
```diff
@@ -41,7 +42,7 @@ func (dst *Bytea) Equal(src CQType) bool {
 
 func (dst *Bytea) String() string {
 	if dst.Status == Present {
-		return hex.EncodeToString(dst.Bytes)
```
erezrokah, line 44: Where do we use this `String` method? Is this not a breaking change as we return a different format of the string?
Trying to understand the impact of this

erezrokah (changes requested):
Code looks good, I have a couple of questions on the impact of the changes

testdata/testdata.go
hermanschaaf: I guess it's because the bitmask filters out the final `.1`, so it amounts to the same, but if Arrow rejects this type of IP we may need to handle it better

erezrokah:
Unrelated but I've disabled the codecov status checks in https://app.codecov.io/account/gh/cloudquery/yaml/

testdata/testdata.go
yevgenypats: I think `192.0.2.1./24` is basically the same as `192.0.2.0/24` after applying the mask.

schema/resource_test.go
yevgenypats: I think it is backward compat the the binary can read older encoding but the encoding for the binary types moving forward will be base64. I don't think it should impact anything though.

testdata/testdata.go
hermanschaaf: @yevgenypats Right, but was this change necessary for the tests to pass, or was it just some cleanup?

schema/bytea.go
yevgenypats: Yeah it is a breaking change (I did add a way for us to support both conversion in `Set`) but yes moving forward we will use base64 encoding for bytea to work like arrow. I can add a new type but I don't think we have anything that depends on that.

erezrokah (approved):
Approving with the concern that we needed to update test data to make the tests pass. I would expect it to pass with existing data.
If we don't expect to receive data in the old format I guess this is ok

testdata/testdata.go
yevgenypats: maybe Im confusing something about inet vs cidr . but I did need it for a test to pass because our own arrow `inet` converted it to `192.0.2.0/24` but I think this happened just from the original `IPNet.String` conversion so I suspect it should affect anything but maybe I missed something and didn't get to the root cause of this, not sure.

PR #729, opened 2023-03-10
feat: Add ability to store table titles and render them in documentation

hermanschaaf:
This adds the ability to set titles for tables, and starts rendering them in the documentation. 

For example, docs for an example AWS table will now be rendered as:

```
# Table: aws_accessanalyzer_analyzer_findings

This table shows data for AWS Access Analyzer Analyzer Findings.
```

This change should help human readers more quickly parse what a table is about, and in some cases developers may want to change the `Title` attribute to make it more descriptive. It is also useful for search, as many search engines will not tokenize on the underscores, leaving these pages excluded from relevant searches.

### DIFF
--- a/caser/caser.go
+++ b/caser/caser.go
@@ -26,7 +26,7 @@ func WithCustomInitialisms(fields map[string]bool) Option {
 }
 
 // WithCustomExceptions allows to specify custom exceptions for caser.
-// The parameter is a map of camel:snake values like map[string]string{"oauth":"OAuth"}
+// The parameter is a map of snake:camel values like map[string]string{"oauth":"OAuth"}
 func WithCustomExceptions(fields map[string]string) Option {
 	return func(c *Caser) {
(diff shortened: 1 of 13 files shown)


erezrokah (approved):

hermanschaaf:
Having `WithTitleTransformer` as an option on the Plugin itself is perhaps not the best. Maybe it should be a property on the table instead? We can then still apply it at the plugin level, if we want, by doing something like:

```
func awsTitleTransfomer...

func AWS() *source.Plugin {
        tbls := tables()
        for i := range tbls {
              tbls[i].TitleTransformer = awsTitleTransformer
        }
	return source.NewPlugin(
		"aws",
		Version,
		tbls,
		client.Configure,
	)
}

```

Happy to change it if the general consensus is that this would be better

yevgenypats (approved):

PR #732, opened 2023-03-14
feat: Support for User Specifying Primary Key Scheme (default or cq-ids)

bbernays:
#### Summary

Initial implementation for https://github.com/cloudquery/cloudquery/issues/8811

---

- [ ] Update or add tests 🧪

### DIFF
--- a/plugins/destination/plugin.go
+++ b/plugins/destination/plugin.go
@@ -310,15 +310,14 @@ func setCqIDColumnOptionsForTables(tables []*schema.Table) {
 func setCQIDAsPrimaryKeysForTables(tables schema.Tables) {
 	for _, table := range tables {
 		for i, col := range table.Columns {
-			if col.Name == schema.CqIDColumn.Name {
+			switch col.Name {
+			case schema.CqIDColumn.Name:
 				table.Columns[i].CreationOptions.PrimaryKey = true
-				continue
+			case schema.CqParentIDColumn.Name:
+				table.Columns[i].CreationOptions.PrimaryKey = table.Parent != nil
+			default:
+				table.Columns[i].CreationOptions.PrimaryKey = false
 			}
-			if table.Parent != nil && col.Name == schema.CqParentIDColumn.Name {
-				table.Columns[i].CreationOptions.PrimaryKey = true
-				continue
-			}
-			table.Columns[i].CreationOptions.PrimaryKey = false
 		}
 		setCQIDAsPrimaryKeysForTables(table.Relations)
 	}
--- a/plugins/destination/plugin.go
+++ b/plugins/destination/plugin.go
@@ -314,12 +314,7 @@ func (p *Plugin) setPKsForTables(tables schema.Tables) {
 func setCQIDAsPrimaryKeysForTables(tables schema.Tables) {
 	for _, table := range tables {
 		for i, col := range table.Columns {
-			switch col.Name {
-			case schema.CqIDColumn.Name:
-				table.Columns[i].CreationOptions.PrimaryKey = true
-			default:
-				table.Columns[i].CreationOptions.PrimaryKey = false
-			}
+			table.Columns[i].CreationOptions.PrimaryKey = col.Name == schema.CqIDColumn.Name
 		}
 		setCQIDAsPrimaryKeysForTables(table.Relations)
 	}
--- a/plugins/destination/plugin.go
+++ b/plugins/destination/plugin.go
@@ -317,8 +317,6 @@ func setCQIDAsPrimaryKeysForTables(tables schema.Tables) {
 			switch col.Name {
 			case schema.CqIDColumn.Name:
 				table.Columns[i].CreationOptions.PrimaryKey = true
-			case schema.CqParentIDColumn.Name:
-				table.Columns[i].CreationOptions.PrimaryKey = table.Parent != nil
 			default:
 				table.Columns[i].CreationOptions.PrimaryKey = false
 			}
--- a/plugins/destination/plugin.go
+++ b/plugins/destination/plugin.go
@@ -320,5 +320,6 @@ func setCQIDAsPrimaryKeysForTables(tables schema.Tables) {
 			}
 			table.Columns[i].CreationOptions.PrimaryKey = false
 		}
+		setCQIDAsPrimaryKeysForTables(table.Relations)
 	}
 }
--- a/plugins/destination/plugin.go
+++ b/plugins/destination/plugin.go
@@ -187,9 +187,7 @@ func (p *Plugin) Init(ctx context.Context, logger zerolog.Logger, spec specs.Des
 func (p *Plugin) Migrate(ctx context.Context, tables schema.Tables) error {
 	SetDestinationManagedCqColumns(tables)
 	setCqIDColumnOptionsForTables(tables)
-	if p.spec.PKMode == specs.PKModeCQID {
-		setCQIDAsPrimaryKeysForTables(tables)
-	}
+	p.setPKsForTables(tables)
 	return p.client.Migrate(ctx, tables)
 }
 
@@ -245,6 +243,7 @@ func (p *Plugin) writeAll(ctx context.Context, sourceSpec specs.Source, tables s
 func (p *Plugin) Write(ctx context.Context, sourceSpec specs.Source, tables schema.Tables, syncTime time.Time, res <-chan schema.DestinationResource) error {
 	syncTime = syncTime.UTC()
 	SetDestinationManagedCqColumns(tables)
+	p.setPKsForTables(tables)
 	switch p.writerType {
 	case unmanaged:
 		if err := p.writeUnmanaged(ctx, sourceSpec, tables, syncTime, res); err != nil {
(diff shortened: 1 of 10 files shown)


erezrokah (commented):
I know this is still in draft, but maybe since PKs and columns are defined on sources, this should be a source option?

I mean, what if I want `_cq_id` PK for AWS, and regular PKs for Azure?

bbernays:
> I know this is still in draft, but maybe since PKs and columns are defined on sources, this should be a source option?
> 
> I mean, what if I want `_cq_id` PK for AWS, and regular PKs for Azure?

I am actually against putting it on the source because then the user will have to fetch the same data multiple times if they want to put the same data in multiple destinations (with different PK settings). In the case where a user has multiple sources and a single destination but want different behavior for each source, they should duplicate the destination.

erezrokah:
> I am actually against putting it on the source because then the user will have to fetch the same data multiple times if they want to put the same data in multiple destinations (with different PK settings). In the case where a user has multiple sources and a single destination but want different behavior for each source, they should duplicate the destination.

Got it, makes sense

disq (approved):
LGTM, looked for a simpler approach but I don't see it

specs/pk_mode_test.go
```diff
@@ -0,0 +1,50 @@
...
+		t.Fatalf("expected:\"cq-id\" got:%s", string(pkModeStr))
+	}
+
+	pkMode = PKModeCompositeKeys
+	if pkModeStr, err := pkMode.MarshalJSON(); err != nil {
+		t.Fatal(err)
+	} else if string(pkModeStr) != `"composite-keys"` {
+		t.Fatalf("expected:\"cq-id\" got:%s", string(pkModeStr))
+	}
+
```
erezrokah, line 49:
```suggestion
```

schema/table.go
```diff
@@ -314,6 +316,7 @@ func (t *Table) GetChanges(old *Table) []TableColumnChange {
 			})
 		}
 	}
+	// Remove columns that have been deleted from table definition.
```
erezrokah, line 319:
```suggestion
	// A column was removed from the table definition
```

schema/table.go
```diff
@@ -297,6 +297,7 @@ func (t *Table) GetChanges(old *Table) []TableColumnChange {
 	var changes []TableColumnChange
 	for _, c := range t.Columns {
 		otherColumn := old.Columns.Get(c.Name)
+		// Add in a column that didn't exist before.
```
erezrokah, line 300:
```suggestion
		// A column was added to the table definition
```

schema/table.go
```diff
@@ -305,6 +306,7 @@ func (t *Table) GetChanges(old *Table) []TableColumnChange {
 			})
 			continue
 		}
+		// column type changed or other column options changed (not null or PK)
```
erezrokah, line 309:
```suggestion
		// Column type or options (e.g. PK, Not Null) changed in the new table definition
```

specs/pk_mode.go
```diff
@@ -0,0 +1,50 @@
...
+
+type PKMode int
+
+const (
+	PKModeCompositeKeys PKMode = iota
+	PKModeCQID
+)
+
+var (
+	pkModeStrings = []string{"composite-keys", "cq-id"}
```
erezrokah, line 17:
```suggestion
	pkModeStrings = []string{"composite-keys", "cq-id"}
```

I'm not sure `composite-keys` is accurate when there's a single PK.
Maybe call these `default` and `cq-id-only`?

erezrokah (approved):
Added a few comments (non blocking)

plugins/destination/plugin.go
```diff
@@ -303,3 +306,20 @@ func setCqIDColumnOptionsForTables(tables []*schema.Table) {
...
+		for i, col := range table.Columns {
+			if col.Name == schema.CqIDColumn.Name {
+				table.Columns[i].CreationOptions.PrimaryKey = true
+				continue
+			}
+			if table.Parent != nil && col.Name == schema.CqParentIDColumn.Name {
+				table.Columns[i].CreationOptions.PrimaryKey = true
+				continue
+			}
+			table.Columns[i].CreationOptions.PrimaryKey = false
```
candiduslynx, line 321:
```suggestion
			switch col.Name {
			case schema.CqIDColumn.Name:
				table.Columns[i].CreationOptions.PrimaryKey = true
			case schema.CqParentIDColumn.Name:
				table.Columns[i].CreationOptions.PrimaryKey = table.Parent != nil
			default:
				table.Columns[i].CreationOptions.PrimaryKey = false
			}
```

candiduslynx (approved):

yevgenypats (approved):

plugins/destination/plugin.go
```diff
@@ -187,6 +187,9 @@ func (p *Plugin) Init(ctx context.Context, logger zerolog.Logger, spec specs.Des
 func (p *Plugin) Migrate(ctx context.Context, tables schema.Tables) error {
 	SetDestinationManagedCqColumns(tables)
 	setCqIDColumnOptionsForTables(tables)
+	if p.spec.PKMode == specs.PKModeCQID {
+		setCQIDAsPrimaryKeysForTables(tables)
```
yevgenypats, line 191: Should that also be called on `Write`?

yevgenypats (changes requested):
I think `setCQIDAsPrimaryKeysForTables` might need to be called also on destination `Write` . We have e2e test suites in the SDK under https://github.com/cloudquery/plugin-sdk/tree/main/internal/memdb we might want to see that this test is covered there as well.

plugins/destination/plugin_testing_migrate.go
```diff
@@ -74,6 +74,17 @@ func testMigration(ctx context.Context, t *testing.T, p *Plugin, logger zerolog.
...
 
+	if p.spec.PKMode == specs.PKModeCQID {
+		for _, tColumn := range target.Columns {
+			if tColumn.Name != schema.CqIDColumn.Name {
+				continue
+			}
+			if !tColumn.CreationOptions.PrimaryKey {
+				return fmt.Errorf("expected primary key on %s", tColumn.Name)
+			}
+		}
```
candiduslynx, line 85:
```suggestion
		col := target.Get(schema.CqIDColumn.Name)
		if col == nil {
			return fmt.Errorf("expected to have %s column", schema.CqIDColumn.Name)
		}
		if !col.CreationOptions.PrimaryKey {
			return fmt.Errorf("expected primary key on %s", col.Name)
		}
```

candiduslynx (changes requested):

plugins/destination/plugin_testing_migrate.go
bbernays: Updated the test to fail if there is a PK on a different column than the cq_id

plugins/destination/plugin_testing_migrate.go
```diff
@@ -74,6 +74,17 @@ func testMigration(ctx context.Context, t *testing.T, p *Plugin, logger zerolog.
...
 	}
 
+	if p.spec.PKMode == specs.PKModeCQID {
+		for _, tColumn := range target.Columns {
+			if tColumn.Name != schema.CqIDColumn.Name && tColumn.CreationOptions.PrimaryKey {
+				return fmt.Errorf("unexpected primary key on %s", tColumn.Name)
+			}
+			if tColumn.Name == schema.CqIDColumn.Name && !tColumn.CreationOptions.PrimaryKey {
+				return fmt.Errorf("expected primary key on %s", tColumn.Name)
+			}
```
candiduslynx, line 84:
```suggestion
			if tColumn.CreationOptions.PrimaryKey != tColumn.Name ==  schema.CqIDColumn.Name {
				return fmt.Errorf("primary key = %t for column %s", tColumn.CreationOptions.PrimaryKey, tColumn.Name)
			}
```

plugins/destination/plugin.go
```diff
@@ -303,3 +305,24 @@ func setCqIDColumnOptionsForTables(tables []*schema.Table) {
...
+	}
+}
+func setCQIDAsPrimaryKeysForTables(tables schema.Tables) {
+	for _, table := range tables {
+		for i, col := range table.Columns {
+			switch col.Name {
+			case schema.CqIDColumn.Name:
+				table.Columns[i].CreationOptions.PrimaryKey = true
+			case schema.CqParentIDColumn.Name:
+				table.Columns[i].CreationOptions.PrimaryKey = table.Parent != nil
```
yevgenypats, line 321: Why a parent should ever be a primary key ?

bbernays: 2 different top level resources could resolve the same nested resource. In that case, the `cq_id` would be the same which means that one of the resources would be dropped

bbernays: I have removed this because the `table.Parent` is always nil

plugins/destination/plugin.go
```diff
@@ -303,3 +305,22 @@ func setCqIDColumnOptionsForTables(tables []*schema.Table) {
...
+}
+func setCQIDAsPrimaryKeysForTables(tables schema.Tables) {
+	for _, table := range tables {
+		for i, col := range table.Columns {
+			switch col.Name {
+			case schema.CqIDColumn.Name:
+				table.Columns[i].CreationOptions.PrimaryKey = true
+			default:
+				table.Columns[i].CreationOptions.PrimaryKey = false
+			}
```
candiduslynx, line 322:
```suggestion
			table.Columns[i].CreationOptions.PrimaryKey = col.Name == schema.CqIDColumn.Name
```

candiduslynx (approved):
also a non-blocking nit

yevgenypats (approved):

PR #737, opened 2023-03-23
feat: Arrow migration for destination

yevgenypats:
This is ready for first review with tests passing. might need some more manual testing and maybe migrating a plugin like mongo/postgres and testing with source  like aws,gcp to stress test common use-cases and not only internal memdb


erezrokah (commented):
As an initial PR this looks good, I'll do a proper review once it ready

plugins/destination/managed_writer.go
```diff
@@ -32,31 +32,27 @@ func (p *Plugin) worker(ctx context.Context, metrics *Metrics, table *schema.Tab
 				}
 				return
 			}
-			if len(resources) == p.spec.BatchSize || sizeBytes+r.Size() > p.spec.BatchSizeBytes {
+			if len(resources) == p.spec.BatchSize {
```
hermanschaaf, line 35: We seem to be missing the condition for flushing on byte size here, compared to what it was before. Is this intentional?

plugins/destination/managed_writer.go
```diff
@@ -150,17 +145,9 @@ func (p *Plugin) writeManagedTableBatch(ctx context.Context, sourceSpec specs.So
...
-	syncTimeColumn := &schema.Timestamptz{}
-	_ = syncTimeColumn.Set(syncTime)
 	for r := range res {
-		// this is a check to keep backward compatible for sources that are not adding
-		// source and sync time
-		if len(r.Data) < len(tables.Get(r.TableName).Columns) {
-			r.Data = append([]schema.CQType{sourceColumn, syncTimeColumn}, r.Data...)
-		}
-		workers[r.TableName].ch <- r.Data
+		tableName := r.Schema().Metadata().Values()[0]
```
hermanschaaf, line 149:
```suggestion
		md := r.Schema().Metadata()
		nameIndex := md.FindKey("table")
		tableName := md.Values()[nameIndex]
```

Let's use `FindKey` here so that we can add more metadata later without the possibility of it causing a bug. We may also want to suggest adding some more methods to the `Metadata` struct so we don't have to do this everywhere

plugins/destination/plugin_testing.go
```diff
@@ -62,6 +66,29 @@ type PluginTestSuiteTests struct {
...
+func RecordDiff(l arrow.Record, r arrow.Record) string {
+	var sb strings.Builder
+	if l.NumCols() != r.NumCols() {
+		return fmt.Sprintf("different number of columns: %d vs %d", l.NumCols(), r.NumCols())
+	}
+	if l.NumRows() != r.NumRows() {
+		return fmt.Sprintf("different number of rows: %d vs %d", l.NumRows(), r.NumRows())
+	}
+	for i := 0; i < int(l.NumCols()); i++ {
+		s, err := array.DiffString(l.Column(i), r.Column(i), memory.DefaultAllocator)
```
hermanschaaf, line 78: Just FYI, based on the review I think this function signature will change. But probably not by much.

plugins/destination/plugin_testing.go
```diff
@@ -170,40 +197,10 @@ func PluginTestSuiteRunner(t *testing.T, newPlugin NewPluginFunc, destSpec specs
...
-		if !resources[i].Data[syncTimeIndex].Equal(resources[j].Data[syncTimeIndex]) {
-			return resources[i].Data[syncTimeIndex].Get().(time.Time).Before(resources[j].Data[syncTimeIndex].Get().(time.Time))
-		}
-		return resources[i].Data[cqIDIndex].String() < resources[j].Data[cqIDIndex].String()
-	})
-}
-
-func sortCQTypes(table *schema.Table, resources []schema.CQTypes) {
-	cqIDIndex := table.Columns.Index(schema.CqIDColumn.Name)
+func sortRecordsBySyncTime(table *schema.Table, records []arrow.Record) {
```
hermanschaaf, line 200: This assumes that the records each only contain a single row, right? I guess it's fine for testing

plugins/destination/plugin_testing_overwrite.go
```diff
@@ -32,40 +34,39 @@ func (*PluginTestSuite) destinationPluginTestWriteOverwrite(ctx context.Context,
...
-	updatedResource := schema.DestinationResource{
-		TableName: table.Name,
-		Data:      make(schema.CQTypes, len(resources[0].Data)),
-	}
-	copy(updatedResource.Data, resources[0].Data)
-	_ = updatedResource.Data[1].Set(secondSyncTime)
-
+	u := resources[0].Column(2).(*types.UUIDArray).Value(0)
+	us := u.String()
+	fmt.Println(us)
```
hermanschaaf, line 68:
```suggestion
```

I think this print line can be removed

plugins/destination/managed_writer.go
hermanschaaf: Oh I see we have a `TableNameFromSchema` function, maybe we can use that here?

testdata/testdata.go
```diff
@@ -105,121 +110,101 @@ func TestTable(name string) *schema.Table {
...
-			} else {
-				data[i] = &schema.Text{
-					Str:    "test",
-					Status: schema.Present,
+			} else if arrow.TypeEqual(c.Type, arrow.BinaryTypes.Binary) {
+				bldr.Field(i).(*array.BinaryBuilder).Append([]byte{1, 2, 3})
+			} else if arrow.TypeEqual(c.Type, arrow.ListOf(arrow.BinaryTypes.String)) {
+				if c.Name == "text_array_with_null" {
+					bldr.Field(i).(*array.ListBuilder).Append(true)
+					bldr.Field(i).(*array.ListBuilder).ValueBuilder().(*array.StringBuilder).AppendString("test1")
```
hermanschaaf, line 143: I think these ValueBuilders need to be `Released` to avoid memory leaks 🤔 Which makes me think, we may want to use `CheckedAllocator` in tests?

plugins/destination/managed_writer.go
yevgenypats: There is no `record.Len` method but I'll add it right now (it exist in the python implementation and should be relatively straight forward)

internal/memdb/memdb.go
```diff
@@ -154,17 +161,22 @@ func (c *client) Write(ctx context.Context, tables schema.Tables, resources <-ch
...
-			c.overwrite(tables.Get(resource.TableName), resource.Data)
+			c.overwrite(tables.Get(tableName), resource)
 		}
 		c.memoryDBLock.Unlock()
 	}
 	return nil
 }
 
-func (c *client) WriteTableBatch(ctx context.Context, table *schema.Table, resources [][]any) error {
+func (c *client) WriteTableBatch(ctx context.Context, table *schema.Table, resources []arrow.Record) error {
```
hermanschaaf, line 179: Should this signature use `resources arrow.Record`, rather than `resources []arrow.Record`? I think it's clearer and easier to handle on the destination side if it's a single record (which would contain multiple rows of the same schema)

yevgenypats: I think we want to enable `[]arrow.Record` i.e both array of records and arrow batches. depending on the source the plugin would group them together that the total records is X. To avoid copying and creating one `record` with all the batches I think it's good to keep it this way for now.

hermanschaaf: Isn't the writing of the record in this case handled by the SDK? I'm thinking the source would send on the results channel, the SDK would then append to the current record builder, and when the batch is full the record will be created and sent to the destination. Not sure why we would need multiple records in this scenario.

plugins/destination/managed_writer.go
yevgenypats: Added `GetValue` to `MetaData` via https://github.com/cloudquery/arrow/pull/4

go.mod
```diff
@@ -26,6 +28,25 @@ require (
...
+	github.com/minio/asm2plan9s v0.0.0-20200509001527-cdd76441f9d8 // indirect
+	github.com/minio/c2goasm v0.0.0-20190812172519-36a3d3bbc4f3 // indirect
+	github.com/zeebo/xxh3 v1.0.2 // indirect
+	golang.org/x/mod v0.8.0 // indirect
+	golang.org/x/tools v0.6.0 // indirect
+	golang.org/x/xerrors v0.0.0-20220609144429-65e65417b02f // indirect
+)
+
+// replace github.com/apache/arrow/go/v12 => github.com/cloudquery/arrow/go/v12 v12.0.0-20230403113355-afc433992be2
+replace github.com/apache/arrow/go/v12 => ../arrow/go
```
hermanschaaf, line 48: We will need to update this before merging. @yevgenypats what changes / branch are you using to test this locally?

hermanschaaf: Oh, seems like it's `cqmain1`

plugins/destination/plugin.go
```diff
@@ -184,66 +187,62 @@ func (p *Plugin) Init(ctx context.Context, logger zerolog.Logger, spec specs.Des
 }
 
 // we implement all DestinationClient functions so we can hook into pre-post behavior
-func (p *Plugin) Migrate(ctx context.Context, tables schema.Tables) error {
-	SetDestinationManagedCqColumns(tables)
```
erezrokah, line 188: So is the expectation for the internal columns to be set in source plugins? I think that's fine, just means will need to update all sources before releasing any destinations

schema/arrow.go
```diff
@@ -0,0 +1,393 @@
...
+	}
+	if opts.Unique {
+		keys = append(keys, MetadataUnique)
+		values = append(values, MetadataTrue)
+	}
+
+	return arrow.NewMetadata(keys, values)
+}
+
+func MdIsPk(md arrow.Metadata) bool {
```
erezrokah, line 65:
```suggestion
func MetadataIsPk(md arrow.Metadata) bool {
```

Maybe easier to understand like this? Not sure if we can embed the struct and add the method like that

schema/arrow.go
```diff
@@ -0,0 +1,393 @@
...
+	pk, ok := md.GetValue(MetadataPrimaryKey)
+	return ok && pk == MetadataTrue || pk == MetadataTrue
+}
+
+func MdIsUnique(md arrow.Metadata) bool {
```
erezrokah, line 70:
```suggestion
func MetadataIsUnique(md arrow.Metadata) bool {
```

erezrokah (approved):
This is quite a bit of code, but I think I get the general idea of replacing our type system with Arrow's + some extensions. 

Might be better to merge and work of `main` so we can test thoroughly.

Had some comment, the most important (non blocking one) is about the internal columns

erezrokah:
We also need to fix the CI and tests

plugins/destination/plugin.go
yevgenypats: I think we already do that on the source side (if not we might want to keep it here but I believe we changed it long time a go and this is why I add an error if it doesn't happen)

go.mod
hermanschaaf: I've updated the `cqmain` branch and update the go.mod reference here

yevgenypats: Seems it's missing `array.DiffString` somehow. maybe we need to rebase `cqmain` on top of `main`

hermanschaaf: I think I've resolved that error now (maybe make sure you've pulled the latest), but I'm seeing some new errors that we're also getting in CI:

```
plugins/destination/plugin.go:191:11: cannot use p.setPKsForTables(tables) (value of type schemas.Schemas) as schema.Schemas value in assignment
```

hermanschaaf: Oh also `array.DiffString` is gone I see yeah, that name just needs to change (I'll fix it)

schema/arrow.go
yevgenypats: I removed this function.

PR #738, opened 2023-03-27
feat: Add the names of tables to the periodic logger

shimonp21:
if the number of running tables is small (less than e.g. 30), log all their names in the periodic 'sync in progress' logs.

Example log lines:

```
2023-03-27T08:11:52Z INF Sync in progress module=gcp-src num_in_progress_tables=160 total_errors=0 total_panics=0 total_resources=81
2023-03-27T08:11:53Z INF Sync in progress module=gcp-src num_in_progress_tables=88 total_errors=2 total_panics=0 total_resources=1240
2023-03-27T10:48:27Z INF Sync in progress in_progress_tables=["gcp_compute_zones"] module=gcp-src num_in_progress_tables=1 total_errors=0 total_panics=0 total_resources=5837
```

#### Summary

---

- [ ] Update or add tests 🧪


plugins/source/metrics.go
```diff
@@ -123,3 +124,42 @@ func (s *Metrics) TotalResourcesAtomic() uint64 {
 	}
 	return total
 }
+
+func (s *Metrics) MarkStart(table *schema.Table, clientID string) {
```
erezrokah, line 128: Is this code be thread safe? It's accessed from different Go routines

plugins/source/metrics.go
```diff
@@ -123,3 +124,42 @@ func (s *Metrics) TotalResourcesAtomic() uint64 {
...
+}
+
+func (s *Metrics) markEndRecursive(table *schema.Table, clientID string) {
+	s.TableClient[table.Name][clientID].EndTime = time.Now()
+	for _, relation := range table.Relations {
+		s.markEndRecursive(relation, clientID)
+	}
+}
+
+func (s *Metrics) InProgressTables() []string {
```
erezrokah, line 150: I would say it's better to print "tables in progress longer that X time", otherwise we will print tables that are short to sync along with tables that are long to sync

erezrokah (changes requested):
Thanks for the PR @shimonp21, added a couple of comments

plugins/source/metrics.go
shimonp21: Aaaaa good point

plugins/source/metrics.go
shimonp21: I think we can leave this optimization out of the first version of this PR, because most tables start at the same time in almost all cases.
Also, the logggin code makes sure to only log.

This PR tries to help with the annoying pattern of cloudquery where one (or couple) of tables take 10X or more time than all the rest - and this should be way more visible with this change.

plugins/source/metrics.go
```diff
@@ -123,3 +130,51 @@ func (s *Metrics) TotalResourcesAtomic() uint64 {
...
+	} else {
+		s.TableClient[table.Name][clientID].Mutex.Lock()
+		defer s.TableClient[table.Name][clientID].Mutex.Unlock()
+		s.TableClient[table.Name][clientID].EndTime = time.Now()
+	}
+}
+
+func (s *Metrics) markEndRecursive(table *schema.Table, clientID string) {
+	// We don't use defer with Unlock(), because we want to unlock the mutex as soon as possible.
+	s.TableClient[table.Name][clientID].Mutex.Lock()
```
erezrokah, line 156: Maybe we should lift the locking here to `MarkEnd`? I'm not sure we'd like to have interleaving locks and updating the times is not an expensive operation.

plugins/source/metrics.go
```diff
@@ -123,3 +130,51 @@ func (s *Metrics) TotalResourcesAtomic() uint64 {
 	}
 	return total
 }
+
+func (s *Metrics) MarkStart(table *schema.Table, clientID string) {
+	s.TableClient[table.Name][clientID].Mutex.Lock()
```
erezrokah, line 135: We might want to do the start time calculation before the lock, WDYT? That would be more accurate to when the method was called.

Same for end time calculation (and maybe even pass that value to relations)

plugins/source/scheduler_dfs.go
```diff
@@ -98,6 +98,9 @@ func (p *Plugin) syncDfs(ctx context.Context, spec specs.Source, client schema.C
 }
 
 func (p *Plugin) resolveTableDfs(ctx context.Context, table *schema.Table, client schema.ClientMeta, parent *schema.Resource, resolvedResources chan<- *schema.Resource, depth int) {
+	p.metrics.MarkStart(table, client.ID())
+	defer p.Metrics().MarkEnd(table, client.ID())
+
 	var validationErr *schema.ValidationError
 	clientName := client.ID()
```
erezrokah, line 105: Maybe we could move `clientName` up and use it for the metrics methods too?

erezrokah (commented):
Thanks for the follow up @shimonp21, added a few more comments

plugins/source/metrics.go
shimonp21: Added a mutex :)
We can have a different mutex per object["table"]["clientId"], because the map is never read/written concurrently (it is pre-initialized, and only the properties of the object are changed during runtime).

Thanks again for the catch, embarassing!

plugins/source/metrics.go
```diff
@@ -1,22 +1,28 @@
...
 	TableClient map[string]map[string]*TableClientMetrics
 }
 
 type TableClientMetrics struct {
+	// These should only be accessed with 'Atomic*' methods
 	Resources uint64
 	Errors    uint64
 	Panics    uint64
+
+	// These accesses must be proteceted by the mutex.
```
disq, line 22:
```suggestion
	// These accesses must be protected by the mutex.
```

plugins/source/metrics.go
```diff
@@ -123,3 +129,54 @@ func (s *Metrics) TotalResourcesAtomic() uint64 {
...
+
+	s.TableClient[table.Name][clientID].StartTime = time.Now()
+}
+
+// If the table is a top-level table, we need to mark all of its descendents as 'done' as well.
+// This is because, when a top-level table is empty (no resources), its descendants are never actually
+// synced.
+func (s *Metrics) MarkEnd(table *schema.Table, clientID string) {
+	if table.Parent == nil {
+		s.markEndRecursive(table, clientID)
```
disq: You can just `return` here and unindent the following block.

disq (changes requested):

plugins/source/metrics.go
```diff
@@ -1,22 +1,28 @@
...
 type TableClientMetrics struct {
+	// These should only be accessed with 'Atomic*' methods
 	Resources uint64
 	Errors    uint64
 	Panics    uint64
+
+	// These accesses must be proteceted by the mutex.
 	StartTime time.Time
 	EndTime   time.Time
+	Mutex     sync.Mutex
```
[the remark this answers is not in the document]
disq: I've also considered maybe `sync.RWMutex`, but I don't think `InProgressTables` is called from multiple goroutines, so would only confuse the code

erezrokah (approved):
Approving pending @disq comments

disq (approved):

plugins/source/scheduler.go
```diff
@@ -144,11 +145,19 @@ func (p *Plugin) periodicMetricLogger(ctx context.Context, wg *sync.WaitGroup) {
...
+			logLine := p.logger.Info().
 				Uint64("total_resources", p.metrics.TotalResourcesAtomic()).
 				Uint64("total_errors", p.metrics.TotalErrorsAtomic()).
 				Uint64("total_panics", p.metrics.TotalPanicsAtomic()).
-				Msg("Sync in progress")
+				Int("num_in_progress_tables", len(inProgressTables))
+
+			if len(inProgressTables) <= periodicMetricLoggerLogTablesLimit {
+				logLine.Strs("in_progress_tables", inProgressTables)
+			}
```
hermanschaaf, line 158: I think we should have an `else` case here, so that these logs still appear if there are many tables getting synced (as that can be a clue as well). Maybe something like 

```
logLine.Strs("in_progress_tables", inProgressTables[:periodicMetricLoggerLogTablesLimit])
logLine.Bool("truncated", true)
```

(And we can put `logLine.Bool("truncated", false)` in the other case). WDYT?

hermanschaaf (approved):

yevgenypats (approved):

PR #740, opened 2023-03-29
feat: Add Support For time based partitioning in `overwrite` write_mode 

bbernays:
#### Summary

---

Use the following steps to ensure your PR is ready to be reviewed

- [ ] Read the [contribution guidelines](../blob/main/CONTRIBUTING.md) 🧑‍🎓
- [ ] Run `go fmt` to format your code 🖊
- [ ] Lint your changes via `golangci-lint run` 🚨 (install golangci-lint [here](https://golangci-lint.run/usage/install/#local-installation))
- [ ] Update or add tests 🧪
- [ ] Ensure the status checks below are successful ✅

PR #741, opened 2023-03-29
fix: Update Overwrite and OverwriteDeleteStale Test

bbernays:
#### Summary

when we use `copy()` to copy the resources it does a shallow copy so the timestamp that is stored in the `memdb` gets updated as well, which is not what should be happening here. This was discovered as part of #740

### DIFF
--- a/plugins/destination/plugin_testing_overwrite_delete_stale.go
+++ b/plugins/destination/plugin_testing_overwrite_delete_stale.go
@@ -75,7 +75,7 @@ func (*PluginTestSuite) destinationPluginTestWriteOverwriteDeleteStale(ctx conte
 	updatedResource := createTestResources(table, sourceName, secondSyncTime, 1)[0]
 	for _, colIndex := range []int{2, 3, 7} {
 		old := resources[0].Data[colIndex].Get()
-		updatedResource.Data[colIndex].Set(old)
+		return updatedResource.Data[colIndex].Set(old)
 	}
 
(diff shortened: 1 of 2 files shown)


plugins/destination/plugin_testing_overwrite_delete_stale.go
```diff
@@ -72,13 +72,11 @@ func (*PluginTestSuite) destinationPluginTestWriteOverwriteDeleteStale(ctx conte
...
 	secondSyncTime := syncTime.Add(time.Second).UTC()
 
-	// copy first resource but update the sync time
-	updatedResource := schema.DestinationResource{
-		TableName: table.Name,
-		Data:      make(schema.CQTypes, len(resources[0].Data)),
+	updatedResource := createTestResources(table, sourceName, secondSyncTime, 1)[0]
+	for _, colIndex := range []int{2, 3, 7} {
+		old := resources[0].Data[colIndex].Get()
+		return updatedResource.Data[colIndex].Set(old)
```
candiduslynx, line 78: you exit early, so only `colIndex=2` is processed

candiduslynx (changes requested):

plugins/destination/plugin_testing_overwrite_delete_stale.go
bbernays: fixed

candiduslynx (approved):

PR #760, opened 2023-04-02
feat(pk): Move PK duplicate report to source side

candiduslynx:
Move PK duplicates check & reporting to Sentry from destination plugins to source ones.

This allows us to achieve the following:

1. Only the plugins we develop & maintain would be reporting the issues
2. We would be able to clearly tie the issue with the source plugin version
3. We will be able to utilize versioning & releases inSentry, too

### DIFF
--- a/plugins/source/scheduler_dfs.go
+++ b/plugins/source/scheduler_dfs.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"runtime/debug"
+	"strings"
 	"sync"
 	"sync/atomic"
 
@@ -68,6 +69,12 @@ func (p *Plugin) syncDfs(ctx context.Context, spec specs.Source, client schema.C
 	for i, table := range tables {
 		table := table
 		clients := preInitialisedClients[i]
+		pkCacheWG := new(sync.WaitGroup)
+		pkCacheWG.Add(len(clients))
+		go func() {
+			pkCacheWG.Wait()
+			p.pkCache.Done(table.Name)
+		}()
 		for _, client := range clients {
 			client := client
 			if err := p.tableSems[0].Acquire(ctx, 1); err != nil {
@@ -82,6 +89,7 @@ func (p *Plugin) syncDfs(ctx context.Context, spec specs.Source, client schema.C
 			go func() {
 				defer wg.Done()
 				defer p.tableSems[0].Release(1)
+				defer pkCacheWG.Done()
 				// not checking for error here as nothing much todo.
 				// the error is logged and this happens when context is cancelled
 				p.resolveTableDfs(ctx, table, client, nil, resolvedResources, 1)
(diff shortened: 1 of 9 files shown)


plugins/destination/managed_writer.go
```diff
@@ -71,46 +69,20 @@ func (p *Plugin) flush(ctx context.Context, metrics *Metrics, table *schema.Tabl
...
-
-		reported = true
-		pkSpec := "(" + strings.Join(table.PrimaryKeys(), ",") + ")"
-
-		// log err
-		p.logger.Error().
-			Str("table", table.Name).
-			Str("pk", pkSpec).
-			Str("value", key).
-			Msg("duplicate primary key")
```
bbernays, line 103: I understand and am for getting rid of the Sentry call below, but why get rid of the log message? Don't we consider that part of our public API?

candiduslynx: nice catch

plugins/source/scheduler_dfs.go
```diff
@@ -134,8 +137,29 @@ func (p *Plugin) resolveTableDfs(ctx context.Context, table *schema.Table, clien
...
+				Msg("duplicate primary key")
+
+			// send to Sentry only once per table,
+			// to avoid sending too many duplicate messages
+			sentry.WithScope(func(scope *sentry.Scope) {
+				scope.SetTag("table", table.Name)
+				scope.SetTag("pk", pkSpec)
+				sentry.CurrentHub().CaptureMessage("duplicate primary key " + pkSpec + " in " + table.Name)
+			})
+		}
```
bbernays, line 162: How does the plugin version get set? Is it set during initialization?

schema/resource.go
```diff
@@ -138,6 +139,24 @@ func (r *Resource) Validate() error {
...
+	for i, col := range r.Table.Columns {
+		if !col.CreationOptions.PrimaryKey {
+			continue
+		}
+		parts = append(parts, r.data[i].String())
+	}
+
+	res := "(" + strings.Join(parts, ",") + ")"
+	return &res
+}
```
bbernays, line 158: Because this is going to be part of the public interface, maybe we should just return `[]any` and then do the string concatenation where you need it?

plugins/source/scheduler_dfs.go
candiduslynx: https://github.com/cloudquery/plugin-sdk/blob/main/serve/source.go#L142

schema/resource.go
candiduslynx: Yes, thought about it, too.

candiduslynx: ret `CQTypes`

schema/types.go
```diff
@@ -272,6 +272,23 @@ func (c CQTypes) Len() int {
...
+		}
+		if v == nil {
+			sb.WriteString("nil")
+		} else {
+			sb.WriteString(v.String())
+		}
+	}
+	sb.WriteString("]")
+	return sb.String()
+}
```
bbernays, line 290: I don't think we should be adding this public function because it is so specific for the one output we are using for reporting the PK dups

bbernays (changes requested):

schema/types.go
candiduslynx: Initially it was there, did something change for v2?

candiduslynx: I didn't add it, it was already in place below, however, I moved it a bit to top, as there's a mark for functions used only in tests (and now it isn't used in tests only).

PR #762, opened 2023-04-04
feat: Add ValueString methods

hermanschaaf:
We need this for CSV reading/writing. 

(Note: this PR is to the `feat/arrow_update` branch)

### DIFF
--- a/types/json.go
+++ b/types/json.go
@@ -111,7 +111,7 @@ func (a *JSONArray) ValueString(i int) string {
 	case a.IsNull(i):
 		return "(null)"
 	default:
-		return fmt.Sprintf("\"%s\"", arr.Value(i))
+		return fmt.Sprintf(`"%s"`, arr.Value(i))
 	}
 }
 
--- a/types/json.go
+++ b/types/json.go
@@ -105,6 +105,16 @@ func (a JSONArray) String() string {
 	return o.String()
 }
 
+func (a *JSONArray) ValueString(i int) string {
+	arr := a.Storage().(*array.Binary)
+	switch {
+	case a.IsNull(i):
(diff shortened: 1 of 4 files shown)


types/json.go
```diff
@@ -105,6 +105,16 @@ func (a JSONArray) String() string {
 	return o.String()
 }
 
+func (a *JSONArray) ValueString(i int) string {
+	arr := a.Storage().(*array.Binary)
+	switch {
+	case a.IsNull(i):
+		return "(null)"
+	default:
+		return fmt.Sprintf("\"%s\"", arr.Value(i))
```
disq, line 114:
```suggestion
		return fmt.Sprintf("%q", arr.Value(i))
```
So that the value itself can be quoted as well?

types/mac.go
```diff
@@ -119,6 +119,17 @@ func (a MacArray) String() string {
...
 }
 
+func (a *MacArray) ValueString(i int) string {
+	arr := a.Storage().(*array.Binary)
+	switch {
+	case a.IsNull(i):
+		return "(null)"
+	default:
+		mac := net.HardwareAddr(arr.Value(i))
+		return fmt.Sprintf("\"%s\"", mac)
```
disq, line 129:
```suggestion
		return fmt.Sprintf("%q", mac)
```
A macaddr shouldn't need it, but just for consistency

hermanschaaf:
I've tested it now and it works end-to-end with the CSV writer/reader. The only case that isn't working yet is JSON, but I think there's a specific issue there related to the Arrow CSV reader and UnmarshalOne behavior.

yevgenypats (approved):

PR #786, opened 2023-04-17
fix: Make cq_id unique for backward compat

yevgenypats:
#### Summary

---

- [ ] Update or add tests 🧪


internal/servers/destination/v0/destinations.go
```diff
@@ -162,6 +162,11 @@ func setCQIDAsPrimaryKeysForTables(tables schema.Tables) {
 // Overwrites or adds the CQ columns that are managed by the destination plugins (_cq_sync_time, _cq_source_name).
 func SetDestinationManagedCqColumns(tables []*schema.Table) {
 	for _, table := range tables {
+		for i := range table.Columns {
+			if table.Columns[i].Name == schema.CqIDColumn.Name {
+				table.Columns[i].CreationOptions.Unique = true
```
erezrokah, line 167: If we really want to go back we could set it to non null (I think that was the old logic)

erezrokah (approved):

PR #795, opened 2023-04-18
fix: Arrow Retain and Release fixes

hermanschaaf:
This fixes a memory leak detected by the `file` and `s3` destination tests, and refactors the retain/release logic a little bit.

Quoting the Go Arrow docs:

> If you send an object over a channel, you must call Retain before sending it as the receiver is assumed to own the object and will later call Release when it no longer needs the object.

I think we should call `Retain` as close as possible to where a record is sent over a channel.


plugins/destination/managed_writer.go
```diff
@@ -81,18 +84,19 @@ func (*Plugin) removeDuplicatesByPK(table *arrow.Schema, resources []arrow.Recor
...
+		}
 		key := pk.String(r)
 		_, ok := pks[key]
-		switch {
-		case !ok:
+		if !ok {
 			pks[key] = struct{}{}
 			res = append(res, r)
 			continue
-		case reported:
```
hermanschaaf, line 93: the `reported` variable was not being used

plugins/destination/managed_writer.go
```diff
@@ -81,18 +84,19 @@ func (*Plugin) removeDuplicatesByPK(table *arrow.Schema, resources []arrow.Recor
 
 	pks := make(map[string]struct{}, len(resources))
 	res := make([]arrow.Record, 0, len(resources))
-	var reported bool
 	for _, r := range resources {
+		if r.NumRows() > 1 {
+			panic(fmt.Sprintf("record with more than 1 row: %d", r.NumRows()))
```
hermanschaaf, line 89: I also added this panic because this function's logic will break down if a record ever contains more than one row.

I wonder if we may want to add a special type that guarantees we're dealing with a `arrow.Record` with only a single row.

plugins/destination/plugin_testing_write_append.go
```diff
@@ -40,7 +40,6 @@ func (s *PluginTestSuite) destinationPluginTestWriteAppend(ctx context.Context,
 		MaxRows:    1,
 	}
 	record1 := testdata.GenTestData(mem, table.ToArrowSchema(), opts)[0]
-	record1.Retain()
```
yevgenypats, line 43: Why this is removed? I specifically called `Retain` here because unlike in the server we do use this after call to `writeOne` or `writeAll` for compare/testing purposes. This is also why I called `defer record1.Release` afterwards.

yevgenypats (changes requested):
Had another counter comment :)

plugins/destination/plugin_testing_write_append.go
hermanschaaf: This is basically just a refactor: no change in functionality. Rather than have many retains, I think it's better to have it only in `writeAll` (and by extension `writeOne`), where it's clear that the resource is being sent over a channel. Here, just by looking at the call to `writeOne`, you can't really know whether it's going to be sent over a channel or not.

The record is still available for comparison until the end of this function because the release is deferred.

yevgenypats: I think it is crucial for it to be here. I can explain :) maybe let's sync tmrw morning.

hermanschaaf:
@yevgenypats Thanks for catching up with me earlier; I've now updated it as we discussed.

The TL;DR is that we'll keep the extra `Retain` calls in the tests to make it clear that we're keeping the records around for later comparison; normally it is still the plugin's responsibility to call `Release`.

yevgenypats (approved):

PR #804, opened 2023-04-19
fix(testing): Migration tests should generate and sort records in-order before comparing

disq:


plugins/destination/plugin_testing_migrate.go
```diff
@@ -62,6 +65,9 @@ func testMigration(ctx context.Context, mem memory.Allocator, _ *testing.T, p *P
 	if err != nil {
 		return fmt.Errorf("failed to read all: %w", err)
 	}
+
+	sortRecordsBySyncTimeArrow(source, resourcesRead)
```
disq, line 69: This makes sure `Read()` returned our test rows in "insertion order"

plugins/destination/plugin_testing.go
```diff
@@ -216,11 +216,28 @@ func PluginTestSuiteRunner(t *testing.T, newPlugin NewPluginFunc, destSpec specs
...
+	if len(syncTimeIndex) != 1 {
+		panic("no CqIDColumn in schema")
+	}
+	sortRecordsBySyncTimeIndex(syncTimeIndex[0], cqIDIndex[0], records)
+}
+
+func sortRecordsBySyncTimeIndex(syncTimeIndex, cqIDIndex int, records []arrow.Record) {
 	sort.Slice(records, func(i, j int) bool {
 		// sort by sync time, then UUID
 		first := records[i].Column(syncTimeIndex).(*array.Timestamp).Value(0).ToTime(arrow.Millisecond)
```
disq, line 237: For completeness this should be
```suggestion
		first := records[i].Column(syncTimeIndex).(*array.Timestamp).Value(0).ToTime(records[i].Column(syncTimeIndex).(*array.Timestamp).DataType().(*arrow.TimestampType).Unit)
```
And so on, but it has no effect for the sake of comparing timestamps.

plugins/destination/plugin_testing_migrate.go
disq: It's not necessary for it to be here if the destination's `Read()` always returns the results sorted, which is another option

disq:
Reimplemented by https://github.com/cloudquery/plugin-sdk/pull/814

PR #810, opened 2023-04-19
fix: Use Go memory allocator for arrow

yevgenypats:
Following this short discussion - https://github.com/apache/arrow/issues/35232

It seems we don't really need to use Retain/Release outside the arrow library. We can always bring this back in the future if we would like to experiment if this brings better memory performance then the default go allocator.

### DIFF
--- a/internal/memdb/memdb.go
+++ b/internal/memdb/memdb.go
@@ -86,10 +86,8 @@ func (c *client) overwrite(table *arrow.Schema, data arrow.Record) {
 			}
 		}
 		if found {
-			tmp := c.memoryDB[tableName][i]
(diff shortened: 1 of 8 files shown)


candiduslynx:
So, we should update the plugin code as well? (Stop using release with this update?)

yevgenypats:
> So, we should update the plugin code as well? (Stop using release with this update?)

Yes.

erezrokah (approved):

hermanschaaf (commented):
Nice 💪

erezrokah:
Follow up question, do we need the memory checks we have in the some of the tests https://github.com/cloudquery/plugin-sdk/blob/a61f34ec34e8ec7c3b7663f09b76843d2529e099/types/inet_test.go#L21 if we use the Go allocator?

PR #811, opened 2023-04-19
fix(testdata): Add old style gen testdata

candiduslynx:
Required by:
* https://github.com/cloudquery/cloudquery/pull/10166
* https://github.com/cloudquery/cloudquery/pull/10060


testdata/testdata.go
```diff
@@ -245,3 +245,122 @@ func GenTestData(mem memory.Allocator, sc *arrow.Schema, opts GenTestDataOptions
 	}
 	return records
 }
+
```
yevgenypats, line 248: Can you also put a comment that it is deprecated and will be removed in the future?

yevgenypats (approved):

testdata/testdata.go
candiduslynx: Lint will fail the deprecation notice then.

candiduslynx: done in https://github.com/cloudquery/plugin-sdk/pull/812

PR #812, opened 2023-04-19
chore(testdata): Deprecation notice

candiduslynx:
Follow-up for #811

### DIFF
--- a/testdata/testdata.go
+++ b/testdata/testdata.go
@@ -246,6 +246,8 @@ func GenTestData(mem memory.Allocator, sc *arrow.Schema, opts GenTestDataOptions
 	return records
 }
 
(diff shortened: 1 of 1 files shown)


testdata/testdata.go
```diff
@@ -246,6 +246,8 @@ func GenTestData(mem memory.Allocator, sc *arrow.Schema, opts GenTestDataOptions
 	return records
 }
 
+// GenTestDataV1 does approximately the same job as GenTestData, however, it's intended for simpler use-cases.
+// Deprecated. Will be removed in future release.
```
disq, line 250:
```suggestion
// Deprecated: Will be removed in future release.
```

According to https://go.dev/blog/godoc

erezrokah (approved):

erezrokah:
Forced merge so we can release this with https://github.com/cloudquery/plugin-sdk/pull/813

Issue #805, opened 2023-04-19
feat: Add all support arrow types to our tests (e.g. int8, int16, float64)


erezrokah:
### Are you positive this feature should be implemented in the SDK?

- [X] I'm positive this is an SDK feature

### Which problem is this feature request solving?

Our test table used in destinations only supports a subset of the arrow types https://github.com/cloudquery/plugin-sdk/blob/506dab186c7eb4766daa060fed26633661b4c320/testdata/testdata.go#L15 so we can't easily test `int8`, `int16`, etc.

### Describe the solution you'd like

This is a feature request to support all arrow types so we can test them properly. We should probable add those as `schema.Type` (or replace our own typing first).

### Pull request (optional)

- [ ] I can submit a pull request

erezrokah:
I believe this was resolved by https://github.com/cloudquery/plugin-sdk/pull/863. Please re-open if not the case cc @hermanschaaf

PR #818, opened 2023-04-21
feat: Test more arrow types

hermanschaaf:
This adds support for testing (almost) the complete range of Arrow types in destinations. 

By default, only the types previously supported by CQTypes are enabled (with the important exception of more uint and int types than before, plus float32). The `TestSourceOptions` struct allows destinations to declare that they want to test with more column types, including maps, structs, times, etc.

Closes https://github.com/cloudquery/plugin-sdk/issues/805


disq (changes requested):
Some nits and a slice copy issue

testdata/schema.go
```diff
@@ -0,0 +1,133 @@
...
+	for i := 0; i < fixedWidthTypesType.NumField(); i++ {
+		fieldName := fixedWidthTypesType.Field(i).Name
+		dataType := fixedWidthTypesValue.FieldByName(fieldName).Interface().(arrow.DataType)
+		fields[i] = arrow.Field{Name: strings.ToLower(fieldName), Type: dataType, Nullable: true}
+	}
+	return fields
+}
+
+func SortAndRemoveDuplicates(fields []arrow.Field) []arrow.Field {
+	newFields := fields
```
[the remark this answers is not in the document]
hermanschaaf: Good catch, thanks

testdata/schema.go
```diff
@@ -0,0 +1,133 @@
...
+	fields := make([]arrow.Field, fixedWidthTypesType.NumField())
+	for i := 0; i < fixedWidthTypesType.NumField(); i++ {
+		fieldName := fixedWidthTypesType.Field(i).Name
+		dataType := fixedWidthTypesValue.FieldByName(fieldName).Interface().(arrow.DataType)
+		fields[i] = arrow.Field{Name: strings.ToLower(fieldName), Type: dataType, Nullable: true}
+	}
+	return fields
+}
+
+func SortAndRemoveDuplicates(fields []arrow.Field) []arrow.Field {
```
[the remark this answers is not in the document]
hermanschaaf: Doesn't need to be; it's somewhat useful but will unexport it for now.

testdata/schema.go
```diff
@@ -0,0 +1,133 @@
...
+	for i := 0; i < len(baseFields); i++ {
+		fields[i] = arrow.Field{Name: baseFields[i].Name + "_map", Type: arrow.MapOf(baseFields[i].Type, baseFields[i].Type), Nullable: true}
+	}
+	return fields
+}
+
+// TestSourceFields returns fields for all Arrow types and composites thereof
+func TestSourceFields() []arrow.Field {
+	// cq fields
+	cqFields := make([]arrow.Field, 0)
```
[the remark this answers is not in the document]
hermanschaaf: Thanks

testdata/testdata.go
```diff
@@ -297,13 +292,20 @@ func getExampleJSON(colName string, dataType arrow.DataType, opts GenTestDataOpt
 			case arrow.FixedWidthTypes.Timestamp_ns:
 				return fmt.Sprintf("%d", t.UnixNano())
 			case arrow.FixedWidthTypes.Time32s:
-				return fmt.Sprintf("%d", t.Unix())
+				h, m, s := t.Clock()
+				return fmt.Sprintf("%d", h*3600+m*60+s)
```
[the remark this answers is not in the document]
hermanschaaf: You know this is still in draft right :P

plugins/destination/plugin_testing_overwrite_delete_stale.go
```diff
@@ -7,21 +7,26 @@ import (
...
 	spec.WriteMode = specs.WriteModeOverwriteDeleteStale
 	if err := p.Init(ctx, logger, spec); err != nil {
 		return fmt.Errorf("failed to init plugin: %w", err)
 	}
 	tableName := fmt.Sprintf("cq_%s_%d", spec.Name, time.Now().Unix())
-	table := testdata.TestTable(tableName).ToArrowSchema()
-	incTable := testdata.TestTableIncremental(tableName + "_incremental").ToArrowSchema()
+	table := testdata.TestSourceSchema(tableName, testSourceOptions)
+	incrementalMetadata := arrow.MetadataFrom(map[string]string{
+		schema.MetadataTableName:   tableName + "_incremental",
```
candiduslynx, line 26: IMO we mustn't add some custom suffix.
The table name should be declarative.
If we want the suffix we need to add it to the table name instead of adding it to the metadata options).

testdata/schema.go
```diff
@@ -0,0 +1,232 @@
...
+				newFields = append(newFields[:i], newFields[i+1:]...)
+				i--
+			}
+		}
+	}
+	return newFields
+}
+
+// ListOfFields returns a list of fields that are lists of the given fields.
+func ListOfFields(baseFields []arrow.Field) []arrow.Field {
```
candiduslynx, line 96: Large lists, fixed size lists?

testdata/schema.go
```diff
@@ -0,0 +1,232 @@
...
+	cqFields = append(cqFields, arrow.Field{Name: schema.CqIDColumn.Name, Type: types.NewUUIDType(), Nullable: false, Metadata: cqIDMetadata})
+	cqFields = append(cqFields, arrow.Field{Name: schema.CqParentIDColumn.Name, Type: types.NewUUIDType(), Nullable: false})
+
+	var basicFields []arrow.Field
+	basicFields = append(basicFields, PrimitiveFields()...)
+	basicFields = append(basicFields, BinaryFields()...)
+	basicFields = append(basicFields, FixedWidthFields()...)
+
+	// add extensions
+	basicFields = append(basicFields, arrow.Field{Name: "uuid", Type: types.NewUUIDType(), Nullable: true})
```
candiduslynx, line 141: Need to add both nullable and not nullable fields of the same type to the schema

testdata/schema.go
```diff
@@ -0,0 +1,232 @@
...
+	// cq fields
+	var cqFields []arrow.Field
+	cqIDMetadata := arrow.NewMetadata([]string{schema.MetadataUnique}, []string{"true"})
+	cqFields = append(cqFields, arrow.Field{Name: schema.CqIDColumn.Name, Type: types.NewUUIDType(), Nullable: false, Metadata: cqIDMetadata})
+	cqFields = append(cqFields, arrow.Field{Name: schema.CqParentIDColumn.Name, Type: types.NewUUIDType(), Nullable: false})
+
+	var basicFields []arrow.Field
+	basicFields = append(basicFields, PrimitiveFields()...)
+	basicFields = append(basicFields, BinaryFields()...)
+	basicFields = append(basicFields, FixedWidthFields()...)
```
candiduslynx, line 138: Need to add both nullable and not nullable fields of the same type to the schema

testdata/schema.go
```diff
@@ -0,0 +1,232 @@
...
+
+	// add extensions
+	basicFields = append(basicFields, arrow.Field{Name: "uuid", Type: types.NewUUIDType(), Nullable: true})
+	basicFields = append(basicFields, arrow.Field{Name: "inet", Type: types.NewInetType(), Nullable: true})
+	basicFields = append(basicFields, arrow.Field{Name: "mac", Type: types.NewMacType(), Nullable: true})
+
+	// sort and remove duplicates (e.g. date32 and date64 appear twice)
+	basicFields = sortAndRemoveDuplicates(basicFields)
+
+	// we don't support float16 right now
```
candiduslynx, line 148: We can, though?
It should map.to float32 in the most cases

testdata/schema.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+
+	// add JSON later, we don't want to include it as a list or map right now (it causes complications with JSON unmarshalling)
+	basicFields = append(basicFields, arrow.Field{Name: "json", Type: types.NewJSONType(), Nullable: true})
+
+	if opts.IncludeStructs {
+		// struct with all the types
+		compositeFields = append(compositeFields, arrow.Field{Name: "struct", Type: arrow.StructOf(basicFields...), Nullable: true})
+
+		// struct with nested struct
```
candiduslynx, line 201: + with field as list of steucts, as map with struct as field

testdata/testdata.go
```diff
@@ -262,6 +67,185 @@ func GenTestData(sc *arrow.Schema, opts GenTestDataOptions) []arrow.Record {
...
+	}
+
+	// handle binary types
+	binaryTypes := []arrow.DataType{
+		arrow.BinaryTypes.Binary,
+		arrow.BinaryTypes.LargeBinary,
+	}
+	for _, binaryType := range binaryTypes {
+		if arrow.TypeEqual(dataType, binaryType) {
+			return `"AQIDBA=="` // base64 encoded 0x01, 0x02, 0x03, 0x04
```
candiduslynx, line 145: Explicitly run base64 encode instead of magic strings

candiduslynx (changes requested):

plugins/destination/plugin_testing_overwrite_delete_stale.go
hermanschaaf: I'm not sure what you're saying here; this isn't new, it's just a refactor of what we had before to use the new method. If you'd like to change the `_incremental` suffix, let's do that separately.

testdata/schema.go
hermanschaaf: Good point, but I think we can add these when the need arises? Right now adding a new type requires handling in every destination, so we shouldn't add a new type to the tests unless there is a need from it from a source plugin. I'm not aware of sources that would need large lists or fixed size lists.

testdata/schema.go
hermanschaaf: Maybe; but what is your reasoning?

testdata/schema.go
hermanschaaf: This comment is the same as the one above?

testdata/schema.go
hermanschaaf: We can, but that doesn't mean we should. As I said above, supporting new types comes with a cost of code that we need to maintain in all destinations right now. Until that changes, we should think carefully about what we add. In this case, I think we shouldn't support it:

 1. Go doesn't support float16, so no struct conversion will require it
 2. Postgres doesn't have float16, so the pg source plugin won't require it
 3. Even the official Arrow CSV and Parquet implementations don't support float16, so we would have to go contribute that upstream to support this

testdata/schema.go
hermanschaaf: We can always add more cases later, I tried to keep it relatively simple for now

testdata/testdata.go
hermanschaaf: There is a comment that says what it is, so flipping the order of comment / string doesn't really matter imo. For me this is useful to show that the base64-encoded string uses padding

plugins/destination/plugin_testing_overwrite_delete_stale.go
candiduslynx: I mean, we need to have `tableName := "a_b_c_incremental"` in the future, but yeah, may be left out of scope

testdata/schema.go
candiduslynx: CH: Array may contain nullable elements, but the array itself shouldn't be marked nullable.
We need to pass these types to verify the correct destination logic:
* nullable array of nullable
* nullable array of non-nullable
* non-nullable array of nullable
* non-nullable array of non-nullable

testdata/schema.go
candiduslynx: yes

yevgenypats (changes requested):
Looks great! I worked on https://github.com/cloudquery/plugin-sdk/pull/832, I think if we can review/merge 832 first and then rebase on top it should also simplify part of this PR as well due to the new abstraction

testdata/schema.go
hermanschaaf: I see, thanks; maybe we can add 4 columns similar to what you suggested just to test nullable behavior specifically, rather than adding a nullable column for every type (doubling the number of columns)? There are already around 200 columns if all the options are enabled, and doubling that will make it harder to figure out what really went wrong in a test.

candiduslynx: we can use `testing.T.Run` + utilize the `fieldPrettify` logic to construct the name to be able to deconstruct easily: https://github.com/cloudquery/plugin-sdk/blob/main/schema/arrow.go#L60

hermanschaaf:
Replaced by https://github.com/cloudquery/plugin-sdk/pull/863

candiduslynx:
@hermanschaaf is it OK to remove the branch as well?

PR #822, opened 2023-04-24
fix: Destination migration testing using incorrect mode

yevgenypats:
#### Summary

---

- [ ] Update or add tests 🧪

### DIFF
--- a/plugins/destination/plugin_testing.go
+++ b/plugins/destination/plugin_testing.go
@@ -146,6 +146,7 @@ func PluginTestSuiteRunner(t *testing.T, newPlugin NewPluginFunc, destSpec specs
 			t.Skip("skipping " + t.Name())
 		}
 		destSpec.WriteMode = specs.WriteModeOverwrite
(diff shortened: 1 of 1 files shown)


disq:
```go
const (
	MigrateModeSafe MigrateMode = iota
	MigrateModeForced
)
```

How is it incorrect? Zero value defaults to safe. Maybe not 'explicit', but shouldn't be incorrect?

hermanschaaf (approved):

hermanschaaf:
Yeah, I think it's just not explicit, but explicit is better than implicit I think, so we may want to just update the title of this PR if it's not fixing anything in reality.

yevgenypats:
> ```go
> const (
> 	MigrateModeSafe MigrateMode = iota
> 	MigrateModeForced
> )
> ```
> 
> How is it incorrect? Zero value defaults to safe. Maybe not 'explicit', but shouldn't be incorrect?

I think it might be due to a race where the variable was changed in another test

PR #823, opened 2023-04-25
feat(arrow): Streamline Apache Arrow extension types

candiduslynx:
### Common

Use pointer receivers:
1. For Array types: to have all functions have the same receiver type (other functions do have pointer receivers, so updating `String()` method receiver is natural, see `net.IPNet` as a reference.
2. For `Type` implementations: we do return a pointer from `NewXYZType` function, so it's only natural to define the receiver as pointer as well.

#### `InetBuilder`

1. Now accepts `*net.IPNet` (pointer) in `Append`, `UnsafeAppend` & `AppendValues`
3. `AppendValues` now accounts for `valid` param

#### `InetArray`

1. `String` method uses `%q` format as mainstream Apache Arrow types
2. Added `Value(i int) *net.IPNet` function
4. `GetOneForMarshal` utilizes `Value` implementation

#### `JSONBuilder`

1. Marshaling with `json.DisableHTMLEscape()` option per #622
2. `AppendValues` now accounts for `valid` param
3. `UnmarshalJSON` implementation is streamlined with helper `Unmarshal(dec *json.Decoder) error` function (as mainstream Apache Arrow types)

#### `JSONArray`

1. `String` method uses `%q` format as mainstream Apache Arrow types
2. Added `Value(i int) any` function that will unmarshal the stored data & return the concrete type
3. Use `json.UnmarshalNoEscape` per #622
5. `GetOneForMarshal` now doesn't perform unmarshaling and, instead, returns the stored `json.RawMessage`

#### `MacBuilder`

1. `AppendValues` now accounts for `valid` param

#### `MacArray`

1. `String` method uses `%q` format as mainstream Apache Arrow types
2. Added `Value(i int) net.HardwareAddr` function

#### `UUIDBuilder`

1. `AppendValues` now accounts for `valid` param
2. Use `uuid.Must` func instead of panic in code

### `UUIDArray`

1. `String` method uses `%q` format as mainstream Apache Arrow types
2. Added `Value(i int) uuid.UUID` func
3. Use `uuid.Must` func instead of panic in code

### DIFF
--- a/types/inet.go
+++ b/types/inet.go
@@ -128,7 +128,7 @@ func (a InetArray) String() string {
 		case a.IsNull(i):
 			o.WriteString(array.NullValueStr)
 		default:
-			fmt.Fprintf(o, "\"%s\"", arr.Value(i))
+			fmt.Fprintf(o, "%q", a.ValueStr(i))
 		}
 	}
 	o.WriteString("]")
--- a/types/inet.go
+++ b/types/inet.go
@@ -135,24 +135,30 @@ func (a InetArray) String() string {
 	return o.String()
 }
 
-func (a *InetArray) ValueStr(i int) string {
+func (a *InetArray) Value(i int) *net.IPNet {
+	if a.IsNull(i) { // IsValid ~ !IsNull
+		return nil
+	}
 	arr := a.Storage().(*array.String)
+	_, ipnet, err := net.ParseCIDR(arr.Value(i))
+	if err != nil {
+		panic(fmt.Errorf("invalid ip+net: %w", err))
+	}
+	return ipnet
+}
+
+func (a *InetArray) ValueStr(i int) string {
 	switch {
 	case a.IsNull(i):
 		return array.NullValueStr
 	default:
-		return arr.Value(i)
+		return a.Value(i).String()
 	}
 }
 
 func (a *InetArray) GetOneForMarshal(i int) any {
-	arr := a.Storage().(*array.String)
-	if a.IsValid(i) {
-		_, ipnet, err := net.ParseCIDR(arr.Value(i))
-		if err != nil {
-			panic(fmt.Errorf("invalid ip+net: %w", err))
-		}
-		return ipnet.String()
+	if val := a.Value(i); val != nil {
+		return val.String()
 	}
 	return nil
 }
--- a/types/inet.go
+++ b/types/inet.go
@@ -126,7 +126,7 @@ func (a InetArray) String() string {
 		}
 		switch {
 		case a.IsNull(i):
-			o.WriteString("(null)")
+			o.WriteString(array.NullValueStr)
 		default:
 			fmt.Fprintf(o, "\"%s\"", arr.Value(i))
 		}
@@ -139,7 +139,7 @@ func (a *InetArray) ValueStr(i int) string {
 	arr := a.Storage().(*array.String)
 	switch {
 	case a.IsNull(i):
-		return "(null)"
+		return array.NullValueStr
 	default:
 		return arr.Value(i)
 	}
--- a/types/inet.go
+++ b/types/inet.go
@@ -208,6 +208,6 @@ func (u *InetType) ExtensionEquals(other arrow.ExtensionType) bool {
 	return u.ExtensionName() == other.ExtensionName()
 }
 
-func (*InetType) NewBuilder(builder *array.ExtensionBuilder) array.Builder {
-	return NewInetBuilder(builder)
+func (*InetType) NewBuilder(bldr *array.ExtensionBuilder) array.Builder {
+	return NewInetBuilder(bldr)
 }
--- a/types/inet.go
+++ b/types/inet.go
@@ -136,7 +136,7 @@ func (a InetArray) String() string {
 }
 
 func (a *InetArray) Value(i int) *net.IPNet {
-	if a.IsNull(i) { // IsValid ~ !IsNull
+	if a.IsNull(i) {
 		return nil
 	}
 	arr := a.Storage().(*array.String)
--- a/types/inet.go
+++ b/types/inet.go
@@ -182,6 +182,7 @@ func NewInetType() *InetType {
 			Storage: &arrow.StringType{}}}
 }
 
+// ArrayType returns TypeOf(InetArray{}) for constructing Inet arrays
 func (*InetType) ArrayType() reflect.Type {
 	return reflect.TypeOf(InetArray{})
 }
--- a/types/inet.go
+++ b/types/inet.go
@@ -16,9 +16,9 @@ type InetBuilder struct {
 	*array.ExtensionBuilder
 }
 
-func NewInetBuilder(bldr *array.ExtensionBuilder) *InetBuilder {
+func NewInetBuilder(builder *array.ExtensionBuilder) *InetBuilder {
 	b := &InetBuilder{
-		ExtensionBuilder: bldr,
+		ExtensionBuilder: builder,
 	}
 	return b
 }
@@ -121,7 +121,7 @@ type InetArray struct {
 	array.ExtensionArrayBase
 }
 
-func (a InetArray) String() string {
+func (a *InetArray) String() string {
 	arr := a.Storage().(*array.String)
 	o := new(strings.Builder)
 	o.WriteString("[")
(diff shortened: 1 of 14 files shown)


types/inet.go
```diff
@@ -135,24 +135,30 @@ func (a InetArray) String() string {
 	return o.String()
 }
 
-func (a *InetArray) ValueStr(i int) string {
+func (a *InetArray) Value(i int) *net.IPNet {
+	if a.IsNull(i) { // IsValid ~ !IsNull
```
hermanschaaf, line 139: I'm not 100% sure what this comment is trying to say; is it saying `IsValid is approximately the same as !IsNull` ? Why not just use `IsNull` and leave it at that?

candiduslynx: I saw several times that we used `if a.IsNull(i) || !a.IsValid(i)` and wanted to express that those are, in fact, just opposites (judging by the code, the following is true: `a.IsNull(i) == !a.IsValid(i)` for every `i`

hermanschaaf: Okay, yeah it's fine to remove it I think. We should just go and clean up the other spots in the cloudquery codebase so the `!IsValid()` version doesn't keep getting copy-pasted around

types/mac.go
```diff
@@ -122,7 +122,7 @@ func (a MacArray) String() string {
 		}
 		switch {
 		case a.IsNull(i):
-			o.WriteString("(null)")
+			o.WriteString(array.NullValueStr)
 		default:
 			mac := net.HardwareAddr(arr.Value(i))
 			fmt.Fprintf(o, "\"%s\"", mac)
```
hermanschaaf, line 128: Could this now be:

```
fmt.Fprintf(o, "\"%s\"", a.Value(i))
```

?

types/uuid.go
```diff
@@ -137,7 +137,7 @@ func (a UUIDArray) String() string {
 		}
 		switch {
 		case a.IsNull(i):
-			o.WriteString("(null)")
+			o.WriteString(array.NullValueStr)
 		default:
 			uuidStr, err := uuid.FromBytes(arr.Value(i))
 			if err != nil {
```
hermanschaaf, line 143: Also seems like `a.Value(i)` could be used here now instead (just a refactor basically)

candiduslynx: I'd like to scope it to values mostly (maybe NullValueStr belongs to a separate PR?)

types/json.go
```diff
@@ -113,13 +113,19 @@ func (a JSONArray) String() string {
 	return o.String()
 }
 
+func (a *JSONArray) Value(i int) []byte {
```
hermanschaaf, line 116: For better or worse, `AppendValues` takes `[]any`, so probably `Value` should return `any` (the unmarshalled bytes). Though I'm not sure `AppendValues` is doing the right thing here to be honest. What do you think?

types/uuid.go
candiduslynx: Followed up in 8b840b7 for other types (json had to use valuestr, though)

types/json.go
candiduslynx: It also doesn't account for invalid values being appended...
I'll take a look

candiduslynx: But I think we can return `json.RawMessage` (~ `[]byte`) to be safe, as we can always say that unmarshaled is returned by `GetOneForMarshal`

candiduslynx: Moved ret type to `json.RawMessage` in 9a305de

types/json.go
```diff
@@ -104,22 +107,28 @@ func (a JSONArray) String() string {
 		}
 		switch {
 		case a.IsNull(i):
-			o.WriteString("(null)")
+			o.WriteString(array.NullValueStr)
 		default:
-			fmt.Fprintf(o, "\"%s\"", arr.Value(i))
+			fmt.Fprintf(o, "\"%s\"", a.ValueStr(i))
```
candiduslynx, line 112: I do think that this is wrong (as JSON can have quotes in it, so we should use `%q` here instead)

candiduslynx: Addressed in 8ea6a1c

hermanschaaf: Yeah, doesn't look right to me either

types/inet.go
```diff
@@ -126,33 +126,39 @@ func (a InetArray) String() string {
...
-			fmt.Fprintf(o, "\"%s\"", arr.Value(i))
+			fmt.Fprintf(o, "%q", a.ValueStr(i))
 		}
 	}
 	o.WriteString("]")
 	return o.String()
 }
 
-func (a *InetArray) ValueStr(i int) string {
+func (a *InetArray) Value(i int) *net.IPNet {
```
candiduslynx, line 138: @hermanschaaf I wonder if we need to return `net.IPNet` instead (receivers work on top of the pointer, but we might as well consider it)

types/json.go
```diff
@@ -104,22 +107,28 @@ func (a JSONArray) String() string {
...
 		default:
-			fmt.Fprintf(o, "\"%s\"", arr.Value(i))
+			fmt.Fprintf(o, "%q", a.ValueStr(i))
 		}
 	}
 	o.WriteString("]")
 	return o.String()
 }
 
+func (a *JSONArray) Value(i int) json.RawMessage {
```
candiduslynx, line 119: TODO decided:
* `Value` -> `json.RawMessage`, `GetOneForMarshal` -> parsed value
* `Value` -> parsed value, `GetOneForMarshal` -> `json.RawMessage`
cc: @yevgenypats @hermanschaaf

types/inet.go
hermanschaaf: Yeah, I think because `Append(v net.IPNet)` on the builder takes a non-pointer, we can be consistent and also return the same here. The fields of `net.IPNet` are slices so there should be very little difference performance-wise.

types/json.go
hermanschaaf: The JSON type was the most difficult one to deal with in the CSV and Parquet writers/readers, and I think it's because it's unclear what it represents, and so some of the methods are inconsistent. 

For example, should calling MarshalJSON on the JSON type give you unescaped JSON, or an escaped JSON string? When you append values, should you append strings that are already marshalled JSON, or the raw objects?

In other words, does it represent any Go object (which happens to be stored on disk as JSON), or does it represent a valid JSON string?

I think it doesn't matter what we decide, we just need to pick one and stick with it. The other implementations can then know what to expect and handle it accordingly.

If I have to pick one, I'd vote for treating everything like a string, so callers need to call json.Marshal before sending it to the builder, and Value() will just return a string. The main purpose of this type will then be to document that the stored string contains valid JSON, which can be useful in some databases.

candiduslynx: Discussed internally, proposed plan:

* `Builder` implementation:
  * `Append(v any)` takes any value and marshals it (including `struct{...}`, `slice`, `nil` & `json.RawMessage`)
  * `AppendValues(v []any, valid []bool)` is same as `Append` with respect to `valid` slice
  * `AppendValueFromString(s string) error` will unmarshal to verify that the value is indeed the correct JSON, then save the `json.RawMessage`
  * `UnmarshalOne(dec *json.Decoder) error` will unmarshal & store using `Append` / `AppendNull`
  * **new**: `Unmarshal(dec *json.Decoder)` - unmarshal encoded slice
  * `UnmarshalJSON([]byte) error` - same as one, but accounts for value being marshaled slice
* `Array` implementation
  * `String() string` - joined of quoted `json.RawMessage` values cast to strings
  * `Value(i int) any` - unmarshaled value (`map`, `slice`, etc.)
  * `ValueStr(i int) string` - `string(json.RawMesaage)`, no quoting performed
  * `MarshalJSON() ([]byte, error)` - marshal as slice of `json.RawMessage`
  * `GetOneForMarshal(i int) any` - return stored `json.RawMessage`
```

The inspiration for the `Unmarshal` functions is taken from Apache Arrow builders

Comment (candiduslynx) [reply types/json.go]: Done in 7f6afc8

File: types/inet.go (hunk 6 above)
Comment (candiduslynx) [reply types/inet.go:143]: Decided to accept pointer in Append instead (as the receiver for IPNet is pointer).
Done in 63de964

**Review by hermanschaaf — APPROVED** [review]

**Review by yevgenypats — CHANGES_REQUESTED** [review]
Im not sure I understand the reasoning or the value of using pointer receivers in some of the extensions instead of regular params? Were there any performance issue? we turn them into `string` anyway and by using `pointers` we just introduce more risk of the data being changed inside while we don't really want that to happen so I can't see the value unless Im missing something.

**Comment by candiduslynx** [comment]
@yevgenypats 
> we just introduce more risk of the data being changed inside

We actually don't, as we still store the string (for `*net.IPNet` we check if the passed value is `nil` (we do `AppendNull` and return) or not (store the `String()` result, just like before).
However, having the `*net.IPNet` being returned from `Value(int)` method allows us to have `Value(i).String()` chain, whereas returning value would make us to do
```go
ipnet := a.Value(i)
return ipnet.String()
```

Additionally, upstream Apache Arrow builders/arrays always have the same value type accepted inside the builder & returned  from the respective array. So that's the reason why the builder for `Inet` will accept the pointer.

types/json_test.go
```diff
@@ -37,7 +37,7 @@ func TestJSONBuilder(t *testing.T) {
 	require.Zero(t, b.Len(), "unexpected ArrayBuilder.Len(), NewJSONBuilder did not reset state")
 	require.Zero(t, b.Cap(), "unexpected ArrayBuilder.Cap(), NewJSONBuilder did not reset state")
 	require.Zero(t, b.NullN(), "unexpected ArrayBuilder.NullN(), NewJSONBuilder did not reset state")
-	require.Equal(t, `["{"a":1,"b":2}" (null) "{"c":3,"d":4}" (null) "{"e":5,"f":6}" "{"g":7,"h":8}"]`, a.String())
+	require.Equal(t, `["{\"a\":1,\"b\":2}" (null) "{\"c\":3,\"d\":4}" (null) "{\"e\":5,\"f\":6}" "{\"g\":7,\"h\":8}"]`, a.String())
```
yevgenypats, line 40: That doesn't seem right. `String()` should return a string representation for json.

yevgenypats (changes requested):
Seems something is off with `JSON.String` method

types/json_test.go
candiduslynx: It isn't JSON, though, it is JSONArray.
And to be able to parse this properly we do need to use quoted strings.

candiduslynx: Specifically, see `ValueStr` test below

yevgenypats (approved):

PR #832, opened 2023-04-29
feat: Create Table abstraction on top of arrow

yevgenypats:
This is something we discussed while we were in process of migration but I wasn't able to keep everything in head so wanted to do that in two steps. Now that the 1st step is complete and arrow seems to work at least on destination side we can create a better Go abstraction (basically the same Table abstraction we have just with the `arrow.DataType` instead `ValueType` for `Type`) this should simplify any sdk and destination code very much as well make it more testable and more maintinable as the arrow schema abstraction is too low level to work with on the  in the "day-2-day".


internal/memdb/memdb.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/apache/arrow/go/v12/arrow/array"
 	"github.com/cloudquery/plugin-sdk/v2/plugins/destination"
 	"github.com/cloudquery/plugin-sdk/v2/schema"
+	"github.com/cloudquery/plugin-sdk/v2/schemav2"
```
candiduslynx, line 15: it's already v2, but we wanted to make another plugin-sdk bump for the source side change.
I'd go with
```suggestion
	"github.com/cloudquery/plugin-sdk/v3/schema"
```

yevgenypats: no need to do that yet. it's on the destination side and only we have destinations so we can keep it as `v2`

candiduslynx: then `github.com/cloudquery/plugin-sdk/v2/schema/arrow`?

yevgenypats: I think `schemav2` is fine as it's a second version of the schema and we want to have an option to access an old one. we can improve this in the future with better versioning but the codebase is not there yet.

hermanschaaf:
I haven't dug into this much yet, but what is the reason for needing to copy the code from `schema` to `schemav2`? Couldn't we use the `Table` struct from the `schema` package, but introduce a function to go from Arrow schema to `Table`? What changed between the old version and the new version?

yevgenypats:
> I haven't dug into this much yet, but what is the reason for needing to copy the code from `schema` to `schemav2`? Couldn't we use the `Table` struct from the `schema` package, but introduce a function to go from Arrow schema to `Table`? What changed between the old version and the new version?

@hermanschaaf  Basically the reason is the `Type` and we need to be able to use the old one because when we introduce it on the source side we need to still be able to go to `v1` `Table` and send it over the wire

yevgenypats:
maybe we can do `schema` and deprecate the old one but that would be a bit more painful. we will need to bump to `v2` and then import `v3`

schema/table.go
```diff
@@ -301,10 +292,6 @@ func (t *Table) ValidateName() error {
 	return nil
 }
 
-func (t *Table) ToArrowSchema() *arrow.Schema {
```
hermanschaaf, line 304: We used this in quite a few places (e.g. the filetypes library), maybe we should keep it? We can mark it as deprecated instead.

I guess the idea here is to remove all the Arrow references from v1. If that's the way we want to go it's fine, we can update the couple of destinations that used it.

internal/servers/destination/v0/destinations.go
```diff
@@ -192,11 +193,8 @@ func (s *Server) DeleteStale(ctx context.Context, req *pb.DeleteStale_Request) (
 		return nil, status.Errorf(codes.InvalidArgument, "failed to unmarshal tables: %v", err)
 	}
 	SetDestinationManagedCqColumns(tables)
-	schemas := make(schema.Schemas, len(tables.FlattenTables()))
-	for i, table := range tables.FlattenTables() {
-		schemas[i] = table.ToArrowSchema()
-	}
-	if err := s.Plugin.DeleteStale(ctx, schemas, req.Source, req.Timestamp.AsTime()); err != nil {
+	tavlesV2 := TablesV1ToV2(tables)
+	if err := s.Plugin.DeleteStale(ctx, tavlesV2, req.Source, req.Timestamp.AsTime()); err != nil {
```
hermanschaaf, line 197:
```suggestion
	tablesV2 := TablesV1ToV2(tables)
	if err := s.Plugin.DeleteStale(ctx, tablesV2, req.Source, req.Timestamp.AsTime()); err != nil {
```

schemav2/resource.go
```diff
@@ -0,0 +1,136 @@
...
+	"encoding/json"
+
+	"github.com/apache/arrow/go/v12/arrow/array"
+)
+
+type Resources []*Resource
+
+// Resource represents a row in it's associated table, it carries a reference to the original item, and automatically
+// generates an Id based on Table's Columns. Resource data can be accessed by the Get and Set methods
+type Resource struct {
```
hermanschaaf, line 13: Let's update the comment on `Resource` here to talk about what it's for and how it's supposed to work with Arrow. For example, once all the fields have been set and finalized, I guess the `Build` method should be called to create an `arrow.Record` with a single row, which should then be accessed through some other method.

hermanschaaf (changes requested):
Looks good, I think this will make our lives a bit easier. Left a couple of comments; I also think, before merging, we should port this to at least one destination and source just to check that it all works and see what it looks like. Maybe we will want to make some tweaks based on that

yevgenypats:
@hermanschaaf per our discussion - this PR now might look a bit daunting but actually I simplified things and removed `source` just in the meanwhile - this way we wont accidentally merge it to source as `v3` doesn't yet support it. `dest` though should be much easier and we don't have `schema` and `schemav2` so we make the breaking change in a more goish way.

schema/table.go
```diff
@@ -80,19 +73,42 @@ type Table struct {
...
+	description, _ := tableMD.GetValue(MetadataTableDescription)
+	fields := sc.Fields()
+	columns := make(ColumnList, len(fields))
+	for i, field := range fields {
+		columns[i] = NewColumnFromArrowField(field)
+	}
+	table := &Table{
+		Name:        name,
+		Description: description,
+		Columns:     columns,
```
hermanschaaf, line 107: Couple more metadata items we'll need here I think (e.g. title, pk constraint name, is incremental)

schema/resource.go
```diff
@@ -87,56 +78,48 @@ func (r *Resource) GetItem() any {
...
-}
-
-func (r *Resource) storeCQID(value uuid.UUID) error {
-	b, err := value.MarshalBinary()
-	if err != nil {
-		return err
-	}
-	return r.Set(CqIDColumn.Name, b)
-}
+// func (r *Resource) CalculateCQID(deterministicCQID bool) error {
```
hermanschaaf, line 82: Probably need to bring this back or delete it if it was unused

internal/servers/destination/v0/schemav1v2.go
```diff
@@ -0,0 +1,245 @@
+package destination
```
hermanschaaf, line 1: This file should probably be renamed `schemav2tov3.go`

hermanschaaf:
The general idea looks good from a read-through, let's see how it goes in an actual destination. The `Resource` type also needs to be fleshed out a bit more before it can be used in a source, it looks like.

schema/resource.go
```diff
@@ -20,46 +18,41 @@ type Resource struct {
 	// internal fields
 	Table *Table
 	// This is sorted result data by column name
-	data CQTypes
+	data map[string]any
+	bldr array.Builder
```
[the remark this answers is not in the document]
yevgenypats: I kept it an interface but yeah can do `RecordBuilder` first and only change to interface if this will be needed.

schema/table.go
hermanschaaf: Yeah, title is relatively minor and no destinations use it afaik, so maybe no need to include it here. It just depends on whether we're okay with some information being lost in the conversion. Generally speaking it's used by source plugins to generate table docs.

schema/resource.go
yevgenypats: I brought it back with `panic("not implemented")` as `Resource` is here just for some of the table definition to compile but it is not used anywhere as v3 still support only destination.

PR #833, opened 2023-04-30
fix(arrow): Add missing table options

candiduslynx:


yevgenypats (changes requested):
This should be handled now here - https://github.com/cloudquery/plugin-sdk/pull/832

candiduslynx:
> This should be handled now here - #832

@yevgenypats I've looked through the #832 and didn't find the incremental & PK constraint handling.

PR #841, opened 2023-05-01
feat: Add column to test table to test PG destination pk handling

bbernays:


hermanschaaf:
Let's leave this PR open for a bit, there are a couple of other open PRs right now that are overhauling the testdata and if we merge this in now we may lose it in the ensuing merge conflict handling. (See https://github.com/cloudquery/plugin-sdk/pull/832 and https://github.com/cloudquery/plugin-sdk/pull/818)

yevgenypats:
Don't we already have a pk by default for `_cq_id` ?

bbernays:
> Don't we already have a pk by default for `_cq_id` ?

As far as I can see the CQID column is defined by the SDK, but it is the Source that actually elevates it to a PK rather than just being not null and unique

https://github.com/cloudquery/plugin-sdk/blob/2884bed21867d46847bab4660ad6be735cfa6c78/schema/meta.go#L16-L22C2

hermanschaaf (approved):

candiduslynx:
@bbernays we now use the insert & upsert tests ti diff between tables with & without PK.
Do we still need this PR or can it be safely closed?

PR #855, opened 2023-05-09
feat: Support open/close callbacks in managed destinations

disq:
Untested

(here be dragons)


yevgenypats:
I think let's discuss this live shortly, as https://github.com/cloudquery/plugin-sdk/issues/858 approach can be indeed simpler and we might save another painful breaking SDK change/upgrade.

disq:
@yevgenypats This is fully backwards compatible though.

disq:
Implemented in https://github.com/cloudquery/plugin-sdk/pull/870/commits/fcb5e56035e9f744d9b92bcbed8af6c3ff9d7007

PR #857, opened 2023-05-10
fix(writers): Assert unimplemented interface at compile-time

candiduslynx:


plugins/destination/plugin.go
```diff
@@ -33,22 +33,26 @@ type ManagedWriter interface {
...
-type UnimplementedUnmanagedWriter struct{}
+var _ UnmanagedWriter = UnimplementedUnmanagedWriter{}
 
-func (*UnimplementedManagedWriter) WriteTableBatch(context.Context, *arrow.Schema, []arrow.Record) error {
-	panic("WriteTableBatch not implemented")
-}
+type UnimplementedUnmanagedWriter struct{}
 
-func (*UnimplementedUnmanagedWriter) Write(context.Context, schema.Schemas, <-chan arrow.Record) error {
+func (UnimplementedUnmanagedWriter) Write(context.Context, schema.Tables, <-chan arrow.Record) error {
```
candiduslynx, line 51: using value receiver here as empty struct consumes no memory

yevgenypats (approved):

PR #863, opened 2023-05-12
feat: Allow testing of more Arrow types

hermanschaaf:
This adds support for testing (almost) the complete range of Arrow types in destinations.

By default, only the types previously supported by CQTypes are enabled (with the important exception of more uint and int types than before, plus float32). The TestSourceOptions struct allows destinations to declare that they want to test with more column types, including maps, structs, times, etc.

I have not included all the nullable column variations. These are a bit more difficult to do now that we are not using `arrow.Field` directly anymore, plus we can rather do that as a follow-up if and when it becomes necessary to test this.

This PR is a rebased version for v3 that replaces #818.

### DIFF
--- a/schema/column.go
+++ b/schema/column.go
@@ -14,17 +14,6 @@ type ColumnList []Column
 // resource holds the current row we are resolving the column for.
 type ColumnResolver func(ctx context.Context, meta ClientMeta, resource *Resource, c Column) error
 
-// ColumnCreationOptions allow modification of how column is defined when table is created
-type ColumnCreationOptions struct {
-	PrimaryKey bool
-	NotNull    bool
-	// IncrementalKey is a flag that indicates if the column is used as part of an incremental key.
-	// It is mainly used for documentation purposes, but may also be used as part of ensuring that
-	// migrations are done correctly.
-	IncrementalKey bool
-	Unique         bool
-}
-
 // Column definition for Table
 type Column struct {
 	// Name of column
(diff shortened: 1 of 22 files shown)


schema/testdata.go
```diff
@@ -1,131 +1,242 @@
...
-				Name: "macaddr",
-				Type: types.ExtensionTypes.Mac,
-			},
-			{
-				Name: "macaddr_array",
-				Type: arrow.ListOf(types.ExtensionTypes.Mac),
-			},
-		},
+// TestSourceOptions controls which types are included by TestSourceColumns.
+type TestSourceOptions struct {
```
disq, line 20: What about adding a shorthand to set all options to `true`? Something like `func AllTestSourceOptions() TestSourceOptions` and then every field set to true.

candiduslynx: I'd even swap for `SkipTypeX` semantics, so that the default would be everything and if not, the destination has to be mindful about it (just like of all the skip modes for writing)

schema/testdata.go
```diff
@@ -141,119 +252,33 @@ type GenTestDataOptions struct {
...
-				bldr.Field(i).(*array.ListBuilder).Append(true)
-				bldr.Field(i).(*array.ListBuilder).ValueBuilder().(*types.MacBuilder).Append(mac)
-				mac, err = net.ParseMAC("11:22:33:44:55:66")
-				if err != nil {
-					panic(err)
-				}
-				bldr.Field(i).(*array.ListBuilder).ValueBuilder().(*types.MacBuilder).Append(mac)
-			} else {
-				panic("unknown type: " + c.Type.String() + " column: " + c.Name)
+			example := getExampleJSON(c.Name, c.Type, opts)
```
yevgenypats, line 271: I think you mentioned it but I can't recall, why you went with `json` instead of regular `append` to builder?

schema/testdata.go
yevgenypats: seems that this testing suite is growing I would do two things, one is more important then the other imo:

1) I would use `Skip` - this way everything is included by default as usually destinations should support everything, unlike modes of operation the SDK doesn't error out (at least for now) if a mode is not supported so if a destinations updates to a newer SDK with new arrow types this would error out in "prod" instead of tests.
2) More of a preference - I think mostly we used `WithOption()` pattern so I would stick to that but that's not a must.

yevgenypats (changes requested):
Looks good. Had mostly one comment around `Skip` vs `Include`

schema/testdata.go
hermanschaaf: Both can work, and I don't recall the exact reason anymore either 😅  I think it was just easier and cleaner to split it into a separate recursive function that can also handle composite types, and doesn't require any type switches. Using JSON is also a common pattern in the Arrow library testing files, which is where I got the idea.

schema/column.go
```diff
@@ -27,6 +27,7 @@ type ColumnCreationOptions struct {
 
 // Column definition for Table
 type Column struct {
+	arrow.Field
```
candiduslynx, line 30: should be removed?

hermanschaaf: yes, thanks

hermanschaaf:
Updated to use `Skip` and options columns. Most destinations will probably start failing their tests when we merge this now, but we'll just have to update them to skip the types they don't support yet (until they do support them)

yevgenypats (approved):

PR #864, opened 2023-05-14
feat: Support sources in SDK V3

yevgenypats:
Few notes:

- Most of the code here was removed in https://github.com/cloudquery/plugin-sdk/pull/854 to prevent accidental use before it was ready
- Sources moved to use the new `schema.Table` which means use of `arrow.DataType` instead of our old `ValueType`
- Protocol in sources and destinations are upgraded to use native arrow format (sources v2 and dest v1) - https://github.com/cloudquery/plugin-pb-go
- Introduces a new `scalar` package which is mostly used by our "managed" sources and basically our old cqtypes just now using arrow types (which also support nesting). This will hopefully go upstream one day but I don't think we should block on that as the upstream package is not a fit right now and will take time for such refactor to make it upstream. 

Follow-up PRs with example sources and dest will follow soon.

IMPORTANT: all destinations should be released first as we decided new sources will work only with new destination to avoid un-necessary backward compatibility maintenance.  

Example PRs:
- PostgreSQL: https://github.com/cloudquery/cloudquery/pull/10783
- GCP: https://github.com/cloudquery/cloudquery/pull/10784
- CLI: https://github.com/cloudquery/cloudquery/pull/10785


scalar/float.go
```diff
@@ -0,0 +1,220 @@
...
+	if val == nil {
+		s.Valid = false
+		return nil
+	}
+
+	switch value := val.(type) {
+	case int8:
+		s.Value = float32(value)
+	case int16:
+		if value > math.MaxInt8 {
```
hermanschaaf, line 51: I don't really understand why this check is here - what's wrong with `int16` being given to `float32`? Seems like it should work regardless of being bigger than `math.MaxInt8`? (Same for the cases below)

scalar/float_test.go
```diff
@@ -0,0 +1,39 @@
+package scalar
+
+import "testing"
+
+func TestFloat64Set(t *testing.T) {
```
hermanschaaf, line 5: Seems like we're missing a Float32 test

yevgenypats:
Going to put an auto-merge here so it will be easier to test other plugins

yevgenypats:
I did some small changes to the testing code and commented some stuff out so we can more easily follow-up with fixes and review them given the large code change here (due to the copy from a previous PR).

schema/testdata.go
```diff
@@ -155,9 +155,9 @@ func TestSourceColumns(testOpts ...func(o *TestSourceOptions)) []Column {
 		compositeColumns = append(compositeColumns, listOfColumns(basicColumnsWithExclusions)...)
 	}
 
-	if !opts.SkipMaps {
-		compositeColumns = append(compositeColumns, mapOfColumns(basicColumnsWithExclusions)...)
-	}
+	// if !opts.SkipMaps {
+	// 	compositeColumns = append(compositeColumns, mapOfColumns(basicColumnsWithExclusions)...)
+	// }
```
hermanschaaf, line 160: Why not just use `SkipMaps` to skip this?

yevgenypats: It was failing the plugin tests themselves. seems there is an error in `arrow` compare function.

hermanschaaf: Oh weird, it was passing for me yesterday (but I did find it a bit suspicious 😄 )

PR #869, opened 2023-05-15
feat(schema): Embed column creation options

candiduslynx:


schema/meta.go
```diff
@@ -14,45 +14,48 @@ type ClientMeta interface {
...
 }
 var CqSourceNameColumn = Column{
-	Name:        "_cq_source_name",
-	Type:        arrow.BinaryTypes.String,
+	Name:        CqSourceNameField.Name,
+	Type:        CqSourceNameField.Type,
 	Description: "Internal CQ row that references the source plugin name data was retrieved",
 }
 
 var CqIDField = arrow.Field{
```
yevgenypats, line 42: I think this can be removed

schema/meta.go
```diff
@@ -14,45 +14,48 @@ type ClientMeta interface {
...
-	Type: types.ExtensionTypes.UUID,
-	Metadata: arrow.MetadataFrom(map[string]string{
-		MetadataUnique: MetadataTrue,
-	}),
+	Name:     "_cq_id",
+	Type:     types.ExtensionTypes.UUID,
+	Metadata: arrow.MetadataFrom(map[string]string{MetadataUnique: MetadataTrue}),
 }
+
+var CqParentIDField = arrow.Field{
```
yevgenypats, line 48: I think this can be deleted as well

yevgenypats (changes requested):
Looks good. small comment.

schema/meta.go
candiduslynx: done

yevgenypats (changes requested):
Why rename `ColumnCreationOptions ` to `CreationOptions` if we are doing breaking change we can remove the struct all together

candiduslynx:
> Why rename `ColumnCreationOptions ` to `CreationOptions` if we are doing breaking change we can remove the struct all together

I think the options should be grouped together, so having them in a separate struct might be better.
The impact of removing can be seen in 1615ad9fe0b0f39f75003899aa3c4f2b8e040712

yevgenypats (approved):

yevgenypats (approved):

Issue #858, opened 2023-05-11
feat: Extract `ManagedWriter` to helper package, use single `Write()` for destinations


disq:
### Are you positive this feature should be implemented in the SDK?

- [X] I'm positive this is an SDK feature

### Which problem is this feature request solving?

Currently the `ManagedWriter` is built-in in the SDK which makes it impossible to extend or swap it with another implementation without changing/breaking the SDK.

### Describe the solution you'd like

Let's make the destinations specify their own 'custom writers' if they want to, something like: `destination.WithManagedWriter(<some constructor or interface>)` so that it can be swapped out or injected from an external package. We can still supply the traditional ManagedWriter as an helper package in the SDK.

### Pull request (optional)

- [x] I can submit a pull request

candiduslynx:
This needs to be done after moving the PK dup filter to the source side, as the managed writer is the only place ATM to do so.

disq:
@candiduslynx It can be an option to the constructor, dedup or not, so the choice remains in the destination side.

yevgenypats:
Done in SDK V4

PR #870, opened 2023-05-15
feat(destination)!: Extract batching writer to separate package

disq:
Implements #858

Usage:
```go
import "github.com/cloudquery/plugin-sdk/v3/plugins/destination/batchingwriter"

func main() {
	p := destination.NewPlugin("file", plugin.Version, client.New,
		destination.WithManagedWriter(batchingwriter.New(batchingwriter.WithDedupPK(false))),
	)
	serve.Destination(p, serve.WithDestinationSentryDSN(sentryDSN))
}
```


plugins/destination/batchingwriter/options.go
```diff
@@ -0,0 +1,34 @@
...
+
+type Option func(*Batching)
+
+func WithLogger(l zerolog.Logger) Option {
+	return func(w *Batching) {
+		w.logger = l
+	}
+}
+
+func WithDedupPK(dedup bool) Option {
```
candiduslynx, line 17: I'd make it `WithNoPKDedup()` so that it'll be obvious that the default is to dedup (or removed altogether)

plugins/destination/plugin.go
```diff
@@ -65,6 +49,12 @@ type Client interface {
 	Close(ctx context.Context) error
 }
 
+type BatchingWriter interface {
```
disq, line 52: This can just be an `UnmanagedWriter` but kept as this for development/clarity purposes.
`type UnmanagedWriterFunc func(ManagedWriter) UnmanagedWriter` line would look a bit weird.

plugins/destination/batchingwriter/options.go
disq: In that case there would be no explicit way of specifying what you want though, depending on the default behavior set by that specific SDK version? (As `WithNoPKDedup(bool)` would be funny)

plugins/destination/plugin.go
```diff
@@ -12,50 +11,35 @@ import (
...
-
-var _ ManagedWriter = UnimplementedManagedWriter{}
-
-func (UnimplementedManagedWriter) WriteTableBatch(context.Context, *schema.Table, []arrow.Record) error {
-	panic("WriteTableBatch not implemented")
-}
-
 type UnmanagedWriter interface {
-	Write(ctx context.Context, tables schema.Tables, res <-chan arrow.Record) error
+	Write(context.Context, specs.Source, schema.Tables, time.Time, <-chan arrow.Record) error
```
disq, line 17: These (`Source` and `Time`) were added to both writer signatures to somewhat unify how we push the data (only difference being a channel vs. a slice) and could be useful eventually. Also helps us to [remove](https://github.com/cloudquery/plugin-sdk/pull/870/files#diff-e8ee472a920c34b2e1ac3e57c75020a57a5ec9691465ebee51927e3496bfa9a6) `unmanaged_writer.go`

internal/memdb/memdb_test.go
```diff
@@ -187,12 +188,14 @@ func TestPluginInit(t *testing.T) {
 		"test",
 		"development",
 		func(ctx context.Context, logger zerolog.Logger, s specs.Destination) (destination.Client, error) {
+			// TODO this won't work because we don't set these back on the spec
```
disq, line 191: This still needs addressing (we need to be able to change the default batching limits from the dest spec), will do in a later commit.

plugins/destination/batchingwriter/batchingwriter.go
```diff
@@ -0,0 +1,209 @@
...
+	workers     map[string]*worker
+	workersLock *sync.Mutex
+
+	dedupPK        bool
+	batchSize      int64
+	batchSizeBytes int64
+	batchTimeout   time.Duration
+}
+
+func New(opts ...Option) destination.BatchingWriterFuncFunc {
```
disq, line 34: Crazy I know

plugins/destination/batchingwriter/batchingwriter.go
```diff
@@ -0,0 +1,209 @@
...
+			opt(w)
+		}
+		// Set the spec batch sizes again, in case the options changed them. Used in testing
+		spec.BatchSize = int(w.batchSize)
+		spec.BatchSizeBytes = int(w.batchSizeBytes)
+		spec.SetDefaults(
+			10000,
+			5*1024*1024, // 5 MiB
+		)
+		w.batchSize, w.batchSizeBytes = int64(spec.BatchSize), int64(spec.BatchSizeBytes)
```
disq, line 58: Crazy I know

internal/memdb/memdb_test.go
```diff
@@ -172,30 +171,3 @@ func TestOnWriteCtxCancelled(t *testing.T) {
 		t.Fatal(err)
 	}
 }
-
-func TestPluginInit(t *testing.T) {
```
disq, line 176: This was moved to https://github.com/cloudquery/plugin-sdk/pull/870/files#diff-f4ba5b8e6a3a7d42e936c4f99905060cbf7cace1ec07b3c284d6a2fda303dd8e

plugins/destination/batchingwriter/batchingwriter.go
```diff
@@ -0,0 +1,268 @@
...
+
+func (w *Batching) Write(ctx context.Context, sourceSpec specs.Source, tables schema.Tables, syncTime time.Time, res <-chan arrow.Record) error {
+	workers := make(map[string]*worker, len(tables))
+
+	w.workersLock.Lock()
+	for _, table := range tables {
+		table := table
+		metrics := &destination.Metrics{}
+		w.metricsLock.Lock()
+		w.metrics[table.Name] = metrics
```
disq, line 174: This was broken/missing from main. Also we now do `atomic. LoadUint64` when tallying metrics.

plugins/destination/plugin_testing.go
```diff
@@ -64,6 +64,8 @@ type PluginTestSuiteTests struct {
 
 	MigrateStrategyOverwrite MigrateStrategy
 	MigrateStrategyAppend    MigrateStrategy
+
+	Validate func(t *testing.T, plugin *Plugin, destination specs.Destination)
```
disq, line 68: New method to do final checks after each test run.

plugins/destination/batchingwriter/batchingwriter.go
```diff
@@ -0,0 +1,268 @@
...
+}
+
+// OpenCloseWriter is an optional interface that can be implemented by a Client which already implements destination.ManagedWriter.
+type OpenCloseWriter interface {
+	OpenTable(ctx context.Context, sourceSpec specs.Source, table *schema.Table, syncTime time.Time) error
+	CloseTable(ctx context.Context, sourceSpec specs.Source, table *schema.Table, syncTime time.Time) error
+	destination.ManagedWriter
+}
+
+func New(opts ...Option) destination.BatchingWriterFuncFunc {
```
candiduslynx, line 44: `destination.BatchingWriterFuncFunc` is odd

plugins/destination/batchingwriter/batchingwriter.go
```diff
@@ -0,0 +1,268 @@
...
+	batchSizeBytes int64
+	batchTimeout   time.Duration
+
+	// underlyingOCW is set if the given underlyingWriter implements OpenCloseWriter
+	underlyingOCW OpenCloseWriter
+}
+
+// OpenCloseWriter is an optional interface that can be implemented by a Client which already implements destination.ManagedWriter.
+type OpenCloseWriter interface {
+	OpenTable(ctx context.Context, sourceSpec specs.Source, table *schema.Table, syncTime time.Time) error
```
candiduslynx, line 39: why do we need source spec here?

plugins/destination/plugin.go
candiduslynx: I don't think passing source spec to the destination os a good idea.
IMO, it's the source job to provide the data and the destination has to store it without looking at what the source spec was.

plugins/destination/plugin_unimplemented.go
```diff
@@ -0,0 +1,65 @@
...
+	return nil // Special case, we don't want to error here
+}
+
+type UnimplementedClose struct{}
+
+func (UnimplementedClose) Close(context.Context) error {
+	return nil // Special case, we don't want to error here
+}
+
+type UnimplementedClient struct {
```
candiduslynx, line 52: We can just use a single `UnimplementedClient` empty struct and move all functions to it.
This way we simply embed it into the destination client and implement only the required methods.

candiduslynx (changes requested):

plugins/destination/batchingwriter/batchingwriter.go
disq: source name is used to separate between data from separate sources and written to separate streams/files.

plugins/destination/plugin_unimplemented.go
disq: I don't remember why I separated them, but some of the functionality shouldn't error when not implemented, maybe that was the reason behind this... I'll refactor once we get some more reviews of that's OK. (We could also do an ErrNotImplemented type thing and then make the SDK ignore those errors maybe)

plugins/destination/batchingwriter/batchingwriter.go
candiduslynx: could sourceName work instead?

disq: it can, I just wanted to make sure we don't need to change the interface soon for another source property.

plugins/destination/batchingwriter/batchingwriter.go
disq: @candiduslynx this is due to the initialization order... there's a chicken-egg problem :)

disq:
Closing for sdk v4.

PR #872, opened 2023-05-15
fix: Fix test column generation

hermanschaaf:
There was a bug in the last implementation causing many types to be missing; it's always that final refactor you do 😄 

This time I added a sanity check for the testdata generation to make sure it can't be catastrophically broken without us knowing about it.

### DIFF
--- a/schema/testdata_test.go
+++ b/schema/testdata_test.go
@@ -31,7 +31,7 @@ func TestTestSourceColumns(t *testing.T) {
 		t.Fatal("expected no times when WithTestSourceSkipTimes is used")
 	}
 	if skipAll.Get("timestamp_us") == nil {
-		t.Fatal("expected no microsecond timestamps even when WithTestSourceSkipTimestamps is used")
+		t.Fatal("expected microsecond timestamps even when WithTestSourceSkipTimestamps is used (microsecond timestamps must always be supported)")
 	}
 	if skipAll.Get("timestamp_ns") != nil {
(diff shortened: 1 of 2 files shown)


schema/testdata_test.go
```diff
@@ -0,0 +1,42 @@
...
+		t.Fatal("expected no maps when WithTestSourceSkipMaps is used")
+	}
+	if skipAll.Get("date32") != nil {
+		t.Fatal("expected no date32 when WithTestSourceSkipDates is used")
+	}
+	if skipAll.Get("time32") != nil {
+		t.Fatal("expected no times when WithTestSourceSkipTimes is used")
+	}
+	if skipAll.Get("timestamp_us") == nil {
+		t.Fatal("expected no microsecond timestamps even when WithTestSourceSkipTimestamps is used")
```
candiduslynx, line 34:
```suggestion
		t.Fatal("expected no microsecond timestamps when WithTestSourceSkipTimestamps is used")
```

candiduslynx (changes requested):

schema/testdata_test.go
hermanschaaf: Thanks; fixed the message (but in a slightly different way). We must always test microsecond timestamps, because they were supported by CQTypes

disq (approved):

yevgenypats (approved):

yevgenypats (approved):

PR #876, opened 2023-05-16
fix(test): Remove extra `v2/schema` import

candiduslynx:


schema/testdata.go
```diff
@@ -361,7 +360,7 @@ func getExampleJSON(colName string, dataType arrow.DataType, opts GenTestDataOpt
 	if arrow.IsListLike(dataType.ID()) {
 		if dataType.ID() == arrow.MAP {
 			k := getExampleJSON(colName, dataType.(*arrow.MapType).KeyType(), opts)
-			v := getExampleJSON(colName, dataType.(*arrow.MapType).ValueType().Field(1).Type, opts)
+			v := getExampleJSON(colName, dataType.(*arrow.MapType).ItemType(), opts)
```
hermanschaaf, line 363: This didn't work when I tried it initially, but maybe it works now? Has the implementation changed? Have you tested it?

hermanschaaf: Oh no, I'm probably misremembering: I think it was that `ValueType()` didn't return what I expected in the same way as `KeyType()`, so yeah maybe `ItemType()` is the way to go. I'll need to check the implementation though

hermanschaaf: Yeah looks fine... bit confusing naming, but anyway :)

hermanschaaf (approved):

schema/testdata.go
candiduslynx: Yeah, I had to dig a bit during CH implementation for it.
Basically, `Map(keyType, itemType)` is `List(Struct("key" keyType, "value" nullable(valueType)))` in Arrow.
Then we have 3 different sets of functions:
* `KeyType` & `KeyField` for key element
* `ItemType` & `ItemField` for value element
* `ValueType` & `ValueField` for `list element` (as map = list(struct(...)))

yevgenypats (approved):

PR #882, opened 2023-05-16
fix: Flatten V2 tables

yevgenypats:
A bit more footgun in old protocol.

### DIFF
--- a/internal/servers/destination/v0/destinations.go
+++ b/internal/servers/destination/v0/destinations.go
@@ -59,7 +59,7 @@ func (s *Server) Migrate(ctx context.Context, req *pb.Migrate_Request) (*pb.Migr
 	if err := json.Unmarshal(req.Tables, &tablesV2); err != nil {
 		return nil, status.Errorf(codes.InvalidArgument, "failed to unmarshal tables: %v", err)
 	}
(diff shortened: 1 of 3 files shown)


internal/servers/destination/v0/destinations.go
```diff
@@ -59,7 +59,7 @@ func (s *Server) Migrate(ctx context.Context, req *pb.Migrate_Request) (*pb.Migr
 	if err := json.Unmarshal(req.Tables, &tablesV2); err != nil {
 		return nil, status.Errorf(codes.InvalidArgument, "failed to unmarshal tables: %v", err)
 	}
-	tables := TablesV2ToV3(tablesV2)
+	tables := TablesV2ToV3(tablesV2).FlattenTables()
```
hermanschaaf, line 62: If this is the destination server, presumably it got the list from the source - so shouldn't the tables already be flattened? I know there can be issues if you flatten a list of tables twice, so we either need to be 100% sure or make `FlattenTables` idempotent :)

yevgenypats: Done.

hermanschaaf (approved):

PR #884, opened 2023-05-16
fix(tables): Flatten stripping relations

candiduslynx:


disq (approved):

candiduslynx (commented):
@yevgenypats now FlattenTables will strip the relations info, so, if the code calls flattening it has to treat the tables as a flat list instead of forest-like structure

schema/table.go
```diff
@@ -228,19 +229,13 @@ func (tt Tables) FilterDfs(tables, skipTables []string, skipDependentTables bool
...
 	tables := make(Tables, 0, len(tt))
 	for _, t := range tt {
-		tables = append(tables, t)
+		table := *t
+		table.Relations = nil
+		tables = append(tables, &table)
 		tables = append(tables, t.Relations.FlattenTables()...)
 	}
-	tableNames := make(map[string]bool)
-	dedupedTables := make(Tables, 0, len(tables))
```
yevgenypats, line 235: I believe we still want this dedup safety because in the current way there is nothing that protects us from duplicates or another functions that was used to flatten it.

yevgenypats (changes requested):
looks good. one nit.

schema/table.go
candiduslynx: @yevgenypats I don't think that we need the dedup logic here.
If we want to enforce `Tables` to be deduped, we should change the type to be `map[string]*Table`.

Like, if someone creates a list of tables with the same names, they should both be present (as they might differ and we don't have any way of knowing what def was intended).
Alternatively, enforcing uniqueness by making `Tables` be `map[string]*Table` (and it could be supported by a helper deduping func) could improve the contract.

candiduslynx: I've returned the dedup in 436357140a465b345fe612ffc7621c1bab7a10f5 + added tests to verify

yevgenypats (approved):

PR #886, opened 2023-05-17
fix(destination): Don't duplicate tables to be removed

candiduslynx:
BEGIN_COMMIT_OVERRIDE
fix(destination): Don't duplicate tables to be removed (#886)

fix(arrow): `schema.Table` <-> `arrow.Schema` conversion (#886)
END_COMMIT_OVERRIDE

### DIFF
--- a/internal/memdb/memdb.go
+++ b/internal/memdb/memdb.go
@@ -19,7 +19,6 @@ import (
 // client is mostly used for testing the destination plugin.
 type client struct {
 	spec          specs.Destination
-	sourceSpec    specs.Source // used to verify that we don't remove incremental tables when src backend != none
 	memoryDB      map[string][]arrow.Record
 	tables        map[string]*schema.Table
 	memoryDBLock  sync.RWMutex
@@ -72,16 +71,6 @@ func NewClient(_ context.Context, _ zerolog.Logger, spec specs.Destination) (des
 
 func NewClientErrOnNew(context.Context, zerolog.Logger, specs.Destination) (destination.Client, error) {
 	return nil, fmt.Errorf("newTestDestinationMemDBClientErrOnNew")
-}
-
-func NewClientWithSrcSpec(srcSpec specs.Source) destination.NewClientFunc {
-	return func(ctx context.Context, logger zerolog.Logger, spec specs.Destination) (destination.Client, error) {
-		c, err := NewClient(ctx, logger, spec)
-		if c != nil {
-			c.(*client).sourceSpec = srcSpec
-		}
-		return c, err
-	}
 }
 
 func (c *client) overwrite(table *schema.Table, data arrow.Record) {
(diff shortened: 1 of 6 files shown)


schema/table.go
```diff
@@ -116,21 +116,20 @@ func NewTableFromArrowSchema(sc *arrow.Schema) (*Table, error) {
...
+		Description:      description,
+		PkConstraintName: constraintName,
+		Columns:          columns,
 	}
-	if constraintName, found := tableMD.GetValue(MetadataConstraintName); found {
-		table.PkConstraintName = constraintName
-	}
-	if title, found := tableMD.GetValue(MetadataIncremental); found {
-		table.Title = title
+	if isIncremental, found := tableMD.GetValue(MetadataIncremental); found {
```
hermanschaaf, line 131: We do still need to set the title here as well

hermanschaaf: Oh actually sorry no, I think we decided not to support Title because it's not used by destinations at this stage

internal/memdb/memdb.go
```diff
@@ -205,6 +216,10 @@ func (c *client) Close(context.Context) error {
 
 func (c *client) DeleteStale(ctx context.Context, tables schema.Tables, source string, syncTime time.Time) error {
 	for _, table := range tables {
+		if c.sourceSpec.Backend != specs.BackendNone && table.IsIncremental {
```
hermanschaaf, line 219: I'd rather have this assertion in a test than in memdb itself, if possible? The assertion might look different, but it would basically be asserting that an incremental table didn't have all its records cleared after a write

candiduslynx: I'll try to add this as an assertion then

plugins/destination/plugin_testing_overwrite_delete_stale.go
```diff
@@ -117,19 +120,19 @@ func (*PluginTestSuite) destinationPluginTestWriteOverwriteDeleteStale(ctx conte
...
 
 	// we expect the only resource returned to match the updated resource we wrote
-	if !array.RecordApproxEqual(updatedResources, resourcesRead[0]) {
-		diff := RecordDiff(updatedResources, resourcesRead[0])
+	if !array.RecordApproxEqual(updatedResources[0], resourcesRead[0]) {
+		diff := RecordDiff(updatedResources[0], resourcesRead[0])
 		return fmt.Errorf("after delete stale expected resource to be equal. diff: %s", diff)
 	}
 
 	// we expect the incremental table to still have 2 resources, because delete-stale should
```
hermanschaaf, line 128:
```suggestion
	// we expect the incremental table to still have 3 resources, because delete-stale should
```

hermanschaaf (approved):
The fix looks good, just one more update so the comment is in sync with the code

yevgenypats (approved):

PR #889, opened 2023-05-19
fix: Add null-row case for append-only tests

hermanschaaf:


schema/testdata.go
```diff
@@ -329,8 +329,7 @@ func GenTestData(table *Table, opts GenTestDataOptions) []arrow.Record {
 			if nullRow && !c.NotNull && !c.PrimaryKey &&
 				c.Name != CqSourceNameColumn.Name &&
 				c.Name != CqSyncTimeColumn.Name &&
-				c.Name != CqIDColumn.Name &&
-				c.Name != CqParentIDColumn.Name {
```
hermanschaaf, line 333: This wasn't a bug as such, but I don't see a reason to prevent CQParentIDColumn from being set to null, as in many cases it will be null

yevgenypats: How a parentIDColumn can be null?

hermanschaaf: For all top-level tables

yevgenypats (approved):

PR #892, opened 2023-05-19
feat: Refactor test options and allow skipping of nulls in lists

hermanschaaf:
- Skipping of null values in lists when doing comparisons is necessary for BigQuery tests to pass right now, because BigQuery does not support null values in repeated columns. To keep the migration to Arrow backwards-compatible, we are opting to strip nulls in that case (for now), but we still want our automated tests to pass.
- This also refactors the test options. Previously you had to call `destination.PluginTestSuiteRunner` with options from the `schema` package, which is a bit unusual and counter-intuitive, and makes it hard to extend with options that relate specifically to the test suite runner. This changes it so that the options passed in are from the `destination` package and passed through to the `schema` package. It's a breaking change, but this functionality hasn't been out for long, will be easy to update, and only affects tests, so I'm hoping users will forgive us for this one.

### DIFF
--- a/internal/memdb/memdb_test.go
+++ b/internal/memdb/memdb_test.go
@@ -112,7 +112,7 @@ func TestOnWriteError(t *testing.T) {
 	if err := p.Init(ctx, getTestLogger(t), specs.Destination{}); err != nil {
 		t.Fatal(err)
 	}
-	table := schema.TestTable("test")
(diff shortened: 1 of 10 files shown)


yevgenypats (approved):
Nice. Need that also for mongo but for different option - changing the timestamp precision (mongodb only supports milliseconds).

PR #894, opened 2023-05-21
fix: Inet extension MarshalJSON

yevgenypats:
Right now there is an inconsistency in our extensions: Some of them are using `BinaryType` as the underlying storage and some `StringType`.

There is another issue in the parquet reader that seems not to like extensions with `StringType` but we can solve this bug if we want our Extensions to be of `StringType` rather then `BinaryType`.

Right now I changed `Inet` to `BinaryType` as well for consistency but we have a chance to change all of them to `StringType` before we migrate sources without breaking backward compatibility because we didn't send anything over the wire yet - so right it's just an implementation change rather then a breaking change.

### DIFF
--- a/types/inet.go
+++ b/types/inet.go
@@ -25,11 +25,11 @@ func (b *InetBuilder) Append(v *net.IPNet) {
 		b.AppendNull()
 		return
 	}
-	b.ExtensionBuilder.Builder.(*array.StringBuilder).Append(v.String())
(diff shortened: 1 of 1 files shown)


yevgenypats:
self merging this one so I can open a few PRs for filetypes in the meanwhile but can change back all extensions to StringType as mentioned in the previous message due to it's not being a breaking change for now (as not being send over the wire).

PR #896, opened 2023-05-22
fix: Better handling for Arrow type strings in docs

hermanschaaf:
This fixes the rendering of some Arrow types in the docs, as well as in the JSON


disq (approved):

plugins/source/docs.go
```diff
@@ -229,6 +234,102 @@ func formatMarkdown(s string) string {
...
+				if i > 0 {
+					s += ","
+				}
+				s += f.Name + ":" + formatType(f.Type)
+			}
+			return s + ">"
+		case arrow.LIST:
+			return "list<" + formatType(v.(*arrow.ListType).Elem()) + ">"
+		case arrow.LARGE_LIST:
+			return "large_list<" + formatType(v.(*arrow.LargeListType).Elem()) + ">"
```
candiduslynx, line 251: you lose the info about elem nullability here (and it might be vital to know if the elem can be nullable or not (think CH destination))

candiduslynx (changes requested):
I do feel that the Arrow Types are OK in terms of defining what is actually going on.
So, maybe just do an upstream PR to remove the timezone if the tz is `UTC`? (it's the default, so no need to actually print it out if not explicitly asked for)

Additionally, there's an issue with unhandled nested type nullability in this proposal.

hermanschaaf:
@candiduslynx Updated. The main thing I wanted to fix was that JSON was being printed as `extension_type<fixed_byte_list>` or something along those lines; turns out this was part of its `String()` definition, so I could fix it there. I don't love the Arrow `String()` implementation, as it seems inconsistent between nested types and `utf8` is not as good a name as `string`, but I think we can live with these things for now.

plugins/source/docs.go
```diff
@@ -143,7 +147,7 @@ func (p *Plugin) jsonifyTables(tables schema.Tables) []jsonTable {
 		for c, col := range table.Columns {
 			jsonColumns[c] = jsonColumn{
 				Name:             col.Name,
-				Type:             col.Type.String(),
+				Type:             formatType(col.Type),
```
candiduslynx, line 150:
```suggestion
				Type:             col.Type.String(),
```

candiduslynx (changes requested):

plugins/source/docs.go
candiduslynx: and remove the `formatType` func

candiduslynx (approved):

yevgenypats (approved):

PR #897, opened 2023-05-22
feat: Use StringType for extensions

yevgenypats:
This is a follow-up to this - https://github.com/cloudquery/plugin-sdk/pull/894.

During our migration work I saw quite a few pieces of code where we transform extensions to stringType back and forth so it might make sense to use them with string as the underlying storage type given they have a string representation anyway. and we can use binaryType only for those that are real binary types and have no reasonable string representation.

### DIFF
--- a/types/json.go
+++ b/types/json.go
@@ -26,7 +26,7 @@ func (b *JSONBuilder) AppendBytes(v []byte) {
 		return
 	}
 
-	b.ExtensionBuilder.Builder.(*array.BinaryBuilder).Append(v)
+	b.ExtensionBuilder.Builder.(*array.StringBuilder).Append(string(v))
 }
 
 func (b *JSONBuilder) Append(v any) {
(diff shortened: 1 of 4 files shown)


hermanschaaf (approved):

types/json.go
```diff
@@ -26,7 +26,7 @@ func (b *JSONBuilder) AppendBytes(v []byte) {
 		return
 	}
 
-	b.ExtensionBuilder.Builder.(*array.BinaryBuilder).Append(v)
```
candiduslynx, line 29: I do think we need Binary here

types/mac.go
```diff
@@ -143,24 +143,28 @@ func (a *MACArray) Value(i int) net.HardwareAddr {
 	if a.IsNull(i) {
 		return nil
 	}
-	return net.HardwareAddr(a.Storage().(*array.Binary).Value(i))
```
candiduslynx, line 146: Here Binary is more suitable as well

types/uuid.go
```diff
@@ -51,14 +51,14 @@ func (b *UUIDBuilder) AppendValues(v []uuid.UUID, valid []bool) {
...
-	data := make([][]byte, len(v))
+	data := make([]string, len(v))
 	for i := range v {
 		if len(valid) > 0 && !valid[i] {
 			continue
 		}
-		data[i] = v[i][:]
+		data[i] = v[i].String()
 	}
-	b.ExtensionBuilder.Builder.(*array.FixedSizeBinaryBuilder).AppendValues(data, valid)
```
candiduslynx, line 61: I do think fixed size binary is more suitable here, per the fact that we know that the len will match uuid.UUID len and we don't need to make the check for that

candiduslynx (changes requested):

types/json.go
yevgenypats: why?

types/mac.go
yevgenypats: why? it is always string representable and in most places it is stored at string.

candiduslynx: Hm, with this I'm actually OK to go in any direction (as the other platforms/tools might not guarantee the same byte layout).

types/json.go
candiduslynx: Although it is true, we do want to save the extra cycles of casting from/to json.RawMessage when we can.
Now it will cause the extra conversions and will result in unneeded allocations as well.

types/uuid.go
disq: I agree, also parquet has built-in uuid logical-type which is based on fixed size binary: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#uuid

PR #901, opened 2023-05-24
feat(transform): Nullability

candiduslynx:
This change marks the non-nullable in Go fields as non-nullable in the schema.

> **Note**: no extra logic like PK -> NotNullable was added

Additional changes: split the `struct.go` into separate files for readability.


candiduslynx:
> making more things NotNull just makes more breaking changes in any future schema change like dropping a column which wouldn't be breaking if it's Null By default

We usually don't drop columns, as the specs/structs tend to get more fields rather than drop them.

PR #913, opened 2023-05-29
feat(test): Add `AllowNull` option for test data

candiduslynx:
This allows to control whether some of the data types should be left non-nullable.
Inspired by ClickHouse and the fact that [Arrays, Maps & Tuples are non-nullable](clickhouse.com/docs/en/sql-reference/data-types/nullable).


plugins/destination/plugin_testing.go
```diff
@@ -101,10 +101,24 @@ func getTestLogger(t *testing.T) zerolog.Logger {
...
+
+	// AllowNull is a custom func to determine whether a data type may be correctly represented as null.
+	// Destinations that have problems representing some data types should provide a custom implementation here.
+	// If this param is empty, the default is to allow all data types to be nullable.
+	AllowNull func(arrow.DataType) bool
+
 	schema.TestSourceOptions
 }
 
+func WithTestSourceAllowNull(allowNull func(arrow.DataType) bool) func(o *PluginTestSuiteRunnerOptions) {
```
disq, line 116: I would do

```go
type AllowNullFunc func(arrow.DataType) bool
```

and then this becomes

```go
func WithTestSourceAllowNullFunc(allowNullFunc AllowNullFunc) func(o *PluginTestSuiteRunnerOptions)
```

Then we can just add

```go
func WithTestSourceAllowNulls(...arrow.DataType) func(o *PluginTestSuiteRunnerOptions)
```

Which can use a generic `slice.Contains[arrow.DataType]` inside. Could also add a generator function that takes an `...arrow.DataType` and returns a bool-returning contains function but it would be unnecessary if we're already using a generic contains fn.

plugins/destination/plugin_testing.go
```diff
@@ -101,10 +101,24 @@ func getTestLogger(t *testing.T) zerolog.Logger {
...
 type PluginTestSuiteRunnerOptions struct {
-	IgnoreNullsInLists bool // strip nulls from lists before checking equality. Destination setups that don't support nulls in lists should set this to true.
+	// IgnoreNullsInLists allows stripping null values from lists before comparison.
+	// Destination setups that don't support nulls in lists should set this to true.
+	IgnoreNullsInLists bool
+
+	// AllowNull is a custom func to determine whether a data type may be correctly represented as null.
+	// Destinations that have problems representing some data types should provide a custom implementation here.
+	// If this param is empty, the default is to allow all data types to be nullable.
+	AllowNull func(arrow.DataType) bool
```
hermanschaaf, line 111: Destinations don't really control the schema of the data they receive; what will happen if a destination that doesn't support nulls receives a nullable schema? How can we be sure if we're not testing it? I think we should still be testing that case, but we can be more flexible with the comparisons (like in `IgnoreNullsInLists` above). WDYT?

candiduslynx: Well, that's true, but it's also tru for the `skip maps` & other skip options for test tables.

IMO, we should instead base what types are available by the available scalar implementations.

As for this one, it's just a work-around for tests.
I think another option could be allowing treating empty values as null ones for specific types, but then we won't be able to use Arrow record approx equal and would instead need to traverse the record ourselves.

plugins/destination/plugin_testing.go
candiduslynx: The `slice.Contains` requires comparable types for the slice, and the interfaces aren't.

I don't mind extracting `AllowNullFunc` as type, it that's a strong opinion.

plugins/destination/plugin_testing.go
hermanschaaf: > I think another option could be allowing treating empty values as null ones for specific types, but then we won't be able to use Arrow record approx equal and would instead need to traverse the record ourselves.

Yes, I think this would be a better solution. Like in the `IgnoreNullsInLists` case, we still want to test the plugins with these constraints, but we need to adapt the test expectations for them. And like in `IgnoreNullsInLists`, I think we should still be able to use record approx equal if we apply a transformation to the relevant types first (e.g. converting empty values to null, or whatever the case needs to be).

disq (approved):

hermanschaaf (approved):

PR #914, opened 2023-05-29
feat: More scalars

disq:
- [x] basic types
- [x] date/time/duration types
- [x] interval types
- [x] struct types
- [x] decimal types
- [ ] ~map types~


scalar/date32.go
```diff
@@ -0,0 +1,117 @@
...
+	"fmt"
+	"math"
+	"time"
+
+	"github.com/apache/arrow/go/v13/arrow"
+)
+
+type Date32 struct {
+	Valid bool
+	Value int32 // days since unix epoch
```
candiduslynx, line 14: use `arrow.Date32`?

scalar/date32.go
```diff
@@ -0,0 +1,117 @@
...
+	r, ok := rhs.(*Date32)
+	if !ok {
+		return false
+	}
+	return s.Valid == r.Valid && s.Value == r.Value
+}
+
+func (s *Date32) String() string {
+	if !s.Valid {
+		return "(null)"
```
candiduslynx, line 38: use `array.NullValueStr` as it's now used as const to be returned in `arrays` Arrow package?

scalar/date32.go
```diff
@@ -0,0 +1,117 @@
...
+	case uint64:
+		if value > math.MaxInt32 {
+			return &ValidationError{Type: s.DataType(), Msg: "uint64 bigger than MaxInt32", Value: value}
+		}
+		s.Value = int32(value)
+	case time.Time:
+		return s.Set(arrow.Date32FromTime(value))
+	case *time.Time:
+		if value == nil {
+			return nil
```
candiduslynx, line 79: shouldn't this also clear the already set value?

candiduslynx: `s.Valid = false`

scalar/date32.go
```diff
@@ -0,0 +1,117 @@
...
+		}
+
+		p, err := time.Parse("2006-01-02", value)
+		if err != nil {
+			return &ValidationError{Type: s.DataType(), Msg: "cannot parse date", Value: value, Err: err}
+		}
+		return s.Set(p)
+	case *string:
+		if value == nil {
+			return nil
```
candiduslynx, line 95: shouldn't `s.Valid = false` also be present?

scalar/date64.go
```diff
@@ -0,0 +1,114 @@
...
+	"fmt"
+	"math"
+	"time"
+
+	"github.com/apache/arrow/go/v13/arrow"
+)
+
+type Date64 struct {
+	Valid bool
+	Value int64 // int64 milliseconds since the UNIX epoch
```
candiduslynx, line 14: use `arrow.Date64`

scalar/date64.go
```diff
@@ -0,0 +1,114 @@
...
+	case uint64:
+		if value > math.MaxInt64 {
+			return &ValidationError{Type: s.DataType(), Msg: "uint64 bigger than MaxInt64", Value: value}
+		}
+		s.Value = int64(value)
+	case time.Time:
+		return s.Set(arrow.Date64FromTime(value))
+	case *time.Time:
+		if value == nil {
+			return nil
```
candiduslynx, line 76: `s.Valid = false`?

scalar/date64.go
```diff
@@ -0,0 +1,114 @@
...
+		}
+
+		p, err := time.Parse("2006-01-02", value)
+		if err != nil {
+			return &ValidationError{Type: s.DataType(), Msg: "cannot parse date", Value: value, Err: err}
+		}
+		return s.Set(p)
+	case *string:
+		if value == nil {
+			return nil
```
candiduslynx, line 92: `s.Valid = false`?

scalar/duration.go
```diff
@@ -0,0 +1,64 @@
...
+	case arrow.Microsecond:
+		return arrow.FixedWidthTypes.Duration_us
+	default:
+		panic("unknown duration unit")
+	}
+}
+
+func (s *Duration) String() string {
+	if !s.Int.IsValid() {
+		return "(null)"
```
candiduslynx, line 31:
```suggestion
		return array.NullValueStr
```

scalar/date32.go
candiduslynx:
```suggestion
		return array.NullValueStr
```

scalar/date64.go
```diff
@@ -0,0 +1,114 @@
...
+	r, ok := rhs.(*Date64)
+	if !ok {
+		return false
+	}
+	return s.Valid == r.Valid && s.Value == r.Value
+}
+
+func (s *Date64) String() string {
+	if !s.Valid {
+		return "(null)"
```
candiduslynx, line 38:
```suggestion
		return array.NullValueStr
```

scalar/duration.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+	r, ok := rhs.(*Duration)
+	if !ok {
+		return false
+	}
+	return s.Valid == r.Valid && s.Unit == r.Unit && s.Value == r.Value
+}
+
+func (s *Duration) Set(value any) error {
+	if dur, ok := value.(arrow.Duration); ok {
```
candiduslynx, line 49: what about `time.Duration`?

scalar/float.go
```diff
@@ -7,160 +7,52 @@ import (
...
 	if rhs == nil {
 		return false
 	}
-	r, ok := rhs.(*Float64)
+	r, ok := rhs.(*Float)
 	if !ok {
 		return false
 	}
-	return s.Valid == r.Valid && s.Value == r.Value
+	return s.getBitWidth() == r.getBitWidth() && s.Valid == r.Valid && s.Value == r.Value
```
candiduslynx, line 45: are we sure we want the comparison to depend on the bit width?
what about float16(0) vs float32(0) vs float64(0)?

scalar/float.go
```diff
@@ -7,160 +7,52 @@ import (
...
 }
 
-func (s *Float64) String() string {
+func (s *Float) String() string {
 	if !s.Valid {
 		return "(null)"
```
candiduslynx, line 50:
```suggestion
		return array.NullValueStr
```

scalar/int.go
```diff
@@ -7,42 +7,54 @@ import (
...
+func (s *Int) DataType() arrow.DataType {
+	switch s.getBitWidth() {
+	case 64:
+		return arrow.PrimitiveTypes.Int64
+	case 8:
+		return arrow.PrimitiveTypes.Int8
+	case 16:
+		return arrow.PrimitiveTypes.Int16
+	case 32:
+		return arrow.PrimitiveTypes.Int32
```
candiduslynx, line 29:
```suggestion
	case 64:
		return arrow.PrimitiveTypes.Int64
	case 32:
		return arrow.PrimitiveTypes.Int32
	case 16:
		return arrow.PrimitiveTypes.Int16
	case 8:
		return arrow.PrimitiveTypes.Int8
```

scalar/int.go
```diff
@@ -7,42 +7,54 @@ import (
...
+	default:
+		panic("invalid bit width")
+	}
 }
 
-func (s *Int64) String() string {
+func (s *Int) String() string {
 	if !s.Valid {
 		return "(null)"
```
candiduslynx, line 37:
```suggestion
		return array.NullValueStr
```

scalar/int.go
```diff
@@ -162,8 +205,33 @@ func (s *Int64) Set(val any) error {
...
-		return &ValidationError{Type: arrow.PrimitiveTypes.Int64, Msg: noConversion, Value: value}
+		return &ValidationError{Type: s.DataType(), Msg: noConversion, Value: value}
 	}
 	s.Valid = true
 	return nil
 }
+
+func (s *Int) validateValue(value int64) error {
+	switch {
+	case arrow.TypeEqual(s.DataType(), arrow.PrimitiveTypes.Int8):
```
candiduslynx, line 216: please do switch over the bit width instead of data type

scalar/int.go
```diff
@@ -162,8 +205,33 @@ func (s *Int64) Set(val any) error {
...
+		}
+	case arrow.TypeEqual(s.DataType(), arrow.PrimitiveTypes.Int16):
+		if value > math.MaxInt16 {
+			return &ValidationError{Type: s.DataType(), Msg: "value bigger than MaxInt16", Value: value}
+		}
+	case arrow.TypeEqual(s.DataType(), arrow.PrimitiveTypes.Int32):
+		if value > math.MaxInt32 {
+			return &ValidationError{Type: s.DataType(), Msg: "value bigger than MaxInt32", Value: value}
+		}
+	}
```
candiduslynx, line 228:
```suggestion
	switch s.getBitWidth() {
	case 32:
		if value > math.MaxInt32 {
			return &ValidationError{Type: s.DataType(), Msg: "value bigger than MaxInt32", Value: value}
		}
	case 16:
		if value > math.MaxInt16 {
			return &ValidationError{Type: s.DataType(), Msg: "value bigger than MaxInt16", Value: value}
		}
	case 8:
		if value > math.MaxInt8 {
			return &ValidationError{Type: s.DataType(), Msg: "value bigger than MaxInt8", Value: value}
		}
	}
```

scalar/interval.go
```diff
@@ -0,0 +1,252 @@
...
+	return arrow.FixedWidthTypes.DayTimeInterval
+}
+
+func (s *DayTimeInterval) IsValid() bool {
+	return s.Valid
+}
+
+func (s *DayTimeInterval) String() string {
+	if !s.Valid {
+		return "(null)"
```
candiduslynx, line 85:
```suggestion
		return array.NullValueStr
```

scalar/interval.go
```diff
@@ -0,0 +1,252 @@
...
+		var dti arrow.DayTimeInterval
+		if err := json.Unmarshal(v, &dti); err != nil {
+			return err
+		}
+		s.Valid = true
+		s.Value = dti
+		return nil
+	case *string:
+		if v == nil {
+			return s.Set(nil)
```
candiduslynx, line 143: `s.Valid = false`?

scalar/interval.go
```diff
@@ -0,0 +1,252 @@
...
+	return arrow.FixedWidthTypes.MonthDayNanoInterval
+}
+
+func (s *MonthDayNanoInterval) IsValid() bool {
+	return s.Valid
+}
+
+func (s *MonthDayNanoInterval) String() string {
+	if !s.Valid {
+		return "(null)"
```
candiduslynx, line 177:
```suggestion
		return array.NullValueStr
```

scalar/interval.go
```diff
@@ -0,0 +1,252 @@
...
+		var dti arrow.MonthDayNanoInterval
+		if err := json.Unmarshal(v, &dti); err != nil {
+			return err
+		}
+		s.Valid = true
+		s.Value = dti
+		return nil
+	case *string:
+		if v == nil {
+			return s.Set(nil)
```
candiduslynx, line 235: `s.Valid = false`?

scalar/interval.go
```diff
@@ -0,0 +1,252 @@
...
+		var mi monthIntervalData
+		if err := json.Unmarshal(v, &mi); err != nil {
+			return err
+		}
+		s.Valid = true
+		s.Value = int64(mi.Months)
+		return nil
+	case *string:
+		if v == nil {
+			return s.Int.Set(nil)
```
candiduslynx, line 59: `s.Valid = false`?

scalar/int_test.go
```diff
@@ -1,33 +1,35 @@
...
-		{source: uint8(1), result: Int64{Value: 1, Valid: true}},
-		{source: uint16(1), result: Int64{Value: 1, Valid: true}},
-		{source: uint32(1), result: Int64{Value: 1, Valid: true}},
-		{source: uint64(1), result: Int64{Value: 1, Valid: true}},
-		{source: float32(1), result: Int64{Value: 1, Valid: true}},
-		{source: float64(1), result: Int64{Value: 1, Valid: true}},
-		{source: "1", result: Int64{Value: 1, Valid: true}},
-		{source: _int8(1), result: Int64{Value: 1, Valid: true}},
-		{source: &Int64{Value: 1, Valid: true}, result: Int64{Value: 1, Valid: true}},
+		{source: int8(1), result: Int{Value: 1, Valid: true}},
```
yevgenypats, line 12: What about tests for `Int64`, `Int32`, `Int16`, `Int8` (Now that they are supported)?

scalar/scalar.go
```diff
@@ -46,19 +47,46 @@ func (v Vector) Equal(r Vector) bool {
...
+		switch {
+		case arrow.TypeEqual(dt, arrow.FixedWidthTypes.Timestamp_ms):
+			return &Timestamp{Unit: arrow.Millisecond}
+		case arrow.TypeEqual(dt, arrow.FixedWidthTypes.Timestamp_ns):
+			return &Timestamp{Unit: arrow.Nanosecond}
+		case arrow.TypeEqual(dt, arrow.FixedWidthTypes.Timestamp_s):
+			return &Timestamp{Unit: arrow.Second}
+		default:
+			return &Timestamp{Unit: arrow.Microsecond}
+		}
```
candiduslynx, line 59:
```suggestion
		return &Timestamp{Unit: dt.(*arrow.TimestampType).Unit}
```

scalar/scalar.go
```diff
@@ -46,19 +47,46 @@ func (v Vector) Equal(r Vector) bool {
...
 		return &Binary{}
 	case arrow.STRING:
 		return &String{}
+	case arrow.LARGE_BINARY:
+		return &LargeBinary{}
+	case arrow.LARGE_STRING:
+		return &LargeString{}
 	case arrow.INT64:
-		return &Int64{}
+		return &Int{}
```
candiduslynx, line 69:
```suggestion
		return &Int{BitWidth: 64}
```
for explicitness

scalar/scalar.go
```diff
@@ -46,19 +47,46 @@ func (v Vector) Equal(r Vector) bool {
...
+		return &Int{BitWidth: 32}
+	case arrow.INT16:
+		return &Int{BitWidth: 16}
+	case arrow.INT8:
+		return &Int{BitWidth: 8}
 	case arrow.UINT64:
-		return &Uint64{}
-	case arrow.FLOAT32:
-		return &Float32{}
+		return &Uint{}
```
candiduslynx, line 77:
```suggestion
		return &Uint{BitWidth: 64}
```

scalar/scalar.go
```diff
@@ -46,19 +47,46 @@ func (v Vector) Equal(r Vector) bool {
...
+	case arrow.UINT32:
+		return &Uint{BitWidth: 32}
+	case arrow.UINT16:
+		return &Uint{BitWidth: 16}
+	case arrow.UINT8:
+		return &Uint{BitWidth: 8}
 	case arrow.FLOAT64:
-		return &Float64{}
+		return &Float{}
```
candiduslynx, line 85:
```suggestion
		return &Float{BitWidth: 64}
```

yevgenypats (approved):

scalar/scalar.go
```diff
@@ -78,6 +106,71 @@ func NewScalar(dt arrow.DataType) Scalar {
...
+			return &Duration{
+				Int:  Int{BitWidth: 64},
+				Unit: arrow.Second,
+			}
+		default:
+			return &Duration{
+				Int:  Int{BitWidth: 64},
+				Unit: arrow.Microsecond,
+			}
+		}
```
candiduslynx, line 135:
```suggestion
		return &Duration{Int:  Int{BitWidth: 64}, Unit: dt.(*arrow.DurationType).Unit}
```

scalar/scalar.go
```diff
@@ -78,6 +106,71 @@ func NewScalar(dt arrow.DataType) Scalar {
...
+	case arrow.TIME32, arrow.TIME64:
+		switch {
```
candiduslynx, line 137: split into 2 different cases & work with the cast dt

candiduslynx (changes requested):
1st round (reviewed till scalar.go)

scalar/float.go
disq: They are separate types in the arrow world so I'm guessing that's OK :)

scalar/interval.go
disq: No need since s.Int.Set(nil) is called here

scalar/duration.go
disq: added

scalar/int_test.go
disq: Added

scalar/duration.go
```diff
@@ -0,0 +1,73 @@
...
+		return arrow.FixedWidthTypes.Duration_s
+	case arrow.Millisecond:
+		return arrow.FixedWidthTypes.Duration_ms
+	case arrow.Nanosecond:
+		return arrow.FixedWidthTypes.Duration_ns
+	case arrow.Microsecond:
+		return arrow.FixedWidthTypes.Duration_us
+	default:
+		panic("unknown duration unit")
+	}
```
candiduslynx, line 27:
```suggestion
	return &arrow.DurationType{Unit: s.Unit}
```

scalar/decimal.go
```diff
@@ -0,0 +1,145 @@
...
+		return false
+	}
+	return s.Valid == r.Valid && s.Value == r.Value
+}
+
+func (s *Decimal256) String() string {
+	if !s.Valid {
+		return nullValueStr
+	}
+	return s.DataType().String() + " value"
```
candiduslynx, line 37: do we need ` + " value"` & why?

scalar/decimal.go
```diff
@@ -0,0 +1,145 @@
...
+		return false
+	}
+	return s.Valid == r.Valid && s.Value == r.Value
+}
+
+func (s *Decimal128) String() string {
+	if !s.Valid {
+		return nullValueStr
+	}
+	return s.DataType().String() + " value"
```
candiduslynx, line 110: do we need ` + " value"` & why?

scalar/decimal.go
```diff
@@ -0,0 +1,145 @@
...
+	Valid bool
+	Value decimal256.Num
+}
+
+func (s *Decimal256) IsValid() bool {
+	return s.Valid
+}
+
+func (*Decimal256) DataType() arrow.DataType {
+	return &arrow.Decimal256Type{}
```
candiduslynx, line 19: nope, need to properly set scale & precision

scalar/decimal.go
```diff
@@ -0,0 +1,145 @@
...
+	Valid bool
+	Value decimal128.Num
+}
+
+func (s *Decimal128) IsValid() bool {
+	return s.Valid
+}
+
+func (*Decimal128) DataType() arrow.DataType {
+	return &arrow.Decimal128Type{}
```
candiduslynx, line 92: nope, need to properly set scale & precision

scalar/float.go
```diff
@@ -58,175 +68,129 @@ func (s *Float32) Set(val any) error {
...
-			return &ValidationError{Type: arrow.PrimitiveTypes.Int8, Msg: "uint32 bigger than MaxInt8", Value: value}
-		}
-		s.Value = float32(value)
+		return s.Set(float64(value))
 	case uint64:
-		if value > math.MaxInt32 {
-			return &ValidationError{Type: arrow.PrimitiveTypes.Int8, Msg: "uint64 bigger than MaxInt8", Value: value}
-		}
-		s.Value = float32(value)
+		return s.Set(float64(value))
```
candiduslynx, line 85: precision loss, BTW

scalar/scalar.go
```diff
@@ -95,27 +157,82 @@ func AppendToBuilder(bldr array.Builder, s Scalar) {
...
+
+	case arrow.STRUCT:
+		sb := bldr.(*array.StructBuilder)
+		sb.Append(true)
+
+		v := s.(*Struct).Value
+		m := v.(map[string]any)
+		st := sb.Type().(*arrow.StructType)
+
+		for i, f := range st.Fields() {
```
candiduslynx, line 219: there are several issues with this code:
1. extra fields in map don't trigger error (they should)
2. missing fields in maps: IMO they should result in append null/ append empty value (now they result in empty scalars appended)

scalar/string.go
```diff
@@ -78,12 +79,37 @@ func (s *String) Set(val any) error {
 	return nil
 }
 
-type _smallString String
-
 type LargeString struct {
-	_smallString
+	s String
```
candiduslynx, line 83:
```suggestion
	String
```
would also work & wouldn't require from you to redefine all of the functions

scalar/struct.go
```diff
@@ -0,0 +1,74 @@
...
+		return false
+	}
+	return s.Valid == r.Valid && reflect.DeepEqual(s.Value, r.Value)
+}
+
+func (s *Struct) String() string {
+	if !s.Valid {
+		return nullValueStr
+	}
+	return s.Type.String() + " value"
```
candiduslynx, line 36: do we need `+ " value"` & why?

scalar/struct.go
```diff
@@ -0,0 +1,74 @@
...
+		var x map[string]any
+		if err := json.Unmarshal([]byte(str), &x); err != nil {
+			return err
+		}
+		s.Value = x
+		s.Valid = true
+		return nil
+	}
+
+	s.Value = val
```
candiduslynx, line 67: kindly check what will happen with `Struct.Set((*bool)(nil))`

scalar/time.go
```diff
@@ -0,0 +1,71 @@
...
+		return arrow.FixedWidthTypes.Time32s
+	case s.Unit == arrow.Millisecond && s.Int.BitWidth == 32:
+		return arrow.FixedWidthTypes.Time32ms
+	case s.Unit == arrow.Nanosecond && s.Int.BitWidth == 64:
+		return arrow.FixedWidthTypes.Time64ns
+	case s.Unit == arrow.Microsecond && s.Int.BitWidth == 64:
+		return arrow.FixedWidthTypes.Time64us
+	default:
+		panic("unknown time unit")
+	}
```
candiduslynx, line 24: `s.Int.BitWidth` may be `0`, need to use wrapper:
```suggestion
	switch width := s.getBitWidth(); width {
	case 64:
		return &arrow.Time64Type{Unit: s.Unit}
	case 32:
		return &arrow.Time32Type{Unit: s.Unit}
	default:
		panic("unsupported bit width " + strconv.Itoa(width))
	}
```

scalar/time.go
```diff
@@ -0,0 +1,71 @@
...
+
+func (s *Time) Set(value any) error {
+	switch v := value.(type) {
+	case arrow.Time32:
+		return s.Int.Set(int64(v))
+	case arrow.Time64:
+		return s.Int.Set(int64(v))
+
+	case string:
+		switch s.BitWidth {
```
candiduslynx, line 46:
```suggestion
		switch s.getBitWidth() {
```

scalar/time.go
```diff
@@ -0,0 +1,71 @@
+package scalar
+
+import (
+	"github.com/apache/arrow/go/v13/arrow"
+)
+
+type Time struct {
```
candiduslynx, line 7: this also needs `Value` func, something like
```go
func (s *Time) Value() any {
	if !s.Valid {
		return nil
	}
	switch width := s.getBitWidth(); width {
	case 64:
		return arrow.Time64(s.Int.Value())
	case 32:
		return arrow.Time32(s.Int.Value())
	default:
		panic("unsupported bit width " + strconv.Itoa(width))
	}
}
```

scalar/timestamp.go
```diff
@@ -29,14 +29,26 @@ const (
...
+		return arrow.FixedWidthTypes.Timestamp_s
+	case arrow.Millisecond:
+		return arrow.FixedWidthTypes.Timestamp_ms
+	case arrow.Nanosecond:
+		return arrow.FixedWidthTypes.Timestamp_ns
+	case arrow.Microsecond:
+		return arrow.FixedWidthTypes.Timestamp_us
+	default:
+		panic("unknown timestamp unit")
+	}
```
candiduslynx, line 51:
```suggestion
	return &arrow.TimestampType{Unit: s.Unit}
```

scalar/uint.go
```diff
@@ -1,47 +1,60 @@
...
+func (s *Uint) DataType() arrow.DataType {
+	switch s.getBitWidth() {
+	case 64:
+		return arrow.PrimitiveTypes.Uint64
+	case 8:
+		return arrow.PrimitiveTypes.Uint8
+	case 16:
+		return arrow.PrimitiveTypes.Uint16
+	case 32:
+		return arrow.PrimitiveTypes.Uint32
```
candiduslynx, line 29:
```suggestion
	case 64:
		return arrow.PrimitiveTypes.Uint64
	case 32:
		return arrow.PrimitiveTypes.Uint32
	case 16:
		return arrow.PrimitiveTypes.Uint16
	case 8:
		return arrow.PrimitiveTypes.Uint8
```

scalar/uint.go
```diff
@@ -58,121 +71,167 @@ func (s *Uint64) Set(val any) error {
...
 	case uint:
-		s.Value = uint64(value)
+		return s.Set(uint64(value))
 	case float32:
 		if value < 0 {
-			return &ValidationError{Type: arrow.PrimitiveTypes.Uint64, Msg: "float32 less than 0", Value: value}
+			return &ValidationError{Type: s.DataType(), Msg: "float32 less than 0", Value: value}
 		}
-		s.Value = uint64(value)
+		return s.Set(uint64(value))
```
candiduslynx, line 114: check for `math.MaxUint64`?

scalar/uint.go
```diff
@@ -58,121 +71,167 @@ func (s *Uint64) Set(val any) error {
...
 	case float64:
 		if value < 0 {
-			return &ValidationError{Type: arrow.PrimitiveTypes.Uint64, Msg: "float64 less than 0", Value: value}
+			return &ValidationError{Type: s.DataType(), Msg: "float64 less than 0", Value: value}
 		}
-		s.Value = uint64(value)
+		return s.Set(uint64(value))
```
candiduslynx, line 119: check for `math.MaxUint64`?

candiduslynx (changes requested):

scalar/decimal.go
```diff
@@ -9,14 +9,15 @@ import (
 type Decimal256 struct {
 	Valid bool
 	Value decimal256.Num
+	Type  arrow.DataType // Stores precision and scale
```
candiduslynx, line 12: Please just store properly typed dt:
```suggestion
	Type  *arrow.Decimal256Type
```

scalar/string.go
disq: Doesn't work because we have `.String()` from fmt.Stringer

scalar/decimal.go
```diff
@@ -137,9 +175,45 @@ func (s *Decimal128) Set(val any) error {
...
 	s.Valid = true
 	return nil
 }
+
+func (s *Decimal128) Precision() int32 {
+	return s.Type.(arrow.DecimalType).GetPrecision()
+}
+
+func (s *Decimal128) Scale() int32 {
+	return s.Type.(arrow.DecimalType).GetScale()
```
candiduslynx, line 218:
```suggestion
	return s.Type.Scale
```

scalar/map.go
```diff
@@ -0,0 +1,74 @@
...
+	"reflect"
+
+	"github.com/apache/arrow/go/v13/arrow"
+)
+
+type Map struct {
+	Valid bool
+	Value any
+
+	Type arrow.DataType
```
candiduslynx, line 14:
```suggestion
	Type *arrow.MapType
```

scalar/struct.go
disq: `String()` is supposed to stringify the value (for debugging? I don't know why.. since everything else is supposed to use `Get() any`) the +value makes it apparent that it's a (fake) value and not the type we're printing.

scalar/map.go
```diff
@@ -0,0 +1,74 @@
...
+	if sc, ok := val.(Scalar); ok {
+		if !sc.IsValid() {
+			s.Valid = false
+			return nil
+		}
+		return s.Set(sc.Get())
+	}
+
+	if str, ok := val.(string); ok {
+		var x map[string]any
```
candiduslynx, line 58: In Apache Arrow maps aren't simply maps.
Instead, they are lists, with each entry being a struct of 2 fields:
1. Key (field name: `key`)
2. Item (field name: `value`)

scalar/scalar.go
```diff
@@ -228,6 +219,19 @@ func AppendToBuilder(bldr array.Builder, s Scalar) {
 			AppendToBuilder(sb.FieldBuilder(i), sc)
 		}
 
+	case arrow.MAP:
+		mb := bldr.(*array.MapBuilder)
+		mb.Append(true)
+
+		v := s.(*Map).Value
+		_ = v
```
candiduslynx, line 227: Maps are a bit tricky to append.
See this code: https://github.com/cloudquery/cloudquery/blob/main/plugins/destination/clickhouse/typeconv/ch/values/map_test.go

candiduslynx (changes requested):

scalar/timestamp.go
disq: This loses the timezone info and makes things more complicated (arrow.TypeEqual doesn't work)

scalar/time.go
candiduslynx: We need to make sure that we set proper values, and the code in Apache Arrow already fails on the bad unit/type combination

scalar/time.go
disq: added Get()

scalar/uint.go
disq: maxuint64 is bigger than maxfloat32 (or maxfloat64) so no need to check here. uint64 branch calls `validateValue` to check against target bitwidth.

scalar/decimal.go
disq: https://github.com/cloudquery/plugin-sdk/pull/914#discussion_r1210098302

scalar/struct.go
disq: added typed nil check

scalar/struct.go
candiduslynx: IMO the scalar.String just returns the value that we could use in the future for some purposes (e.g., append to `*array.StringBuilder`).
As other scalars show a proper value, I think these should show it, too.
cc: @yevgenypats @hermanschaaf

scalar/struct.go
```diff
@@ -0,0 +1,94 @@
...
+
+	if sc, ok := val.(Scalar); ok {
+		if !sc.IsValid() {
+			s.Valid = false
+			return nil
+		}
+		return s.Set(sc.Get())
+	}
+
+	if rv := reflect.ValueOf(val); rv.Kind() == reflect.Pointer && !rv.Elem().IsValid() { // typed nil
```
candiduslynx, line 57: move this below type switch, as reflect is costly and we should use it as the last resort

candiduslynx (changes requested):

scalar/uint.go
candiduslynx: https://go.dev/play/p/5au3XVMfX4C

scalar/float.go
```diff
@@ -56,177 +66,154 @@ func (s *Float32) Set(val any) error {
...
-		return nil
-	}
-
-	if sc, ok := val.(Scalar); ok {
-		if !sc.IsValid() {
-			s.Valid = false
-			return nil
+	switch s.getBitWidth() {
+	case 16:
+		if value > maxFloat16 {
```
candiduslynx, line 203: negative ones, too?

candiduslynx (changes requested):

scalar/map.go
```diff
@@ -0,0 +1,63 @@
+package scalar
+
+import (
+	"reflect"
+
+	"github.com/apache/arrow/go/v13/arrow"
+)
+
+type Map struct {
```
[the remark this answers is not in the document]
disq: it's not used just dead code, will remove for now

scalar/float.go
```diff
@@ -56,177 +66,154 @@ func (s *Float32) Set(val any) error {
...
-			return &ValidationError{Type: arrow.PrimitiveTypes.Float32, Msg: "int64 bigger than MaxInt8", Value: value}
+		switch {
+		case s.getBitWidth() == 64 && value > maxSafeValue64:
+			return &ValidationError{Type: s.DataType(), Msg: "int64 greater than maximum safe value of 2^53", Value: value}
+		case s.getBitWidth() == 64 && value < minSafeValue64:
+			return &ValidationError{Type: s.DataType(), Msg: "int64 less than minimum safe value of -2^53", Value: value}
+		case s.getBitWidth() == 32 && value > maxSafeValue32:
+			return &ValidationError{Type: s.DataType(), Msg: "int64 greater than maximum safe value of 2^24", Value: value}
+		case s.getBitWidth() == 32 && value < minSafeValue32:
+			return &ValidationError{Type: s.DataType(), Msg: "int64 less than minimum safe value of -2^24", Value: value}
```
candiduslynx, line 92:
```suggestion
		switch s.getBitWidth() {
		case 64:
			if value > maxSafeValue64 {
				return &ValidationError{Type: s.DataType(), Msg: "int64 greater than maximum safe value of 2^53", Value: value}
			}
			if value < minSafeValue64 {
				return &ValidationError{Type: s.DataType(), Msg: "int64 less than minimum safe value of -2^53", Value: value}
			}
		case 32:
			if value > maxSafeValue32 {
				return &ValidationError{Type: s.DataType(), Msg: "int64 greater than maximum safe value of 2^24", Value: value}
			}
			if value < minSafeValue32 {
				return &ValidationError{Type: s.DataType(), Msg: "int64 less than minimum safe value of -2^24", Value: value}
			}
```

scalar/float.go
```diff
@@ -56,177 +66,154 @@ func (s *Float32) Set(val any) error {
...
-}
-
-func (s *Float64) String() string {
-	if !s.Valid {
-		return "(null)"
+func (s *Float) validateValue(value float64) error {
+	const maxFloat16 = 65504.0
+
+	switch {
+	case s.getBitWidth() == 16 && value > maxFloat16:
```
candiduslynx, line 208: transform to switch by `s.getBitWidth()` and move `if` statements to cases

candiduslynx (changes requested):

scalar/scalar.go
disq: 1. added
2. they result in null scalars being added as a new scalar is by default "null" (isvalid is false)

scalar/struct.go
disq: added json.Marshal instead of this

scalar/scalar.go
```diff
@@ -95,27 +148,83 @@ func AppendToBuilder(bldr array.Builder, s Scalar) {
...
+		st := sb.Type().(*arrow.StructType)
+
+		for i, f := range st.Fields() {
+			sc := NewScalar(sb.FieldBuilder(i).Type())
+
+			if sv, ok := m[f.Name]; ok {
+				if err := sc.Set(sv); err != nil {
+					panic(err)
+				}
+				delete(m, f.Name)
```
candiduslynx, line 214: This will mutate the passed map value.
```go
names := make(map[string]struct{}, len(m))
for k := range m {
	names[k] = struct{}{}
}

...
if sv, ok := m[f.Name]; ok {
	delete(names, f.Name)
}
...
if len(names) > 0 {
	panic(fmt.Errorf("struct has extra fields: %+v", m))
}
```

scalar/time.go
```diff
@@ -0,0 +1,81 @@
...
```
candiduslynx, line 7: IMO it requires a separate `String` method:
```go
func (s *Time) String() string {
	if !s.Valid {
		return nullValueStr
	}

	switch s.getBitWidth() {
	case 64:
		return arrow.Time64(s.Int.Value).FormattedString(s.Unit)
	case 32:
		return arrow.Time32(s.Int.Value).FormattedString(s.Unit)
	default:
		panic("unsupported bit width")
	}
}

File: scalar/interval.go [hunk 58]
```diff
@@ -0,0 +1,254 @@
+package scalar
+
+import (
+	"encoding/json"
+
+	"github.com/apache/arrow/go/v13/arrow"
+)
+
+type MonthInterval struct {
+	Int
```
candiduslynx, line 10: IMO it requires a separate `String` method

scalar/decimal.go
```diff
@@ -0,0 +1,203 @@
...
+		return false
+	}
+	return s.Valid == r.Valid && s.Value == r.Value
+}
+
+func (s *Decimal256) String() string {
+	if !s.Valid {
+		return nullValueStr
+	}
+	return s.DataType().String() + " value"
```
candiduslynx, line 38:
```suggestion
	return s.Value.ToString(s.Type.Scale)
```

scalar/decimal.go
```diff
@@ -0,0 +1,203 @@
...
+		return false
+	}
+	return s.Valid == r.Valid && s.Value == r.Value
+}
+
+func (s *Decimal128) String() string {
+	if !s.Valid {
+		return nullValueStr
+	}
+	return s.DataType().String() + " value"
```
candiduslynx, line 140:
```suggestion
	return s.Value.ToString(s.Type.Scale)
```

candiduslynx (changes requested):
Last bits about stringers

scalar/bool.go
```diff
@@ -67,6 +67,10 @@ func (s *Bool) Set(val any) error {
 	case *bool:
 		return s.Set(*value)
 	case *string:
+		if value == nil {
```
erezrokah, line 70: Don't we need this check in the `*bool` case too?

candiduslynx: +100

scalar/time.go
```diff
@@ -0,0 +1,81 @@
...
+	Int
```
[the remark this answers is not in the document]
disq: @hermanschaaf yes we're using its valid/value logic.

scalar/interval.go
disq: Done, since it's passed in from Arrow side as `monthIntervalData` struct I also used it in the stringer.

scalar/struct_test.go
```diff
@@ -0,0 +1,63 @@
...
+				if panicked && !tc.expectPanic {
+					t.Errorf("unexpected panic")
+				}
+				if !panicked && tc.expectPanic {
+					t.Errorf("expected panic")
+				}
+			}()
+
+			defer func() {
+				if r := recover(); r != nil {
```
candiduslynx, line 41: there's [`require.Panics`](https://pkg.go.dev/github.com/stretchr/testify/require#Panics) func family for such cases, FYI (with err can check err string, notpanics requires code not to panic, etc)

candiduslynx (approved):
nice! 🚀

PR #915, opened 2023-05-30
feat!: SDK V4 (One plugin and protocol for sources and destinations)

yevgenypats:
Playing with the idea of having only one plugin and protocol.


writers/batch.go
```diff
@@ -69,71 +82,95 @@ func WithBatchSizeBytes(size int) Option {
...
+					w.flush(ctx, tableName, upsertBatch, resources)
 				}
 				return
 			}
-			if uint64(len(resources)) == 1000 || sizeBytes+util.TotalRecordSize(r) > int64(1000) {
-				w.flush(ctx, table, resources)
-				resources = make([]arrow.Record, 0)
+			if upsertBatch != r.Upsert {
+				w.flush(ctx, tableName, upsertBatch, resources)
+				resources = make([]*plugin.MessageInsert, 0)
```
disq, line 135: Would consider `resources = resources[:0]` type syntax in here and below

plugin/testing_write_migrate.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 func tableUUIDSuffix() string {
-	return strings.ReplaceAll(uuid.NewString(), "-", "_")
+	return strings.ReplaceAll(uuid.NewString(), "-", "_")[:8] // use only first 8 chars
```
disq, line 17: You don't need the `ReplaceAll` if you're only using the first part of the UUID (which is 8 chars before the first `-`)

plugin/plugin.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+type UnimplementedDestination struct{}
+
+func (UnimplementedDestination) Write(ctx context.Context, options WriteOptions, res <-chan Message) error {
+	return ErrNotImplemented
+}
+
+func (UnimplementedDestination) Read(ctx context.Context, table *schema.Table, res chan<- arrow.Record) error {
+	return fmt.Errorf("not implemented")
```
disq, line 29: `return ErrNotImplemented` I think

plugin/messages.go
```diff
@@ -0,0 +1,87 @@
...
+	if err != nil {
+		panic(err)
+	}
+	return table
+}
+
+// MessageDeleteStale is a pretty specific message which requires the destination to be aware of a CLI use-case
+// thus it might be deprecated in the future
+// in favour of MessageDelete or MessageRawQuery
+// The message indeciates that the destination needs to run something like "DELETE FROM table WHERE _cq_source_name=$1 and sync_time < $2"
```
disq, line 38: `indicates`

internal/servers/discovery/v1/discovery.go
```diff
@@ -0,0 +1,16 @@
...
+	pb "github.com/cloudquery/plugin-pb-go/pb/discovery/v1"
+)
+
+type Server struct {
+	pb.UnimplementedDiscoveryServer
+	Versions []uint64
+}
+
+func (s *Server) GetVersions(context.Context, *pb.GetVersions_Request) (*pb.GetVersions_Response, error) {
+	return &pb.GetVersions_Response{Versions: s.Versions}, nil
```
disq, line 15: `[]uint64` vs `[]int32` causes isssues here:

> github.com/cloudquery/plugin-sdk/v4@v4.0.0-20230617094031-84a6a9b7e057/internal/servers/discovery/v1/discovery.go:15:44: cannot use s.Versions (variable of type []uint64) as []int32 value in struct literal

cover
```diff
@@ -0,0 +1,199 @@
+mode: set
```
hermanschaaf, line 1: This file should be removed I think

serve/plugin.go
```diff
@@ -1,67 +1,109 @@
...
-		fmt.Println(err)
-		os.Exit(1)
+	return s
+}
+
+func (s *PluginServe) bufPluginDialer(context.Context, string) (net.Conn, error) {
+	return s.testListenerConn.Dial()
+}
+
+func (s *PluginServe) Serve(ctx context.Context) error {
```
disq, line 94: The returned `error` here is not used (or is a hassle to use) from a `main.go` POV. Maybe make a `ServeConsole()`  helper func that calls Serve with context.Background(), doesn't return error, but prints error to stderr (and exits with proper error code) instead? I know it's

schema/testdata.go
```diff
@@ -28,6 +28,7 @@ type TestSourceOptions struct {
 	SkipTimes      bool // time of day types
 	SkipLargeTypes bool // e.g. large binary, large string
 	TimePrecision  time.Duration
+	SkipDecimals   bool
```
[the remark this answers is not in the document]
erezrokah: Also, it's not shown on the diff for some reason (probably since the PR needs a rebase), but `SkipMaps` is not used either and we don't add the maps at all

erezrokah:
This branch is also missing a bunch of Scalar values, like `scalar.Date32`, `scalar.Time`, `scalar.Float` and `scalar.Int`, so we would need to sync that file with `main`. See https://github.com/cloudquery/plugin-sdk/tree/b7bcd9326eb87f43e77e509ec5151c7f4a6f01b4/scalar

internal/servers/destination/v1/destinations.go
```diff
@@ -58,13 +65,29 @@ func (s *Server) Migrate(ctx context.Context, req *pb.Migrate_Request) (*pb.Migr
...
 
-	return &pb.Migrate_Response{}, s.Plugin.Migrate(ctx, tables)
+	writeCh := make(chan message.Message)
+	eg, ctx := errgroup.WithContext(ctx)
+	eg.Go(func() error {
+		return s.Plugin.Write(ctx, plugin.WriteOptions{
+			MigrateForce: s.migrateMode == plugin.MigrateModeForce,
+		}, writeCh)
+	})
+	for _, table := range tables {
```
disq, line 75:
```suggestion
	for _, table := range tables.FlattenTables() {
```
Maybe?

internal/servers/destination/v1/destinations.go
```diff
@@ -93,32 +116,40 @@ func (s *Server) Write(msg pb.Destination_WriteServer) error {
...
 	eg, ctx := errgroup.WithContext(msg.Context())
+
 	eg.Go(func() error {
-		return s.Plugin.Write(ctx, sourceSpec, tables, syncTime, resources)
+		return s.Plugin.Write(ctx, plugin.WriteOptions{
+			MigrateForce: s.spec.MigrateMode == specs.MigrateModeForced,
+		}, msgs)
 	})
 
+	for _, table := range tables {
```
disq, line 128:
```suggestion
	for _, table := range tables.FlattenTables() {
```
Not sure if tables are unflattened here (or in Migrate) but we seem to be calling MigrateTable a lot?

yevgenypats:
Closing in favour of https://github.com/cloudquery/plugin-sdk/pull/984

PR #917, opened 2023-05-30
feat: Add type support option

hermanschaaf:
(Depends on https://github.com/cloudquery/plugin-pb-go/pull/5)

This adds source plugin support for a new `type_support` config option with two possible values:
 - `limited`: (default): only types that were supported in cqtypes will be used. Any types that don't conform will be converted before sending it to the destination plugin. This is backwards-compatible
 - `full`: all available Arrow types will be used, including lists and structs.

When `limited` is specified, we transform unsupported types to their old equivalents before sending them over the wire. When `DynamicTables` is called, we also respect this option and return the schema matching the setting. 

When `Tables` is called however, the option is ignored (partly because we don't know what it's set to) and we use the `full` type. This means that the full type will also be used in the docs.

Many of the changes here are related to testing: 
 - a new test for `GetDynamicTables` to assert the new behavior works correctly
 - updated tests for `Sync` with more types, both to assert the new behavior and test old ones more thoroughly

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/cloudquery/plugin-sdk/v3
 go 1.19
 
 require (
-	github.com/apache/arrow/go/v13 v13.0.0-20230509040948-de6c3cd2b604
+	github.com/apache/arrow/go/v13 v13.0.0-20230525142029-2d32efeedad8
 	github.com/bradleyjkemp/cupaloy/v2 v2.8.0
 	github.com/cloudquery/plugin-pb-go v1.0.8
 	github.com/cloudquery/plugin-sdk/v2 v2.7.0
--- a/go.mod
+++ b/go.mod
@@ -26,7 +26,8 @@ require (
 	google.golang.org/protobuf v1.30.0
 )
 
-replace github.com/cloudquery/plugin-pb-go v1.0.8 => ../plugin-pb-go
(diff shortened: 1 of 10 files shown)


go.mod
```diff
@@ -26,6 +26,7 @@ require (
 	google.golang.org/protobuf v1.30.0
 )
 
+replace github.com/cloudquery/plugin-pb-go v1.0.8 => ../plugin-pb-go
```
hermanschaaf, line 29: We will need to update this once is https://github.com/cloudquery/plugin-pb-go/pull/5 is merged and released

plugins/source/plugin.go
```diff
@@ -309,24 +320,153 @@ func (p *Plugin) Sync(ctx context.Context, syncTime time.Time, res chan<- *schem
...
+			arrow.PrimitiveTypes.Int8,
+			arrow.PrimitiveTypes.Int16,
+			arrow.PrimitiveTypes.Int32,
+			arrow.PrimitiveTypes.Uint8,
+			arrow.PrimitiveTypes.Uint16,
+			arrow.PrimitiveTypes.Uint32,
+			arrow.PrimitiveTypes.Uint64):
+			newTable.Columns[c].Type = arrow.PrimitiveTypes.Int64
+		case typeOneOf(col.Type,
+			arrow.PrimitiveTypes.Float32):
```
disq, line 434:
```suggestion
			arrow.PrimitiveTypes.Float32, arrow.FixedWidthTypes.Float16):
```

PR #920, opened 2023-05-30
feat: Separate Queued Tables from InProgress

bbernays:
#### Summary

When users limit the concurrency of their sync, tables that have not been started will appear to be "in-progress" even though they are waiting to start...

This PR separates `queued` tables from `in-progress` by defining `in-progress` tables as those that have no `end-time`, but do have a non-zero start time. `queued` tables are those tables that have a zero start and end time

### DIFF
--- a/plugins/source/metrics.go
+++ b/plugins/source/metrics.go
@@ -172,8 +172,9 @@ func (s *Metrics) InProgressTables() []string {
 		for _, clientMetrics := range tableMetrics {
 			clientMetrics.mutex.Lock()
 			endTime := clientMetrics.endTime
+			starTime := clientMetrics.startTime
 			clientMetrics.mutex.Unlock()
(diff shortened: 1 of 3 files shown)


plugins/source/metrics.go
```diff
@@ -172,8 +172,9 @@ func (s *Metrics) InProgressTables() []string {
 		for _, clientMetrics := range tableMetrics {
 			clientMetrics.mutex.Lock()
 			endTime := clientMetrics.endTime
+			starTime := clientMetrics.startTime
```
disq, line 175: ⭐ ⏲️

disq (approved):

plugins/source/metrics.go
bbernays: fixed

yevgenypats (approved):

PR #922, opened 2023-05-31
fix(scalar): Handle nil pointer to []byte in uuid and binary

erezrokah:
#### Summary

Follow up to https://github.com/cloudquery/plugin-sdk/pull/922 to handle nil `*[]byte]` in `uuid` and `binary`.
Also added some tests to floats I can extract to another PR.

This is required for supporting `nil` values in https://github.com/cloudquery/cloudquery/pull/11115.
Discovered by generating test data with the [2 rows option](https://github.com/cloudquery/cloudquery/pull/11115#discussion_r1211284293) cc @candiduslynx.

---

- [x] Update or add tests 🧪

### DIFF
--- a/scalar/uuid.go
+++ b/scalar/uuid.go
@@ -63,8 +63,15 @@ func (s *UUID) Set(src any) error {
 		return s.Set(value2)
 	case [16]byte:
 		s.Value = uuid.UUID(value)
+	case *[]byte:
+		if value == nil {
+			s.Valid = false
+			return nil
+		}
+		return s.Set(*value)
 	case []byte:
 		if value == nil {
+			s.Valid = false
 			return nil
 		}
 		if len(value) != 16 {
@@ -79,6 +86,7 @@ func (s *UUID) Set(src any) error {
 		s.Value = uuidVal
 	case *string:
 		if value == nil {
+			s.Valid = false
 			return nil
 		}
 		return s.Set(*value)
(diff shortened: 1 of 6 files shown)


yevgenypats (approved):

candiduslynx (approved):

candiduslynx (approved):

scalar/uuid.go
```diff
@@ -63,8 +63,15 @@ func (s *UUID) Set(src any) error {
 		return s.Set(value2)
 	case [16]byte:
```
candiduslynx, line 64: I wonder if `uuid.UUID` should also be checked?...

scalar/uuid_test.go
```diff
@@ -54,6 +57,8 @@ func TestUUIDSet(t *testing.T) {
 			source: &UUID{Value: [16]byte{0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15}, Valid: true},
 			result: UUID{Value: [16]byte{0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15}, Valid: true},
 		},
+		{source: nilPointerByteArray, result: UUID{}},
```
candiduslynx, line 60: please add `uuid.UUID` test

scalar/uuid.go
erezrokah: It works since `uuid.UUID` implements `Stringer` interface

PR #938, opened 2023-06-01
feat(scalar): Support pointer dereferencing in decimal

erezrokah:
#### Summary

Follow up to https://github.com/cloudquery/plugin-sdk/pull/937, also needed for https://github.com/cloudquery/cloudquery/pull/11115

When reading data from a database it is common to pass a pointer to a value to support reading `nil` values, so the type of the value read is a pointer to a pointer.

Without this PR, if someone uses the decimal scalar they need to dereferences the pointer manually before setting the scalar value.

---

- [ ] Update or add tests 🧪

### DIFF
--- a/scalar/decimal.go
+++ b/scalar/decimal.go
@@ -166,6 +166,9 @@ func (s *Decimal256) Set(val any) error {
 		}
 		return s.Set(*value)
 	default:
+		if originalSrc, ok := underlyingPtrType(val); ok {
+			return s.Set(originalSrc)
+		}
 		return &ValidationError{Type: s.DataType(), Msg: noConversion, Value: value}
 	}
 	s.Valid = true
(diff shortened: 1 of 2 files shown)


yevgenypats (approved):

yevgenypats (approved):

scalar/decimal.go
```diff
@@ -166,6 +166,9 @@ func (s *Decimal256) Set(val any) error {
 		}
 		return s.Set(*value)
 	default:
+		if originalSrc, ok := underlyingPtrType(val); ok {
+			return s.Set(originalSrc)
```
disq, line 170: I would add: `if originalSrc == nil { s.Valid = false; return }`

disq (approved):

PR #948, opened 2023-06-04
fix(types): Extensions conversion with storage

candiduslynx:


types/inet.go
```diff
@@ -154,6 +149,13 @@ func (a *InetArray) Value(i int) *net.IPNet {
...
 		return nil
 	}
+	cidr := string(a.Storage().(*array.Binary).Value(i))
+	if len(cidr) == 0 {
+		return &net.IPNet{
+			IP:   net.IPv4zero,
+			Mask: make(net.IPMask, 4),
+		}
+	}
 	_, ipnet, err := net.ParseCIDR(string(a.Storage().(*array.Binary).Value(i)))
```
yevgenypats, line 159: seems you can call `cidr` here you call `(a.Storage().(*array.Binary).Value(i))` before that

types/inet.go
```diff
@@ -154,6 +149,13 @@ func (a *InetArray) Value(i int) *net.IPNet {
 	if a.IsNull(i) {
 		return nil
 	}
+	cidr := string(a.Storage().(*array.Binary).Value(i))
+	if len(cidr) == 0 {
+		return &net.IPNet{
+			IP:   net.IPv4zero,
+			Mask: make(net.IPMask, 4),
```
yevgenypats, line 156: What is the purpose of `4` here or where is it coming from?

types/mac.go
```diff
@@ -147,7 +143,12 @@ func (a *MACArray) Value(i int) net.HardwareAddr {
 	if a.IsNull(i) {
 		return nil
 	}
-	return net.HardwareAddr(a.Storage().(*array.Binary).Value(i))
+	b := a.Storage().(*array.Binary).Value(i)
+	if len(b) == 0 {
+		return make(net.HardwareAddr, 6)
```
yevgenypats, line 148: What is the purpose of `6` here or where is it coming from?

yevgenypats (changes requested):
Few minor comments/questions.

types/inet.go
candiduslynx: the `ipv4Len` is private in `net`, but this is just to make the mask of the same size. I can replace this with the len of `net.IPv4zero`

types/inet.go
candiduslynx: Done in a8076df52d738719811bd89c894f3c2a11f1f512

types/mac.go
candiduslynx: minimal proper hardware addr (it was the same in append empty value), that will result in `"00:00:00:00:00:00"`

candiduslynx: Added const in 131cc3760eaf70d0cc9ea49a0495bead200959f4

types/inet.go
candiduslynx: Done in a8076df52d738719811bd89c894f3c2a11f1f512

yevgenypats (approved):

yevgenypats (approved):

yevgenypats (approved):

PR #956, opened 2023-06-07
fix(transformers): Ability to transform `any` with TypeTransformer

disq:
plus testing fixes.

Without this there was no way to transform `any` (unless it's defined as `[]any`) because bare interfaces were ignored in `isTypeIgnored()` along with channels/funcs/unsafeptrs.


transformers/struct.go
```diff
@@ -239,6 +239,9 @@ func (t *structTransformer) addColumnFromField(field reflect.StructField, parent
 			return fmt.Errorf("failed to transform type for field %s: %w", field.Name, err)
 		}
 	}
+	if columnType == nil {
+		return nil // ignored
```
disq, line 243: This ensures backwards compatibility with the current situation where we ignore `reflect.Interface` early in `isTypeIgnored`.

transformers/struct.go
```diff
@@ -348,6 +350,9 @@ func defaultGoTypeToSchemaType(v reflect.Type) (arrow.DataType, error) {
 		}
 		return arrow.ListOf(elemValueType), nil
 
+	case reflect.Interface:
+		return nil, nil // silently ignore
```
disq, line 354: This ensures we don't panic now that we don't ignore this kind in `isTypeIgnored`.

transformers/struct_test.go
```diff
@@ -405,11 +459,21 @@ func TestTableFromGoStruct(t *testing.T) {
...
+
 			for i, col := range table.Columns {
 				if !arrow.TypeEqual(col.Type, tt.want.Columns[i].Type) {
-					t.Fatalf("column %s does not match expected type. got %v, want %v", col.Name, col.Type, tt.want.Columns[i].Type)
+					t.Fatalf("column %q does not match expected type. got %v, want %v", col.Name, col.Type, tt.want.Columns[i].Type)
+				}
+			}
+			for _, exc := range tt.want.Columns {
+				if c := table.Column(exc.Name); c == nil {
+					t.Fatalf("column %q not found. want: %v", exc.Name, exc.Type)
```
hermanschaaf, line 473: Nice 👍

hermanschaaf (approved):

disq:
Ran this on _all_ source plugins - no changes.

yevgenypats (approved):

PR #964, opened 2023-06-12
fix: Don't write last batch in managed writer if the context was canceled

candiduslynx:


plugins/destination/managed_writer.go
```diff
@@ -35,24 +35,37 @@ func (p *Plugin) worker(ctx context.Context, metrics *Metrics, table *schema.Tab
...
 			if len(resources) > 0 {
 				p.flush(ctx, metrics, table, resources)
-				resources = make([]arrow.Record, 0)
+				resources = resources[:0] // allows for mem reuse
 				sizeBytes = 0
 			}
 			done <- true
+		case <-ctx.Done():
+			// this means the request was cancelled
+			func() {
```
disq, line 58:
```suggestion
			{
```

plugins/destination/managed_writer.go
```diff
@@ -35,24 +35,37 @@ func (p *Plugin) worker(ctx context.Context, metrics *Metrics, table *schema.Tab
...
+				// we need to use a separate ctx
+				ctx, cancel := context.WithTimeout(context.Background(), p.batchTimeout)
+				defer cancel()
+
+				p.flush(ctx, metrics, table, resources)
+
+				resources = resources[:0]
+				sizeBytes = 0
+			}()
```
disq, line 67:
```suggestion
			}
```

disq (commented):
Easier to read this way?

plugins/destination/managed_writer.go
candiduslynx: I need the func to use defer

disq: you `return` in the end anyway, that block isn't needed?

yevgenypats (changes requested):
Not sure I understand why this is needed or what it solves. If context is cancelled then it means the user or something requested to exit then we can exit as quickly as possible without writing more stuff.

candiduslynx:
Currently the logic will try flushing out the last batch per done channel being closed.
This will also result in UB for the drivers and possibly panics (DuckDB).
I can change the code to just exit if we see the CTX cancelled instead, wdyt?
@yevgenypats

plugins/destination/managed_writer.go
candiduslynx: removed the call to `flush` completely

plugins/destination/managed_writer.go
```diff
@@ -35,24 +35,27 @@ func (p *Plugin) worker(ctx context.Context, metrics *Metrics, table *schema.Tab
 			}
 			if len(resources) == p.spec.BatchSize || sizeBytes+util.TotalRecordSize(r) > int64(p.spec.BatchSizeBytes) {
 				p.flush(ctx, metrics, table, resources)
-				resources = make([]arrow.Record, 0)
+				resources = resources[:0] // allows for mem reuse
```
yevgenypats, line 38: why this is better? then we should pre-allocate it in line `26` otherwise the buffer size will be quite random.

yevgenypats (approved):

plugins/destination/managed_writer.go
candiduslynx: This just allows for not-allocating extra space if we already allocated enough in the previous append cycle

PR #982, opened 2023-06-23
fix(types-inet): Align logic with scalar package, set `net.IPNet` `IP` field after parsing `ParseCIDR`

erezrokah:
#### Summary

Discovered this while working on the PostgreSQL source v4 migration.
When we set string values to the `inet` scalar we set the `IP` field on the resulting `net.IPNet`:
https://github.com/cloudquery/plugin-sdk/blob/097621f02e4fe1258290c3bfdd744a8fc3ab1c15/scalar/inet.go#L85

However the `inet` arrow type loses that information when returning the value:
https://github.com/cloudquery/plugin-sdk/blob/097621f02e4fe1258290c3bfdd744a8fc3ab1c15/types/inet.go#L159

This means that if you create an arrow `inet` type from the string `192.168.0.1/24`, then call `ValueStr` you get back `192.168.0.0/24`

---

- [ ] Update or add tests 🧪

### DIFF
--- a/types/inet_test.go
+++ b/types/inet_test.go
@@ -10,10 +10,11 @@ import (
 )
 
 func mustParseInet(s string) *net.IPNet {
-	_, ipnet, err := net.ParseCIDR(s)
+	ip, ipnet, err := net.ParseCIDR(s)
 	if err != nil {
 		panic(err)
 	}
+	ipnet.IP = ip
 	return ipnet
 }
 
@@ -36,8 +37,9 @@ func TestInetBuilder(t *testing.T) {
 		mustParseInet("192.168.0.0/27"),
 	}
 	b.AppendValues(values, nil)
+	b.Append(mustParseInet("192.168.0.1/24"))
 
-	require.Equal(t, 6, b.Len(), "unexpected Len()")
+	require.Equal(t, 7, b.Len(), "unexpected Len()")
 
 	a := b.NewArray()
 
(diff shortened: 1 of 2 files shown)


disq (approved):

types/inet_test.go
```diff
@@ -58,7 +60,7 @@ func TestInetBuilder(t *testing.T) {
 	require.NoError(t, err)
 
 	a = b.NewArray()
-	require.Equal(t, `["192.168.0.0/24" (null) "192.168.0.0/25" (null) "192.168.0.0/26" "192.168.0.0/27"]`, a.String())
+	require.Equal(t, `["192.168.0.0/24" (null) "192.168.0.0/25" (null) "192.168.0.0/26" "192.168.0.0/27" "192.168.0.1/24"]`, a.String())
```
candiduslynx, line 63: Based on the mask/cidr `192.168.0.1/24` ~ `192.168.0.0/24` ~ `192.168.0.255/24`
Do we really need this?

candiduslynx (commented):
not sure if that's required

types/inet_test.go
erezrokah: If we don't do this the following test case fails:
https://github.com/cloudquery/cloudquery/pull/11719/files#r1239556847

I think it has value saving the original IP address the inet is based on, even if the values could be considered equal.
Another option is to change the `scalar` logic to ignore the IP address, see https://github.com/cloudquery/plugin-sdk/blob/bc9e6e1ae1bff7a6f2c2699a8a113323d689f1a5/scalar/inet.go#L85

But we should be consistent in both cases (either set it, or ignore it)

candiduslynx: IMO we need to verify only that single IP `192.168.0.1` is stored and obtained correctly, and then leave the code here as is

erezrokah: > IMO we need to verify only that single IP 192.168.0.1 is stored and obtained correctly, and then leave the code here as is

Can you please add more context? Do you mean verify in the plugin tests? We have an inconsistency with how we handle the result of `ParseCIDR` between our `inet` arrow type and the scalar `inet` type. We could drop the usage of `scalar` and only use arrow types maybe to align the logic.
Also `inet` can store both an IP and an subnet, so I'm not sure what you mean by `a single IP`, see https://www.postgresql.org/docs/current/datatype-net-types.html

candiduslynx: I'm fine with any solution, an, if the inet can also hold the mask, then let's add this fix

candiduslynx (approved):

disq:
I think it's a good time to merge this, before the final v4 release. @yevgenypats

yevgenypats (approved):

PR #997, opened 2023-06-26
fix: Set Sync option fields

hermanschaaf:
Set the fields obtained from the protobuf so that they are passed to the plugin Sync call.

### DIFF
--- a/internal/servers/plugin/v3/plugin.go
+++ b/internal/servers/plugin/v3/plugin.go
@@ -70,8 +70,10 @@ func (s *Server) Sync(req *pb.Sync_Request, stream pb.Plugin_SyncServer) error {
 	ctx := stream.Context()
 
 	syncOptions := plugin.SyncOptions{
-		Tables:     req.Tables,
-		SkipTables: req.SkipTables,
+		Tables:              req.Tables,
+		SkipTables:          req.SkipTables,
+		SkipDependentTables: req.SkipDependentTables,
+		DeterministicCQID:   req.DeterministicCqId,
 	}
 
 	go func() {
--- a/scheduler/scheduler.go
+++ b/scheduler/scheduler.go
@@ -41,6 +41,15 @@ var AllSchedulerNames = [...]string{
 	StrategyRoundRobin: "round-robin",
 }
 
+func StrategyForName(s string) (Strategy, error) {
+	for i, name := range AllSchedulerNames {
+		if name == s {
+			return AllSchedulers[i], nil
+		}
+	}
+	return StrategyDFS, fmt.Errorf("unknown scheduler strategy: %s", s)
+}
+
 type Strategies []Strategy
 
 func (s Strategies) String() string {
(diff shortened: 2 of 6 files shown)


internal/servers/plugin/v3/plugin.go
```diff
@@ -70,8 +70,10 @@ func (s *Server) Sync(req *pb.Sync_Request, stream pb.Plugin_SyncServer) error {
 	ctx := stream.Context()
 
 	syncOptions := plugin.SyncOptions{
-		Tables:     req.Tables,
-		SkipTables: req.SkipTables,
+		Tables:              req.Tables,
+		SkipTables:          req.SkipTables,
+		SkipDependentTables: req.SkipDependentTables,
+		DeterministicCQID:   req.DeterministicCqId,
```
hermanschaaf, line 76: Actually, no, this needs to be passed as part of Init because it's passed on to the scheduler (unless we refactor the scheduler)

hermanschaaf: I think we should rather change the scheduler though

hermanschaaf: Refactored and updated this PR

scheduler/scheduler_dfs.go
```diff
@@ -81,7 +81,7 @@ func (s *Scheduler) syncDfs(ctx context.Context, resolvedResources chan<- *schem
 	wg.Wait()
 }
 
-func (s *Scheduler) resolveTableDfs(ctx context.Context, table *schema.Table, client schema.ClientMeta, parent *schema.Resource, resolvedResources chan<- *schema.Resource, depth int) {
+func (s *Scheduler) resolveTableDfs(ctx context.Context, table *schema.Table, client schema.ClientMeta, parent *schema.Resource, resolvedResources chan<- *schema.Resource, depth int, syncOpts *SyncOptions) {
```
hermanschaaf, line 84: These functions have a lot of parameters now but I think we can refactor this at a later point

yevgenypats (changes requested):
looks good. one nit.

scheduler/scheduler.go
```diff
@@ -143,11 +157,16 @@ func (s *Scheduler) SyncAll(ctx context.Context, tables schema.Tables) (message.
 	return messages, err
 }
 
-func (s *Scheduler) Sync(ctx context.Context, tables schema.Tables, res chan<- message.Message) error {
+func (s *Scheduler) Sync(ctx context.Context, tables schema.Tables, res chan<- message.Message, opts ...SyncOption) error {
```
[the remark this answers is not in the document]
hermanschaaf: Example usage from Homebrew plugin:

```
func (c *Client) Sync(ctx context.Context, options plugin.SyncOptions, res chan<- message.Message) error {
	tt, err := c.tables.FilterDfs(options.Tables, options.SkipTables, options.SkipDependentTables)
	if err != nil {
		return err
	}
	return c.scheduler.Sync(ctx, tt, res, scheduler.WithSyncDeterministicCQID(options.DeterministicCQID))
}
```

yevgenypats (approved):

PR #1002, opened 2023-06-27
fix: Put null helpers back

candiduslynx:
Return of the code from #913

### DIFF
--- a/plugin/nulls.go
+++ b/plugin/nulls.go
@@ -46,7 +46,7 @@ func stripNullsFromLists(records []arrow.Record) {
 
 type AllowNullFunc func(arrow.DataType) bool
 
-// TODO: use in v4
+// TODO(v4): use in v4
 //
 //nolint:unused
 func (f AllowNullFunc) replaceNullsByEmpty(records []arrow.Record) {
--- a/plugin/nulls.go
+++ b/plugin/nulls.go
@@ -6,6 +6,9 @@ import (
 	"github.com/apache/arrow/go/v13/arrow/memory"
 )
 
+// TODO: use in v4
+//
+//nolint:unused
 func stripNullsFromLists(records []arrow.Record) {
 	for i := range records {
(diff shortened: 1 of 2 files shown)


plugin/nulls.go
```diff
@@ -0,0 +1,74 @@
+package plugin
+
+import (
+	"github.com/apache/arrow/go/v13/arrow"
+	"github.com/apache/arrow/go/v13/arrow/array"
+	"github.com/apache/arrow/go/v13/arrow/memory"
+)
+
+//nolint:unused
+func stripNullsFromLists(records []arrow.Record) {
```
hermanschaaf, line 10: Why are we adding unused code? Will there be a follow-up where it gets used?

candiduslynx: No, I was just migrating CH when I noticed this missing.

I can comment out the relevant section in CH tests, but it will still be required once we bring the vast testing back

plugin/nulls.go
```diff
@@ -0,0 +1,78 @@
+package plugin
+
+import (
+	"github.com/apache/arrow/go/v13/arrow"
+	"github.com/apache/arrow/go/v13/arrow/array"
+	"github.com/apache/arrow/go/v13/arrow/memory"
+)
+
+// TODO: use in v4
```
hermanschaaf, line 9:
```suggestion
// TODO(v4): use in v4
```

I'm using this convention for easier grepping

plugin/nulls.go
hermanschaaf: We can leave this here as a reminder, but can we rather add it back when it's needed?

candiduslynx: 💯

yevgenypats (approved):
Im approving to unblock but in general I think we should somehow come with an approach that is either on the plugin side or generalize better otherwise the testing code is going to get out of hand quickly.

PR #1011, opened 2023-06-28
fix(writers): Move to sub packages

disq:


writers/streamingbatchwriter/streamingbatchwriter.go
```diff
@@ -1,4 +1,4 @@
-package writers
+package streamingbatchwriter
```
hermanschaaf, line 1: Just for my own reference, what's the difference between streaming batch writer and the other two batch writers?

It would also be good to add the answer to this question as a comment in the package, as well as a similar one for BatchWriter, so users can more easily figure out when to use which one.

disq: @hermanschaaf added a short blurb

writers/writers.go
```diff
@@ -0,0 +1,17 @@
...
+	"github.com/cloudquery/plugin-sdk/v4/message"
+)
+
+type Writer interface {
+	Write(ctx context.Context, res <-chan message.WriteMessage) error
+}
+
+const (
+	DefaultBatchTimeoutSeconds = 20
+	DefaultBatchSize           = 10000
```
yevgenypats, line 15: I guess those are not from this PR but I actually think those shouldn't exist and each plugin must define it's own because what would happen if we change the default - it would change on all plugins without us really being able to test what is the impact. cc @herman

yevgenypats (approved):

PR #1012, opened 2023-06-28
fix(writers): Require explicit batch options, remove defaults

disq:
If we want to make _absolutely sure_ that each dest sets its own sensible batch settings, this could be the only way.

Zero as batch timeout isn't supported (it just means it will flush all the time with `<-time.After(0)`). There's no 'clean' way to prevent that as long as we're in that `select`.

### DIFF
--- a/writers/batchwriter/batchwriter.go
+++ b/writers/batchwriter/batchwriter.go
@@ -47,38 +47,20 @@ func WithLogger(logger zerolog.Logger) Option {
 	}
 }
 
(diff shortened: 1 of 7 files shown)


yevgenypats:
I think having the defaults is fine I just wouldn't export them (if this is with regards to my previous comment).

disq:
Trying https://github.com/cloudquery/plugin-sdk/pull/1013 instead.

PR #1015, opened 2023-06-29
fix: Add back testing for all types

yevgenypats:
There is still more work to do to make the testing function "production" ready because they become quite vital and some of them are public APIs so we need to think how to generalize them correctly so they are testable, expand etc.. but this should be a good start just so we can start releasing some destinations

### DIFF
--- a/plugin/testing_upsert.go
+++ b/plugin/testing_upsert.go
@@ -74,7 +74,7 @@ func (s *WriterTestSuite) testUpsertAll(ctx context.Context) error {
 		return fmt.Errorf("failed to create table: %w", err)
 	}
 
-	records := schema.GenTestData(table, schema.GenTestDataOptions{MaxRows: 1})
+	records := schema.GenTestData(table, schema.GenTestDataOptions{MaxRows: 2})
 	record := records[0]
 
 	if err := s.plugin.writeOne(ctx, &message.WriteInsert{
--- a/plugin/testing_upsert.go
+++ b/plugin/testing_upsert.go
@@ -12,8 +12,8 @@ import (
 	"github.com/cloudquery/plugin-sdk/v4/schema"
 )
 
-func (s *WriterTestSuite) testUpsert(ctx context.Context) error {
-	tableName := fmt.Sprintf("cq_test_upsert_%d", time.Now().Unix())
+func (s *WriterTestSuite) testUpsertBasic(ctx context.Context) error {
+	tableName := fmt.Sprintf("cq_upsert_basic_%d", time.Now().Unix())
 	table := &schema.Table{
 		Name: tableName,
 		Columns: []schema.Column{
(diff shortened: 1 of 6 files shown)


schema/testdata.go
```diff
@@ -242,19 +64,102 @@ func columnsToFields(columns ...Column) []arrow.Field {
...
+	if !opts.SkipTimes {
+		columns = append(columns, ColumnList{
+			{Name: "time32_s", Type: arrow.FixedWidthTypes.Time32s},
+			{Name: "time32_ms", Type: arrow.FixedWidthTypes.Time32ms},
+			{Name: "time64_us", Type: arrow.FixedWidthTypes.Time64us},
+			{Name: "time64_ns", Type: arrow.FixedWidthTypes.Time64ns},
+		}...)
+	}
+
+	if !opts.SkipTimestamps {
```
candiduslynx, line 127: Theres' an issue here that @hermanschaaf covered in original implementation: we added the `timestamp_us` option always, as this was the default transformed from old cq time.

plugin/testing_upsert.go
```diff
@@ -63,3 +63,50 @@ func (s *WriterTestSuite) testUpsert(ctx context.Context) error {
...
+	tableName := fmt.Sprintf("cq_upsert_all_%d", time.Now().Unix())
+	table := schema.TestTable(tableName, s.genDatOptions)
+	table.Columns = append(table.Columns, schema.Column{Name: "name", Type: arrow.BinaryTypes.String, PrimaryKey: true})
+	if err := s.plugin.writeOne(ctx, &message.WriteMigrateTable{
+		Table: table,
+	}); err != nil {
+		return fmt.Errorf("failed to create table: %w", err)
+	}
+
+	records := schema.GenTestData(table, schema.GenTestDataOptions{MaxRows: 1})
```
hermanschaaf, line 77: I think we should use `MaxRows: 2` here (and update the expectations accordingly) so that we are also testing a row with null values

plugin/testing_write_insert.go
```diff
@@ -71,3 +71,53 @@ func (s *WriterTestSuite) testInsert(ctx context.Context) error {
...
+func (s *WriterTestSuite) testInsertAll(ctx context.Context) error {
+	tableName := fmt.Sprintf("cq_insert_all_%d", time.Now().Unix())
+	table := schema.TestTable(tableName, s.genDatOptions)
+	if err := s.plugin.writeOne(ctx, &message.WriteMigrateTable{
+		Table: table,
+	}); err != nil {
+		return fmt.Errorf("failed to create table: %w", err)
+	}
+	records := schema.GenTestData(table, schema.GenTestDataOptions{
+		MaxRows: 1,
```
hermanschaaf, line 84: Same comment here

plugin/testing_upsert.go
yevgenypats: I did `MaxRows=2` for now but I think it should be removed moving forward as it is a "control flag" which doesn't indicate what it does to the full extant. and we should just have an option `WithNulls` and then just call `GenTestData` twice

schema/testdata.go
yevgenypats: I'll address it in a follow-up PR.

plugin/testing_upsert.go
hermanschaaf: Agreed, that would be better 👍

schema/testdata.go
hermanschaaf: It's relatively easy to fix I think, just add `timestamp_us` to the initial list of columns

hermanschaaf (approved):

PR #1020, opened 2023-06-29
fix(writers): Allow zero timeout, remove unused timeout options from mixedbatchwriter

disq:


writers/batchwriter/batchwriter.go
```diff
@@ -133,7 +133,7 @@ func (w *BatchWriter) worker(ctx context.Context, tableName string, ch <-chan *m
 
 			resources = append(resources, r)
 			sizeBytes += util.TotalRecordSize(r.Record)
-		case <-time.After(w.batchTimeout):
+		case <-timer(w.batchTimeout):
```
disq, line 136: Receiving from a `nil` channel blocks forever.

writers/streamingbatchwriter/streamingbatchwriter_test.go
```diff
@@ -280,6 +284,65 @@ func TestStreamingBatchTimeout(t *testing.T) {
...
+	}
+
+	ch <- &message.WriteInsert{
+		Record: record,
+	}
+	ch <- &message.WriteInsert{
+		Record: record,
+	}
+
+	time.Sleep(time.Millisecond * 250)
```
hermanschaaf, line 327: Any way we can avoid the sleep here? Or at least sleep a smaller amount of time at regular intervals, then time out after some period of time. That should make it less flakey on CI systems where there can be resource contention

disq: We could use a fake time object of sorts to emulate timeout conditions in the future. batchingwriter also has some time.Sleeps.

hermanschaaf (approved):

yevgenypats (approved):

PR #1030, opened 2023-06-30
fix(writers): Remove `Close()` requirement from BatchWriter, better testing

disq:
`Close()` seems to be easy to forget and meaningless if we just flush things when the channel is closed. The buffered-channel testing was preventing us from doing that, so that's also upgraded to goroutines and mock timers.

Also we don't need to call `batchwriter.Flush()` after the `.Write()` call as it's always flushed when Write returns.


disq:
This would create a swarm of new workers every time the writer channel is closed (which is per Client...) so any type of heavy multiplexing would be adversely affected.

PR #1038, opened 2023-07-01
feat(writers): More unimplemented writer helpers

disq:
Just to get the other two in parity with streamingbatchwriter helpers, plus tests to make sure the interfaces are satisfied.


yevgenypats:
In which cases this is used?

disq:
> In which cases this is used?

This can be used in some of the schemaless destinations to get rid of the empty MigrateTable implementations.

yevgenypats (approved):

yevgenypats:
Shouldn't they though return an error instead of `nil` if not implemented. For example in BigQuery I need to return an error on `DeleteStale`

disq:
> Shouldn't they though return an error instead of `nil` if not implemented. For example in BigQuery I need to return an error on `DeleteStale`

These are to 'ignore' the messages, but I think DeleteStale should return an error, but not MigrateTable? But I don't like the potential disparity between those two as both are 'Unimplemented*' and one returns an error and the other ignores it.

disq:
Maybe we should name them `UnimplementedDeleteStale` vs. `IgnoreMigrateTable`? @yevgenypats

yevgenypats:
> Maybe we should name them `UnimplementedDeleteStale` vs. `IgnoreMigrateTable`? @yevgenypats

Yeah that seems like a good option.

PR #1039, opened 2023-07-01
fix: State add flush and fix migration bug

yevgenypats:
This fixes a bug in state backend where table weren't created.  

I also tested it with https://github.com/cloudquery/cloudquery/pull/11855 which seems to work with minor modifications (will open a PR shortly to the CLI PR)


internal/clients/state/v3/state.go
```diff
@@ -158,5 +160,5 @@ func (c *Client) GetKey(_ context.Context, key string) (string, error) {
 	if val, ok := c.mem[key]; ok {
 		return val, nil
 	}
-	return "", fmt.Errorf("key not found")
+	return "", nil
```
yevgenypats, line 163: I decided to return `""` if not found but we can later on make this an option which the user can decide if they want to use `""` or `NotFoundErr`

state/state.go
```diff
@@ -14,6 +14,7 @@ import (
 type Client interface {
 	SetKey(ctx context.Context, key string, value string) error
 	GetKey(ctx context.Context, key string) (string, error)
+	Flush(ctx context.Context) error
```
disq, line 17: What's flush for?

yevgenypats: It is not flushing the content every `Set` otherwise data is not written back. It is already implemented on the `state` implementation but just wasn't exposed

PR #1045, opened 2023-07-03
fix(writers): Require `Close()` for `StreamingBatchWriter`

disq:
So that workers aren't stopped when a single client's data ends. Calling `Close()` multiple times (or calling `Write` afterwards) would end up in a panic.

### DIFF
--- a/writers/streamingbatchwriter/streamingbatchwriter.go
+++ b/writers/streamingbatchwriter/streamingbatchwriter.go
@@ -158,6 +158,8 @@ func (w *StreamingBatchWriter) Close(context.Context) error {
 	}
 	w.workersWaitGroup.Wait()
 
+	w.insertWorkers = nil
+
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


writers/streamingbatchwriter/streamingbatchwriter.go
```diff
@@ -158,9 +158,9 @@ func (w *StreamingBatchWriter) stopWorkers() {
 	}
 	w.workersWaitGroup.Wait()
 
-	w.insertWorkers = make(map[string]*streamingWorkerManager[*message.WriteInsert])
-	w.migrateWorker = nil
-	w.deleteWorker = nil
```
disq, line 163: Not setting both of these to `nil` means if there's a write of their type, the channel write will panic with 'write on closed channel'

writers/streamingbatchwriter/streamingbatchwriter.go
```diff
@@ -158,9 +158,9 @@ func (w *StreamingBatchWriter) stopWorkers() {
...
+	w.insertWorkers = nil
```
disq, line 161: Setting the map to `nil` so that if there's an Insert after `Close()` we'll get a nil-map panic.

yevgenypats (approved):

PR #1047, opened 2023-07-03
feat: Add state.NoOpClient

hermanschaaf:
Adds a `state.NoOpClient` that can be used by plugins when `plugin.SyncOptions.BackendOptions` is nil:

```
	var stateClient state.Client
	if options.BackendOptions == nil {
		c.logger.Info().Msg("No backend options provided, using no state backend")
		stateClient = state.NoOpClient{}
	} else {
 		...
```

### DIFF
--- a/state/state.go
+++ b/state/state.go
@@ -31,15 +31,15 @@ func NewClient(ctx context.Context, conn *grpc.ClientConn, tableName string) (Cl
 
 type NoOpClient struct{}
 
-func (NoOpClient) SetKey(ctx context.Context, key string, value string) error {
+func (*NoOpClient) SetKey(ctx context.Context, key string, value string) error {
 	return nil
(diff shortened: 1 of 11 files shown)


state/state.go
```diff
@@ -28,3 +28,20 @@ func NewClient(ctx context.Context, conn *grpc.ClientConn, tableName string) (Cl
 	}
 	return nil, fmt.Errorf("please upgrade your state backend plugin. state supporting version 3 plugin has %v", versions.Versions)
 }
+
+type NoOpClient struct{}
+
+func (NoOpClient) SetKey(ctx context.Context, key string, value string) error {
```
yevgenypats, line 34: should it be pointer here?

yevgenypats (approved):

PR #1053, opened 2023-07-03
feat: Add `WriteInserts.GetRecords()`

disq:


message/write_message.go
```diff
@@ -86,6 +86,14 @@ func (m WriteInserts) Exists(tableName string) bool {
 	})
 }
 
+func (m WriteInserts) GetRecords() []arrow.Record {
```
disq, line 89: Could even add `WriteMigrateTables.GetTables()` but the migrate flag throws us. Could return two lists but it wouldn't look good.

yevgenypats (approved):

PR #1055, opened 2023-07-04
feat: Add batch timeout support to mixed batch writer

hermanschaaf:


writers/mixedbatchwriter/mixedbatchwriter.go
```diff
@@ -94,31 +116,42 @@ func (w *MixedBatchWriter) Write(ctx context.Context, msgChan <-chan message.Wri
...
+				err = insert.append(ctx, v)
+			case *message.WriteDeleteStale:
+				err = deleteStale.append(ctx, v)
+			default:
+				panic("unknown message type")
+			}
+			if err != nil {
+				return err
+			}
+		case <-tick:
```
disq, line 147: Wouldn't this drift and trigger before the interval is done, if the flush() in `<-msgChan` takes long? `tick` is reset only when a previous tick happens?

hermanschaaf: I believe this is what we want. The batch timeout should be a guarantee that a write will never be delayed by more than X amount of time. If we don't do it this way, then a slow trickle of messages into `<-msgChan` could cause a long delay before a batch is flushed. (Imagine batch size of 10000, batch timeout of 20 seconds, one message arriving per second. If we do it the way in this PR, batches of 20 messages will flush periodically. If not, the batch will only flush after 10000 seconds, because the timer gets reset every time a message is received)

If the flush in `<-msgChan` takes long, then yes, the ticker will trigger with a potentially empty batch, but I don't see it as a problem and it's necessary to provide the guarantee of timely delivery.

disq (approved):

yevgenypats (approved):

PR #1058, opened 2023-07-04
refactor(ticker): Don't reset manually

candiduslynx:
Follow-up for https://github.com/cloudquery/plugin-sdk/pull/1056

### DIFF
--- a/writers/batchwriter/batchwriter.go
+++ b/writers/batchwriter/batchwriter.go
@@ -122,7 +122,7 @@ func (w *BatchWriter) Close(context.Context) error {
 func (w *BatchWriter) worker(ctx context.Context, tableName string, ch <-chan *message.WriteInsert, flush <-chan chan bool) {
 	sizeBytes := int64(0)
 	resources := make([]*message.WriteInsert, 0, w.batchSize)
-	tick := timer(w.batchTimeout)
+	tick := time.Tick(w.batchTimeout)
 	for {
 		select {
 		case r, ok := <-ch:
@@ -145,7 +145,6 @@ func (w *BatchWriter) worker(ctx context.Context, tableName string, ch <-chan *m
 				w.flushTable(ctx, tableName, resources)
 				resources, sizeBytes = resources[:0], 0
 			}
-			tick = timer(w.batchTimeout)
 		case done := <-flush:
 			if len(resources) > 0 {
 				w.flushTable(ctx, tableName, resources)
@@ -323,11 +322,4 @@ func (w *BatchWriter) startWorker(ctx context.Context, msg *message.WriteInsert)
 	}()
 	ch <- msg
 	return nil
-}
-
-func timer(timeout time.Duration) <-chan time.Time {
-	if timeout == 0 {
-		return nil
-	}
-	return time.After(timeout)
 }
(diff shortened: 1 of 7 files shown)


hermanschaaf:
What is this fixing?

hermanschaaf:
It seems to me like the other code used to handle a 0 timeout, but this doesn't (I'm not sure what ticker's behavior is if it's passed 0). We don't have tests for that case in these two batch writers right now, but I think a better fix here would be to add a test for a 0 timeout?

disq:
Not supporting a zero/disabled batchtimeout doesn't sit well with the file destinations (with NoRotate or filenames without UUID).

candiduslynx:
@hermanschaaf @disq 
https://pkg.go.dev/time#Tick
https://cs.opensource.google/go/go/+/refs/tags/go1.20.5:src/time/tick.go;l=68
The impl will ret nil just like the manual implementation that's now removed:
```go
// Tick is a convenience wrapper for NewTicker providing access to the ticking
// channel only. While Tick is useful for clients that have no need to shut down
// the Ticker, be aware that without a way to shut it down the underlying
// Ticker cannot be recovered by the garbage collector; it "leaks".
// Unlike NewTicker, Tick will return nil if d <= 0.
func Tick(d Duration) <-chan Time {
	if d <= 0 {
		return nil
	}
	return NewTicker(d).C
}
```

disq (commented):
Missing the fix for streamingbatchwriter :)

hermanschaaf:
Isn't the leak mentioned in the comment of `time.Tick` also something consider here?

disq:
> Isn't the leak mentioned in the comment of `time.Tick` also something consider here?

I think it is if we're going to have long running writers.

candiduslynx:
Fixed leak in e63bc5e260085294691b371f717a164267b451cf

writers/batchwriter/batchwriter.go
```diff
@@ -325,9 +325,9 @@ func (w *BatchWriter) startWorker(ctx context.Context, msg *message.WriteInsert)
 	return nil
 }
 
-func timer(timeout time.Duration) <-chan time.Time {
-	if timeout == 0 {
-		return nil
+func ticker(interval time.Duration) (ch <-chan time.Time, stop func()) {
+	if t := time.NewTicker(interval); t != nil {
```
disq, line 329: 0 interval would panic:
```go
func NewTicker(d Duration) *Ticker {
	if d <= 0 {
		panic(errors.New("non-positive interval for NewTicker"))
	}
```

disq: So something like `if interval == 0 { // create and empty channel and return it, stop fn to close the empty channel` is needed

disq:
Same ticker will be used for all writers, maybe move this to the upper `writers` package? We use `MsgID` from that as well.

writers/batchwriter/batchwriter.go
candiduslynx: 68015d39fd8c4ab30b82969e283a2651cedcfe8b

candiduslynx:
@disq 

> Same ticker will be used for all writers, maybe move this to the upper `writers` package? We use `MsgID` from that as well.

Done in 68015d39fd8c4ab30b82969e283a2651cedcfe8b



> Missing the fix for streamingbatchwriter :)

Done in 8f241a0cebc9b706af1c8fc099f993d8a45d741c

disq (approved):

hermanschaaf (approved):

yevgenypats (approved):

PR #1063, opened 2023-07-04
refactor(writers): Don't leak timers

disq:
Alternative to https://github.com/cloudquery/plugin-sdk/pull/1058

### DIFF
--- a/writers/batchwriter/batchwriter.go
+++ b/writers/batchwriter/batchwriter.go
@@ -122,7 +122,10 @@ func (w *BatchWriter) Close(context.Context) error {
 func (w *BatchWriter) worker(ctx context.Context, tableName string, ch <-chan *message.WriteInsert, flush <-chan chan bool) {
 	sizeBytes := int64(0)
 	resources := make([]*message.WriteInsert, 0, w.batchSize)
-	tick := timer(w.batchTimeout)
+	tick, tickClose := timer(w.batchTimeout)
+	defer func() {
+		tickClose()
+	}()
 	for {
 		select {
 		case r, ok := <-ch:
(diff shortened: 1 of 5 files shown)


hermanschaaf:
I haven't looked into it yet, but what does `time.Ticker` do if you pass in a 0 duration? If that works, maybe we can simplify by using that?

disq:
> I haven't looked into it yet, but what does `time.Ticker` do if you pass in a 0 duration? If that works, maybe we can simplify by using that?

Looks like it will fire immediately (after a certain unspecified) delay:

```go
func when(d Duration) int64 {
	if d <= 0 {
		return runtimeNano()
	}
```

disq:
We could also use `.Reset(time.Duration)` to re-use the same timer over and over again, would require a follow-up PR though... cc @candiduslynx

writers/batchwriter/batchwriter.go
```diff
@@ -145,7 +148,8 @@ func (w *BatchWriter) worker(ctx context.Context, tableName string, ch <-chan *m
 				w.flushTable(ctx, tableName, resources)
 				resources, sizeBytes = resources[:0], 0
 			}
-			tick = timer(w.batchTimeout)
+			tickClose()
```
candiduslynx, line 151: you shouldn't need to manually close on every iteration

candiduslynx (changes requested):

writers/batchwriter/batchwriter.go
disq: incoming commit:

```go
type ticker interface {
	Chan() <-chan time.Time
	Reset(time.Duration)
	Stop() bool
}
```

candiduslynx:
@disq WDYT about https://github.com/cloudquery/plugin-sdk/pull/1058/commits/e63bc5e260085294691b371f717a164267b451cf instead?

PR #986, opened 2023-06-23
fix(scalar): Use `appendTime` instead of `UnixMicro`

erezrokah:
#### Summary

Discovered while working on https://github.com/cloudquery/cloudquery/pull/11719. Using `UnixMicro()` is incorrect as it doesn't account for the internal time unit of the arrow timestamp type. 

`AppendTime` does https://github.com/cloudquery/arrow/blob/0d1f7234124e67b98b3aa0aa3e521d42c2f3165d/go/arrow/array/timestamp.go#L156 and does the correct conversion https://github.com/cloudquery/arrow/blob/8366a2241e66a53ee05ab4a832927e33e5f9c5a2/go/arrow/datatype_fixedwidth.go#L202

---

Use the following steps to ensure your PR is ready to be reviewed

- [x] Read the [contribution guidelines](../blob/main/CONTRIBUTING.md) 🧑‍🎓
- [x] Run `go fmt` to format your code 🖊
- [x] Lint your changes via `golangci-lint run` 🚨 (install golangci-lint [here](https://golangci-lint.run/usage/install/#local-installation))
- [x] Update or add tests 🧪
- [x] Ensure the status checks below are successful ✅

PR #1068, opened 2023-07-04
fix(scalar): Test `AppendTime` on TimestampBuilder

disq:
Brings the test from https://github.com/cloudquery/plugin-sdk/pull/986 (fix was included in https://github.com/cloudquery/plugin-sdk/pull/1064)


erezrokah:
Thanks for adding the tests @disq 🚀

PR #1069, opened 2023-07-04
feat(transformers): Add `Apply` to apply extra transformations

disq:
// Apply applies the given transformers to the given Tables, and recursively on the relations.
// This is useful for applying transformers that are not defined in the table definitions. To apply the table-definition transformers, use TransformTables.

### DIFF
--- a/transformers/tables.go
+++ b/transformers/tables.go
@@ -15,7 +15,7 @@ func SetParents(tables schema.Tables, parent *schema.Table) {
 }
 
 // TransformTables runs given Tables' transformers as defined in the table definitions, and recursively on the relations.
-// To apply additional transformers, use ApplyTransformers.
+// To apply additional transformers, use Apply.
 func TransformTables(tables schema.Tables) error {
 	for _, table := range tables {
 		if table.Transform != nil {
--- a/transformers/tables.go
+++ b/transformers/tables.go
@@ -14,20 +14,30 @@ func SetParents(tables schema.Tables, parent *schema.Table) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


hermanschaaf:
I wonder if we shouldn't rather introduce a function specifically for applying extra transformations to tables, because then you can decide whether to apply the extra transformations before or after the table's defined transformations?

disq:
> I wonder if we shouldn't rather introduce a function specifically for applying extra transformations to tables, because then you can decide whether to apply the extra transformations before or after the table's defined transformations?

Ok, refactored into separate fn

transformers/tables.go
```diff
@@ -28,3 +29,19 @@ func TransformTables(tables schema.Tables) error {
 	}
 	return nil
 }
+
+// ApplyTransformers applies the given transformers to the given Tables, and recursively on the relations.
+// This is useful for applying transformers that are not defined in the table definitions. To apply the table-definition transformers, use TransformTables.
+func ApplyTransformers(tables schema.Tables, extraTransformers ...schema.Transform) error {
```
hermanschaaf, line 35:
```suggestion
func Apply(tables schema.Tables, extraTransformers ...schema.Transform) error {
```

Because it will be called as `transformers.Apply(tables, ...)`

hermanschaaf: Could we add a test for this function? :)

disq: good catch :)

yevgenypats (approved):

hermanschaaf (approved):

PR #1070, opened 2023-07-04
fix(naming): Rename `SyncMessages.InsertMessage()` to `SyncMessages.GetInserts()`

disq:
Also add `SyncInserts.GetRecords()` for parity with `WriteInserts`


yevgenypats:
It will mean we will have to fix it everywhere afterwards :(

yevgenypats (approved):

disq:
Not in many places though

PR #1072, opened 2023-07-05
fix: Deterministic ordering for records returned by readAll in tests

hermanschaaf:
This: 
 - adds an `id` field that is used as an auto-incrementing integer field in tests, and can be used to order records by so that tests can pass regardless of sort order returned by a plugin's Read method
 - makes null rows an explicit option (rather than every other row being null) and starts using null rows in tests again

### DIFF
--- a/plugin/plugin_read.go
+++ b/plugin/plugin_read.go
@@ -2,18 +2,13 @@ package plugin
 
 import (
 	"context"
-	"sort"
 
 	"github.com/apache/arrow/go/v13/arrow"
-	"github.com/apache/arrow/go/v13/arrow/array"
 	"github.com/cloudquery/plugin-sdk/v4/schema"
 )
 
 // readAll is used in tests to read all records from a table.
 func (p *Plugin) readAll(ctx context.Context, table *schema.Table) ([]arrow.Record, error) {
-	// TODO (Go 1.21+) use testing.Testing() to guard against use of this function outside of tests
-	// https://stackoverflow.com/a/75787351
-
 	var err error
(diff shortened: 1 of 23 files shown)


plugin/plugin_read.go
```diff
@@ -0,0 +1,44 @@
...
+	// nolint:prealloc
+	var records []arrow.Record
+	for record := range ch {
+		records = append(records, record)
+	}
+
+	// sort records by "id" column, if present. Because "id" is auto-incrementing in the test
+	// data generator, this should result in records being returned in insertion order.
+	sch := table.ToArrowSchema()
+	if sch.HasField("id") {
```
yevgenypats, line 32: I think this should be done outside of `read` with a helper function because we might want to sort it with different logic and then it will either be impossible or this function would grow.

yevgenypats: The vanilla read should just return it as it is as it is not aware of what columns are there and so on.

yevgenypats (changes requested):
Looks good but have one comment on the design of the read funciton/sorting.

plugin/plugin_read.go
hermanschaaf: This is a private function, I figured we can refactor it later if/when the need arises. But okay, I can do it now

hermanschaaf: Refactored 👍 I also added some random shuffling to the memdb Read function so that we can be sure this is used everywhere it needs to be used

yevgenypats (approved):

PR #1073, opened 2023-07-05
fix: Bring back null handling for destination testing

hermanschaaf:
This brings back the optional removal of nulls when doing test comparisons for destinations that need it.


hermanschaaf:
Closing in favor of https://github.com/cloudquery/plugin-sdk/pull/1074

PR #1074, opened 2023-07-05
fix: Handle null-related test options

candiduslynx:
Follow-up for #1072 & #1002

### DIFF
--- a/plugin/nulls.go
+++ b/plugin/nulls.go
@@ -7,8 +7,8 @@ import (
 )
 
 func stripNullsFromLists(record arrow.Record) arrow.Record {
-	cols := make([]arrow.Array, len(record.Columns()))
-	for c, col := range record.Columns() {
+	cols := record.Columns()
+	for c, col := range cols {
 		list, ok := col.(array.ListLike)
 		if !ok {
 			continue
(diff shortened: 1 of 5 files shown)


disq:
Same as https://github.com/cloudquery/plugin-sdk/pull/1074

hermanschaaf:
This is the same as https://github.com/cloudquery/plugin-sdk/pull/1073 but has some more handling for ClickHouse cases that I couldn't test, so we can go with this one. Very similar either way :)

hermanschaaf:
I get a nil pointer dereference error in `stripNullsFromLists` when I run this one against BigQuery destination though

candiduslynx:
It also puts the sorting where it was missing in the inserts

candiduslynx:
> I get a nil pointer dereference error in `stripNullsFromLists` when I run this one against BigQuery destination though

fixed in fff9fe5

hermanschaaf (approved):

plugin/nulls.go
```diff
@@ -6,73 +6,75 @@ import (
...
+	return array.NewRecord(record.Schema(), cols, record.NumRows())
 }
 
 type AllowNullFunc func(arrow.DataType) bool
 
-// TODO(v4): use in v4
-//
-// nolint:unused
-func (f AllowNullFunc) replaceNullsByEmpty(records []arrow.Record) {
+func (f AllowNullFunc) replaceNullsByEmpty(record arrow.Record) arrow.Record {
```
yevgenypats, line 48: Is the method of a function necessary here?

candiduslynx: moved in 412d679eedeeefaa5a2c0308a6bbe230424abba7

yevgenypats (approved):

PR #1076, opened 2023-07-05
fix: Reset timers on flush

hermanschaaf:
This updates the ticker logic in the batch writers to reset the ticker when a flush happens. This is better, as it still guarantees that a message won't be delayed by more than batch_timeout, but we don't risk flushing a very small batch because we must flush at regular intervals either.

The choice of resetting _after_ the flush is deliberate: it means that the maximum amount of time between flushes is:

```
max_time = flush_time + batch_timeout
```

otherwise we could do:

```
max_time = batch_timeout
```

but if a flush were to then longer than the batch timeout, we can end up in a cycle of flushing again immediately after the previous flush finishes.

### DIFF
--- a/writers/batchwriter/batchwriter.go
+++ b/writers/batchwriter/batchwriter.go
@@ -122,8 +122,8 @@ func (w *BatchWriter) Close(context.Context) error {
 func (w *BatchWriter) worker(ctx context.Context, tableName string, ch <-chan *message.WriteInsert, flush <-chan chan bool) {
 	sizeBytes := int64(0)
 	resources := make([]*message.WriteInsert, 0, w.batchSize)
-	tick, done := writers.NewTicker(w.batchTimeout)
-	defer done()
+	ticker := writers.NewTicker(w.batchTimeout)
+	defer ticker.Stop()
 	for {
 		select {
 		case r, ok := <-ch:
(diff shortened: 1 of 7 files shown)


writers/streamingbatchwriter/streamingbatchwriter_test.go
```diff
@@ -379,9 +379,10 @@ func TestStreamingBatchUpserts(t *testing.T) {
 func waitForLength(t *testing.T, checkLen func(messageType) int, msgType messageType, want int) {
 	t.Helper()
 	lastValue := -1
+	timeout := time.After(5 * time.Second)
 	for {
 		select {
-		case <-time.After(time.Second):
+		case <-timeout:
 			t.Fatalf("timed out waiting for %v message length %d (last value: %d)", msgType, want, lastValue)
 		default:
```
hermanschaaf, line 387: There was a bug here that caused the test to hang indefinitely; moving the timeout up solves it

writers/batchwriter/batchwriter.go
```diff
@@ -122,8 +122,8 @@ func (w *BatchWriter) Close(context.Context) error {
 func (w *BatchWriter) worker(ctx context.Context, tableName string, ch <-chan *message.WriteInsert, flush <-chan chan bool) {
 	sizeBytes := int64(0)
 	resources := make([]*message.WriteInsert, 0, w.batchSize)
-	tick, done := writers.NewTicker(w.batchTimeout)
-	defer done()
+	ticker := writers.NewTicker(w.batchTimeout)
+	defer ticker.Stop()
```
disq, line 126: `ticker.Chan()` should also be closed in the defer. Could maybe add it in a `Stop()` wrapper since we only stop on defer and there would be no erroneous tick read in the `select`:

> Stop turns off a ticker. After Stop, no more ticks will be sent.
> Stop does not close the channel, to prevent a concurrent goroutine
> reading from the channel from seeing an erroneous "tick".

writers/mixedbatchwriter/mixedbatchwriter_test.go
```diff
@@ -214,6 +215,38 @@ func TestMixedBatchWriter(t *testing.T) {
 	}
 }
 
+type mockTicker struct {
```
disq, line 218: Maybe this can be moved to `writers` package next to ticker.go?

yevgenypats (approved):

PR #1085, opened 2023-07-11
feat: Add `plugin.ValidateNoEmptyColumns`

bbernays:
#### Summary

This adds `plugin.ValidateNoEmptyColumns` which will handle all of the empty Column verification. This should be used in nearly every single plugin. 

In the future if we add more validation we can add more validation functions  and have a single wrapper around them like `plugin.ValidateSync()`


message/sync_message.go
```diff
@@ -109,3 +110,27 @@ func (m SyncInserts) GetRecordsForTable(table *schema.Table) []arrow.Record {
 	}
 	return slices.Clip(res)
 }
+
+// Get all records for an  array of tables including any table relations
+func (m SyncInserts) GetRecordsForTables(tables schema.Tables) []arrow.Record {
```
hermanschaaf, line 115: I don't think we really need to add this, because in the tests we still want to know which table a record came from so we can log it as part of the error message. 

I suggest adding a helper function to the SDK, specifically for tests, that takes `t *testing.T, table *schema.Table` does something like this:

```
		records := messages.GetInserts().GetRecordsForTable(table)
		emptyColumns := schema.FindEmptyColumns(table, records)
		if len(emptyColumns) > 0 {
			t.Fatalf("found empty column(s): %v in %s", emptyColumns, table.Name)
		}
	}
```

This way we can still keep the complexity in one place

bbernays: Agreed. Do we have a testing package that this would fit in?

hermanschaaf: The test suite for writers lives in the `plugin` package so it should probably go there for now

disq (approved):

yevgenypats (approved):

Issue #1106, opened 2023-07-20
test: Add a test to validate table names are unique per plugin


erezrokah:
To catch https://github.com/cloudquery/cloudquery/pull/12428

I think we had this so not sure where it went

PR #1107, opened 2023-07-21
refactor: Dedup tables only at the top level in `FlattenTables`

erezrokah:
#### Summary

Noticed this while working on https://github.com/cloudquery/plugin-sdk/issues/1106.
Don't think we need to dedup at each level of the recursion. Also not sure why we're deduping as table names should be unique

---

- [ ] Update or add tests 🧪


hermanschaaf:
If memory serves, deduping was added as a safety measure to make the FlattenTables call idempotent. Back then the `Relations` field also wasn't set to `nil` during `Flatten`, so if you accidentally were to call FlattenTables on an already-flattened list (as did happen in a couple of cases), you'd get duplicate tables, because the child tables would get unrolled again.

It seems to not be necessary anymore, because the `Relations` get set to nil. If we have a test to check that `FlattenTables` is idempotent, then I think we can remove the dedup, but since it's a little higher on the risk scale I would probably not do that right now.

PR #984, opened 2023-06-23
feat!: Update to SDK V4

yevgenypats:
Instead of this https://github.com/cloudquery/plugin-sdk/pull/915 (fix rebase conflicts)

This introduce a single API for both destination and sources. Now plugins should be able to both read and write.
This also goes together with a proto upgrade https://github.com/cloudquery/plugin-pb-go/tree/main/pb/plugin/v3

PR #1108, opened 2023-07-21
fix: Bring back plugin validation

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/plugin-sdk/issues/1106

Plugin validation was removed in https://github.com/cloudquery/plugin-sdk/pull/984/files#diff-95f04007d983a3bfdb080b338b9de5a8bbd73a3d65910fbe664f97788a9021b3.

As a result we don't validate tables and columns anymore, see resulted bug in https://github.com/cloudquery/cloudquery/pull/12428.

This PR brings it back and per a better option puts the validation in the `Init` method. Other options:
1. Have each plugin call `.validate` on the plugin side, probably after transformation [here](https://github.com/cloudquery/cloudquery/blob/03fae6d803a4e2ac00ff83d9e90a4169b247c346/plugins/source/aws/resources/plugin/tables.go#L565). Downside that each plugin needs to implement the call
2. Have the validation done as a part of `schema.TransformTables` [here](https://github.com/cloudquery/plugin-sdk/blob/8a6a2cc87e303d66c810e629a497f9acd3c4982c/transformers/tables.go#L30). Downside it will happen before any custom transformation that are executed on tables like [`titleTransformer`](https://github.com/cloudquery/cloudquery/blob/03fae6d803a4e2ac00ff83d9e90a4169b247c346/plugins/source/aws/resources/plugin/tables.go#L567)

> Please note that before it was removed `validate` was called during `NewPlugin` which means it errored right after calling the `serve` command, meaning the plugin would not start. With this fix the plugin will error out during `Init`, which means you need to run `sync/migrate` to see the error

---

- [ ] Update or add tests 🧪

### DIFF
--- a/schema/table.go
+++ b/schema/table.go
@@ -342,12 +342,13 @@ func (tt Tables) ValidateDuplicateColumns() error {
 }
 
 func (tt Tables) ValidateDuplicateTables() error {
+	tableNames := tt.TableNames()
 	tables := make(map[string]bool, len(tt))
-	for _, t := range tt {
-		if _, ok := tables[t.Name]; ok {
-			return fmt.Errorf("duplicate table %s", t.Name)
+	for _, t := range tableNames {
+		if _, ok := tables[t]; ok {
+			return fmt.Errorf("duplicate table %s", t)
 		}
-		tables[t.Name] = true
+		tables[t] = true
 	}
 	return nil
 }
@@ -405,7 +406,7 @@ func (t *Table) ValidateName() error {
 	if !ok {
 		return fmt.Errorf("table name %q is not valid: table names must contain only lower-case letters, numbers and underscores, and must start with a lower-case letter or underscore", t.Name)
 	}
-	return nil
+	return ValidateTable(t)
 }
 
 func (t *Table) PrimaryKeysIndexes() []int {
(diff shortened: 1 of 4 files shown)


schema/table.go
```diff
@@ -309,12 +309,13 @@ func (tt Tables) ValidateDuplicateColumns() error {
 }
 
 func (tt Tables) ValidateDuplicateTables() error {
+	tableNames := tt.TableNames()
```
erezrokah, line 312: Previous code only validated top level tables. Fixed it so it uses `TableNames()` which gives us relations too, and added a test

schema/table.go
```diff
@@ -372,7 +373,7 @@ func (t *Table) ValidateName() error {
 	if !ok {
 		return fmt.Errorf("table name %q is not valid: table names must contain only lower-case letters, numbers and underscores, and must start with a lower-case letter or underscore", t.Name)
 	}
-	return nil
+	return ValidateTable(t)
```
erezrokah, line 376: There's some refactoring to be done around validation code as we have a separate `schema/validators.go` which also has some code to validate table names. It currently only validates the length, so I added the call here

disq (approved):

yevgenypats (approved):

PR #1109, opened 2023-07-21
fix: Scalar timestamp parsing

disq:
This was broken with https://github.com/cloudquery/plugin-sdk/pull/1095 (at least for the test source, which gets its data from testdata)


scalar/timestamp.go
```diff
@@ -18,6 +18,7 @@ const defaultStringFormat = "2006-01-02 15:04:05.999999999 -0700 MST"
 
 // this is used by arrow string format (time is in UTC)
 const arrowStringFormat = "2006-01-02 15:04:05.999999999"
+const arrowStringFormatNew = "2006-01-02 15:04:05.999999999Z"
```
disq, line 21: Not adding a new line and using the one above seems to break some other things (at least in the test src plugin)

erezrokah (approved):

yevgenypats (commented):
Can we add some tests to that, given it's a "hot path"

disq:
> Can we add some tests to that, given it's a "hot path"

@yevgenypats added

yevgenypats (approved):

PR #1124, opened 2023-08-01
fix: Don't return invalid JSON `""` if it's returned from the storage array

candiduslynx:


disq (approved):

candiduslynx:
the main issue is fixed by #1125 (proper structs & lists unwrapping, so now this code is redundant)

PR #1125, opened 2023-08-01
fix: Process nulls for tested types, too (maps, lists, structs)

candiduslynx:
As there are nested types in Apache Arrow we need to traverse them to properly handle nulls.

This change adds traversal to Structs & ListLike arrays (includes Maps, too, but the proper way, IMO, would be to handle the map items only (keys can't be nullable), however, it may be done as a follow-up when required).

### DIFF
--- a/plugin/nulls_test.go
+++ b/plugin/nulls_test.go
@@ -1,0 +1,62 @@
+package plugin
+
+import (
+	"github.com/apache/arrow/go/v13/arrow"
+	"github.com/apache/arrow/go/v13/arrow/array"
+	"github.com/cloudquery/plugin-sdk/v4/schema"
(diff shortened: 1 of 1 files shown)


hermanschaaf (commented):
Looks good from a quick read through, but since it's complicated can we add a test for this change to check it's correct and prevent future regressions? (Testing the test library 😄)

candiduslynx:
> Looks good from a quick read through, but since it's complicated can we add a test for this change to check it's correct and prevent future regressions? (Testing the test library 😄)

Added in 8cac7c4

disq (approved):

hermanschaaf (approved):

yevgenypats (approved):

PR #1129, opened 2023-08-03
feat: Add more metadata to tables needed for docs generation

hermanschaaf:
This adds metadata fields for table titles and parent relationships, so that this information can be sent over the wire in Arrow format. We will use this in the CLI to generate markdown and json docs for source plugins.


schema/table.go
```diff
@@ -26,7 +26,7 @@ type Transform func(table *Table) error
 
 type Tables []*Table
 
-// This is deprecated
+// Deprecated: SyncSummary is deprecated.
```
hermanschaaf, line 29: This and some other changes here were just some comment clean-up in the file to better conform to Go standards

erezrokah (approved):

disq (approved):

PR #1131, opened 2023-08-04
fix(writers): StreamingBatchWriter hangs with non-append mode

disq:
Fixes the hang [issue](https://github.com/cloudquery/cloudquery/issues/12793) with non-append mode when a dest is using `UnimplementedDeleteStale`.

- This also makes it so that it closes things when Write is done: Otherwise our testing framework confuses things (by invoking Write multiple times for multiple tests in the same plugin, I think).
We're noticing this now because of https://github.com/cloudquery/cloudquery/pull/12765, as other plugins using this writer (file based plugins) don't support migration operations.

Also makes it so that you don't need to call Close() on plugin deinit (but it could still be a good idea if Write was interrupted)

### DIFF
--- a/writers/streamingbatchwriter/unimplemented.go
+++ b/writers/streamingbatchwriter/unimplemented.go
@@ -18,9 +18,12 @@ func (IgnoreMigrateTable) MigrateTable(_ context.Context, ch <-chan *message.Wri
 	return nil
 }
 
-// UnimplementedDeleteStale is a dummy handler to error on DeleteStale messages
+// UnimplementedDeleteStale is a dummy handler to consume and error on DeleteStale messages
 type UnimplementedDeleteStale struct{}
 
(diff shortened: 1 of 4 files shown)


writers/streamingbatchwriter/unimplemented.go
```diff
@@ -18,9 +18,12 @@ func (IgnoreMigrateTable) MigrateTable(_ context.Context, ch <-chan *message.Wri
...
 
-// UnimplementedDeleteStale is a dummy handler to error on DeleteStale messages
+// UnimplementedDeleteStale is a dummy handler to consume and error on DeleteStale messages
 type UnimplementedDeleteStale struct{}
 
-func (UnimplementedDeleteStale) DeleteStale(_ context.Context, _ <-chan *message.WriteDeleteStale) error {
+func (UnimplementedDeleteStale) DeleteStale(_ context.Context, ch <-chan *message.WriteDeleteStale) error {
+	// nolint:revive
+	for range ch {
+	}
```
bbernays, line 27: Fix as part of #1131

PR #1141, opened 2023-08-10
feat: Added a CQ ID Salt to the CQ ID generation to support overlapping sync

yvardhineni:
## ISSUE
https://github.com/cloudquery/cloudquery/issues/14590
Right now the behaviour of overlapping syncs of same resource is undefined as they always refer to the same primary keys while writing

## PROPOSAL
Accept a custom hash salt as a source spec which can be used as a custom salt while generating deterministic cq_ids.
This way we can get different cq_ids across different syncs even for the same resource.

This holds true when we use **deterministic_cq_id: true && pk_mode: _cq_id_only**

## TODO
Need to update source spec and other parts of Plugin SDKs to support this

### DIFF
--- a/scheduler/scheduler.go
+++ b/scheduler/scheduler.go
@@ -140,6 +140,12 @@ func WithStrategy(strategy Strategy) Option {
 
 type SyncOption func(*syncClient)
 
+func WithSyncCQIDSalt(salt string) SyncOption {
+	return func(s *syncClient) {
+		s.cqIDSalt = salt
+	}
+}
+
 func WithSyncDeterministicCQID(deterministicCQID bool) SyncOption {
 	return func(s *syncClient) {
 		s.deterministicCQID = deterministicCQID
@@ -168,6 +174,7 @@ type syncClient struct {
 	client            schema.ClientMeta
 	scheduler         *Scheduler
 	deterministicCQID bool
+	cqIDSalt          string
 	// status sync metrics
 	metrics *Metrics
 	logger  zerolog.Logger
--- a/scheduler/scheduler_dfs.go
+++ b/scheduler/scheduler_dfs.go
@@ -148,7 +148,7 @@ func (s *syncClient) resolveResourcesDfs(ctx context.Context, table *schema.Tabl
 					return
 				}
 
-				if err := resolvedResource.CalculateCQID(s.deterministicCQID); err != nil {
+				if err := resolvedResource.CalculateCQID(s.deterministicCQID, s.cqIDSalt); err != nil {
 					tableMetrics := s.metrics.TableClient[table.Name][client.ID()]
 					s.logger.Error().Err(err).Str("table", table.Name).Str("client", client.ID()).Msg("resource resolver finished with primary key calculation error")
 					if _, found := sentValidationErrors.LoadOrStore(table.Name, struct{}{}); !found {
--- a/scheduler/scheduler_test.go
+++ b/scheduler/scheduler_test.go
@@ -132,6 +132,7 @@ type syncTestCase struct {
 	table             *schema.Table
 	data              []scalar.Vector
 	deterministicCQID bool
+	cqidsalt          string
 }
 
 var syncTestCases = []syncTestCase{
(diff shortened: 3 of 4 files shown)


erezrokah (commented):
Thanks for the PR @yvardhineni.
Do you mind sharing example config files with overlapping resources, and example queries to extract the data once it's synced with different CQId values?

I understand this makes the behavior of overlapping syncs consistent, but I'm not sure how one can use the data without dedicated queries for the overlapping resources (those will get duplicated right?)

yvardhineni:
```yaml
kind: source
spec:
  # Source spec section
  name: "azure-<salt-1>"
  path: "cloudquery/azure"
  version: "v9.3.0"
  destinations: ["postgresql"]
  tables: ["azure_compute_virtual_machines"]
  deterministic_id: true
  cq_id_salt: salt-1 #proposed
  spec:
    # Optional parameters
   subscriptions: [1,2,3]
--
kind: source
spec:
  # Source spec section
  name: "azure-<salt-2>"
  path: "cloudquery/azure"
  version: "v9.3.0"
  destinations: ["postgresql"]
  deterministic_id: true
  cq_id_salt: salt-1 #proposed
  tables: ["azure_compute_virtual_machines"]
  spec:
    # Optional parameters
    subscriptions: [2]
```
Here we can see that subscription id is an overlapping resource across configs, here the duplication of data is expected as that can be one of the use case of User and it will be consistent across parallel runs too as we are taking [pk_mode: cq_id_only](https://www.cloudquery.io/docs/reference/destination-spec#pk_mode )

and also to note it won't alter the existing behaviour 

>  how one can use the data without dedicated queries for the overlapping resources (those will get duplicated correct?)

We can make the queries scoped to **_cq_source_name** as it is different for the different syncs as defined [here](https://www.cloudquery.io/docs/advanced-topics/running-cloudquery-in-parallel#unique-names)

this way we can bring the consistency for [overlapping syncs](https://www.cloudquery.io/docs/advanced-topics/running-cloudquery-in-parallel#no-overlapping-syncs)

erezrokah:
> this way we can bring the consistency for [overlapping syncs](https://www.cloudquery.io/docs/advanced-topics/running-cloudquery-in-parallel#no-overlapping-syncs)

Won't it be better in this case to drop the second source configuration? What's the reason to have it?

yvardhineni:
Agreed, but when we want to run as them as two separate sync jobs and when it runs concurrently they are overriding the source names which is leading to undesired behaviour.

yvardhineni:
This is highly inconsistent  when we have the granular permission on resources and those permissions are on overlapped resources.

For e.g. 
We have a set of **creds-1** which have the access to azure resources **R1, R2, R3**
Then we have a set of **creds-2** which have access to azure resources **R3,R4,R5**

When we run with these creds we will end up in inconsistency

erezrokah:
Ah OK I think I get it. So the use case for having overlapping syncs is having overlapping credentials?

yvardhineni:
Yeah, thats one such case

yvardhineni:
@erezrokah @yevgenypats could you please help with further steps on this

yevgenypats:
> @erezrokah @yevgenypats could you please help with further steps on this

Hi, @yvardhineni, thanks for this PR! This sprint we are working on other priorities at the moment (SDKs for Python/Javascript/Java), but we will review this as soon as we are available. What is the urgency of this? Is it for internal use or commercial offering? If this is urgent and part of commercial offering - take a look at our [support](https://www.cloudquery.io/pricing) so we can prioritize some of the PRs reviews especially like this one that needs additional care and time. Alternatively, you can maintain a private/public fork so you don't get blocked until we merge it in.

yvardhineni:
@yevgenypats Could you please update your ETA to look at this PR? Please let us know if there is any concerns with the current approach

erezrokah:
Sorry for the delayed response @yvardhineni, closing per my comment in https://github.com/cloudquery/cloudquery/tree/3af7f5ee72c1131bb4c8191498dfc50e32546568/plugins/transformer/basic.

Let's use the issue for a discussion if needed

PR #1142, opened 2023-08-10
feat!: Remove docs generation from SDK

hermanschaaf:
This removes docs generation from the SDK. This functionality moved to the CloudQuery CLI in https://github.com/cloudquery/cloudquery/pull/12758. 

Users should now use the `cloudquery tables` command to achieve similar functionality as before.

Note: I decided to remove the JSON generation in this PR as well. We may bring this back in the future, but it might be in a different shape then so I think we can leave that decision for later.


erezrokah (approved):

hermanschaaf:
I still think we should do this, but before merging we need to also make sure the scaffold Makefile is updated to not reference this command anymore.

erezrokah:
I'm going to close this as stale, opened https://github.com/cloudquery/cloudquery-issues/issues/2500 (internal issue) to re-visit when we have time

PR #1143, opened 2023-08-13
feat: Add publish command

yevgenypats:
As we are productionizing and rolling out our own registry we need to move from our format being documented in a yaml file inside goreleaser to being part of our code and part of the sdk.

Right now this just build the package without publishing as our backend is not ready but nailing down this will also help finalize the backend which Im working on as well.

The idea here is **not** to introduce a `package.json` file like  in npm and then create yet another language inside `json` or `yaml` config but rather the other way around and use all the information we already have from the plugin author and build a package with a manifest.json so everything is configurable in the code.

we will have two types of packages in our registry:
- `native` (for languages like Go/Rust/C)
- `docker` (for runtime language like .net, Java, Python)

For native this will looks the following:

```
ls -l ./dist
(base) yevgenyp@yevgenys-mbp aws % ls -l dist 
total 1142360
-rw-r--r--  1 yevgenyp  staff        447 Aug 13 20:08 plugin.json
-rw-r--r--  1 yevgenyp  staff  145537822 Aug 13 20:08 plugin_darwin_amd64.zip
-rw-r--r--  1 yevgenyp  staff  145537822 Aug 13 20:08 plugin_darwin_arm64.zip
-rw-r--r--  1 yevgenyp  staff  145537820 Aug 13 20:07 plugin_linux_amd64.zip
-rw-r--r--  1 yevgenyp  staff  145537824 Aug 13 20:07 plugin_windows_amd64.zip
-rw-r--r--  1 yevgenyp  staff    2726434 Aug 13 20:07 tables.json
(base) yevgenyp@yevgenys-mbp aws % 
```

manifest will look the following:

```
cat plugin.json
{
  "name": "aws",
  "version": "Development",
  "title": "aws",
  "short_description": "",
  "description": "",
  "categories": [],
  "protocols": [
    3
  ],
  "supported_targets": [
    {
      "os": "linux",
      "arch": "amd64"
    },
    {
      "os": "windows",
      "arch": "amd64"
    },
    {
      "os": "darwin",
      "arch": "amd64"
    },
    {
      "os": "darwin",
      "arch": "arm64"
    }
  ],
  "package_type": "native"
}
```

This will make sure the cloudquery package registry has all the information needed to publish it, display it on CloudQuery Hub and give it for client to download without running the plugins itself. Not running the plugin itself is important for the following reasons:
- Security: not running untrusted code in our environment.
- Developer experience: moving any issues closer to the developer and shortening the debugging/publishing loop. 

Once it gets a first review I'll also add tests.


plugin/plugin_publish.go
```diff
@@ -0,0 +1,175 @@
...
+
+func (p *Plugin) writeManifest(ctx context.Context, dir string) error {
+	manifest := Manifest{
+		Name:             p.Name(),
+		Version:          p.Version(),
+		Title:            p.title,
+		ShortDescription: p.shortDescription,
+		Description:      p.description,
+		Categories:       p.categories,
+		Protocols:        []int{3},
```
hermanschaaf, line 92: Small thing but it would be nice if we can rather import the supported protocols as a constant from the relevant package

serve/publish.go
```diff
@@ -0,0 +1,180 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to copy plugin file to zip archive: %w", err)
+	}
+	return nil
+}
+
+func (s *PluginServe) writeManifest(ctx context.Context, dir string) error {
+	manifest := Manifest{
+		Name:             s.plugin.Name(),
+		Version:          s.plugin.Version(),
```
erezrokah, line 123: I think we're missing the part of injecting the plugin version via Go Releaser here https://github.com/cloudquery/cloudquery/blob/db142bc867e631f00cb06f21a4a0566210fc7fd0/plugins/.goreleaser.yaml#L12

Not sure how this would work in this approach. We could configure release please maybe to update the version in the `go` code like we do with python, see https://github.com/cloudquery/cloudquery/pull/12869/files#diff-e10085e6cd58ab584a2a0fd98e5b4bcdda801327fe2f8e571ee049a3bceaf1e0L13

serve/publish.go
```diff
@@ -0,0 +1,180 @@
...
+			}
+			if err := os.MkdirAll(distPath, 0755); err != nil {
+				return err
+			}
+			if err := s.plugin.Init(cmd.Context(), nil, plugin.NewClientOptions{
+				NoConnection: true,
+			}); err != nil {
+				return err
+			}
+			if err := s.writeTablesJson(cmd.Context(), distPath); err != nil {
```
erezrokah, line 164: There's a bit of overlap here with the CLI tables command. Not sure what to do about it yet, just stating it.
Also this is for our internal usage so maybe better to have it separated.

I would create a `schema.json` for both the manifest and tables so we can validate those. We'll need to be aware that changing/adding things in the Column/Table structs can break the JSON

erezrokah (approved):
I think this PR has two features:

1. Adding a manifest file to describe a plugin
2. Adding a publish command for plugins to replace Go Releaser (depends on 1.)

Regarding the first feature, I think usually it's done the other way around, where you have a static manifest file, and the runtime reads properties from it (for example in Go we can embed a JSON file will all the properties).
Then during publishing we just copy the file.

Having a static file makes it easier to declare, modify (for example during release to inject a version) and validate (e.g. if we have a JSON schema of the manifest).

At the moment the plugin manifest properties are spread around the code, and you need to invoke publish to know what is the manifest.

Regarding 2. I think it's good to have plugins more "self contained", we should make sure we don't reimplement Go Releaser, for example at the moment we:

1. [Set some env variables](https://github.com/cloudquery/cloudquery/blob/db142bc867e631f00cb06f21a4a0566210fc7fd0/plugins/.goreleaser.yaml#L8)
2. [Pack policies with plugins](https://github.com/cloudquery/cloudquery/blob/db142bc867e631f00cb06f21a4a0566210fc7fd0/plugins/.goreleaser.template.yaml#L10)
3. Sign plugins with a [checksum](https://github.com/cloudquery/cloudquery/blob/db142bc867e631f00cb06f21a4a0566210fc7fd0/plugins/.goreleaser.template.yaml#L12)
4. Some plugins can only run `publish` in a specific environment https://github.com/cloudquery/cloudquery/blob/db142bc867e631f00cb06f21a4a0566210fc7fd0/.github/workflows/release_plugin_dest_duckdb.yml#L11

I would maybe split this PR into 2:
1. Create the designated manifest file format, embed it and read the properties from it
2. The publish command

With that being said since this is a new feature and doesn't breaking anything we can move forward and see how it work

yevgenypats:
> I think this PR has two features:
> 
> 1. Adding a manifest file to describe a plugin
> 2. Adding a publish command for plugins to replace Go Releaser (depends on 1.)
> 
> Regarding the first feature, I think usually it's done the other way around, where you have a static manifest file, and the runtime reads properties from it (for example in Go we can embed a JSON file will all the properties). Then during publishing we just copy the file.
> 
> Having a static file makes it easier to declare, modify (for example during release to inject a version) and validate (e.g. if we have a JSON schema of the manifest).
> 
> At the moment the plugin manifest properties are spread around the code, and you need to invoke publish to know what is the manifest.
> 
> Regarding 2. I think it's good to have plugins more "self contained", we should make sure we don't reimplement Go Releaser, for example at the moment we:
> 
> 1. [Set some env variables](https://github.com/cloudquery/cloudquery/blob/db142bc867e631f00cb06f21a4a0566210fc7fd0/plugins/.goreleaser.yaml#L8)
> 2. [Pack policies with plugins](https://github.com/cloudquery/cloudquery/blob/db142bc867e631f00cb06f21a4a0566210fc7fd0/plugins/.goreleaser.template.yaml#L10)
> 3. Sign plugins with a [checksum](https://github.com/cloudquery/cloudquery/blob/db142bc867e631f00cb06f21a4a0566210fc7fd0/plugins/.goreleaser.template.yaml#L12)
> 4. Some plugins can only run `publish` in a specific environment https://github.com/cloudquery/cloudquery/blob/db142bc867e631f00cb06f21a4a0566210fc7fd0/.github/workflows/release_plugin_dest_duckdb.yml#L11
> 
> I would maybe split this PR into 2:
> 
> 1. Create the designated manifest file format, embed it and read the properties from it
> 2. The publish command
> 
> With that being said since this is a new feature and doesn't breaking anything we can move forward and see how it work

Thanks for the review. Few notes on 1,2.

1) It's not always the otherway around - for example in Python you have `setup.py` and in `Go` you don't have such thing at all. It's only in `node.js/npm`. The only reason it's manifest.json is because a node package doesn't conform to any API. In our plugin side it is already conforming to an SDK in our control which can generate the manifest instead of a user having to look up our documentation of a `json` file. Also, there are things that only the SDK knows - for example which protocol version it supports so it will have to generate another `manifest2.json` instead of just having one. Having said that we do use the manifest file because the registry can't run any user defined code as it's a security issue.
2) goreleaser has different things and features it was build for - building any go application and publishing to variety of registries. we only need to build cloudquery plugin and publish them to cloudquery hub in one very specific way otherwise our package won't work.

erezrokah:
> 1. It's not always the otherway around - for example in Python you have `setup.py` and in `Go` you don't have such thing at all. It's only in `node.js/npm`. The only reason it's manifest.json is because a node package doesn't conform to any API.

Good point, maybe for other SDKs it would make sense to use a static file instead of code to define the plugin. Chrome extensions also use [a static file](https://developer.chrome.com/docs/extensions/mv3/getstarted/development-basics/#build), VS Code extensions too, they [(ab)use](https://code.visualstudio.com/api/working-with-extensions/publishing-extension) `package.json`.
Same with [Terraform](https://developer.hashicorp.com/terraform/registry/providers/publishing#terraform-registry-manifest-file), [Figma](https://www.figma.com/plugin-docs/manifest/) and [ChatGPT plugins](https://platform.openai.com/docs/plugins/getting-started).

> instead of a user having to look up our documentation of a json file

The user still needs to look up the docs to know which methods to implement. If we read the data from a file they would only need to create the file, instead of implementing the `Title, Description, etc.` methods (and that would be identical to plugins in different languages, like JS, Python, Java).

> Also, there are things that only the SDK knows - for example which protocol version it supports so it will have to generate another manifest2.json instead of just having one

Plugins can declare which SDK version they use (see Terraform and Figma).

> goreleaser has different things and features it was build for - building any go application and publishing to variety of registries. we only need to build cloudquery plugin and publish them to cloudquery hub in one very specific way otherwise our package won't work.

👍 Mostly pointing out that we would need the build command to support the current features we use. Also we would need to somehow handle the case it can only run in a specific environment gracefully.

hermanschaaf:
Looks like the tests are failing on MacOS

PR #1175, opened 2023-08-30
feat: Extensive testing for `delete-stale`

candiduslynx:


.github/workflows/unittest.yml
```diff
@@ -10,6 +10,7 @@ on:
 jobs:
   unitests:
     strategy:
+      fail-fast: false
```
candiduslynx, line 13:
```suggestion
```

hermanschaaf (approved):

PR #1195, opened 2023-09-04
fix: Use tables with primary key in `delete-stale` tests

candiduslynx:
Tested by:
* https://github.com/cloudquery/cloudquery/pull/13568
* https://github.com/cloudquery/cloudquery/pull/13600
BEGIN_COMMIT_OVERRIDE
fix: Use tables with primary key in `delete-stale` tests (#1195)
fix: Return a pointer to the actual table column instead of a copy in `schema.Table.Column` (#1195)
END_COMMIT_OVERRIDE


schema/table.go
```diff
@@ -504,9 +504,9 @@ func (t *Table) ValidateColumnNames() error {
 }
 
 func (t *Table) Column(name string) *Column {
-	for _, c := range t.Columns {
-		if c.Name == name {
-			return &c
+	for i := range t.Columns {
```
erezrokah, line 507: Can this cause unwanted mutations to the column? If the caller assumes it's a copy and mutates the column this might has unexpected results.
What's the reason this is needed?

erezrokah (changes requested):
The test change looks good, a question on the code change

schema/table.go
candiduslynx: We need it to add `pk` setting to the `id` column in the extensive delete-stale test.

IMO, if we have a pointer returned from the func (not the value) we intend the value to be mutable.

candiduslynx: However, I see your point, I'll revert the change & set the option in a different manner.
I think in the next SDK version we should make it `func (t *Table) Column(string) (Column, bool)` to distinguish the missing value separately, as the pointer seemed as an intention to  allow mutating the value.

candiduslynx: done in aa6662099adfdb5f255b5d0c5712b598d3f2b719

erezrokah: Yeah, mostly trying to be careful with the impact that is hard to predict even if the intention was to return a mutable column

erezrokah (approved):

PR #1203, opened 2023-09-05
fix: Relax plugin tables and columns validation

erezrokah:
#### Summary

Follow up to https://github.com/cloudquery/plugin-sdk/pull/1108. This PR relaxes the validation to check only for duplicates.
Some databases allow uppercase table tables and the validation in the SDK seemed to be PostgreSQL oriented.
At some point we might want to have destination specific validation/normalization.

Related to https://discord.com/channels/872925471417962546/873606591335759872/1148643989348679770

---

- [ ] Update or add tests 🧪


schema/table.go
```diff
@@ -407,14 +382,6 @@ func (t *Table) filterDfs(parentMatched bool, include, exclude func(*Table) bool
 	return nil
 }
 
-func (t *Table) ValidateName() error {
```
erezrokah, line 410: There's also a function to validate column names, but I kept it due to https://github.com/cloudquery/plugin-sdk/blob/3dbf2e978e2f650c281ccac342b931a33940c525/transformers/struct.go#L370

hermanschaaf (approved):
We should definitely remove this from plugin initialization, but maybe let's create a ticket to do this kind of validation as part of the source plugin test suite? These checks can then be opt-in by default.

disq (approved):

erezrokah:
> but maybe let's create a ticket to do this kind of validation as part of the source plugin test suite

Yeah, but should we test it on the source? I mean destinations have the restrictions on names, not sources, so how would the source know what to restrict?

erezrokah:
Opened https://github.com/cloudquery/plugin-sdk/issues/1207

PR #1211, opened 2023-09-12
feat: Update package command

hermanschaaf:
This updates the plugin `package` command (previously called `publish`) to build executables into a `dist` directory that will be consumed by the `cloudquery publish` command for publishing to the registry.

Usage example:

```
go run main.go package . v1.2.3
```


erezrokah (approved):
Looks good but doesn't the plugin already knows its own version? What's the reason to pass it via the command?

disq (approved):
LGTM

erezrokah:
We can embed the version in the code like we do in https://github.com/cloudquery/cloudquery/blob/c4de5dab89fcc2845e37b5772bc462c5e8edc7ce/plugins/source/typeform/plugin/plugin.py#L14

hermanschaaf:
>. We can embed the version in the code like we do in ...

@erezrokah That might work for us, but I think we should keep it tool-agnostic as we want it to work for community plugins as well, regardless of whether they use Release Please. We may need a solution like that for Python again, but I think it will be a find-and-replace in that case. Does that make sense?

erezrokah:
> Does that make sense?

💯

PR #1214, opened 2023-09-16
feat: Add support for jsonschema

yevgenypats:
I was a bit blocked on the backend so took a short detour on our SDK. This will be needed for our upcoming cloud version as we need to generate forms automatically for every plugin version.

The JSON schema will also go into the registry backend and we can also integrate it nicely with the hub - with something like https://codesandbox.io/s/material-ui-json-schema-viewer-2u6zq


yevgenypats:
Seems something is off with main tests as don't think this PR relates to the failed tests.

plugin/plugin.go
```diff
@@ -120,6 +140,17 @@ func (p *Plugin) Init(ctx context.Context, spec []byte, options NewClientOptions
 	}
 	defer p.mu.Unlock()
 	var err error
+
+	if p.schemaValidator != nil {
```
erezrokah, line 144: Do we want to validate the top level schema too via JSON schema? Otherwise the validation error messages might be confusing (some in the JSON schema validator format and others in the JSON decoder format). We'll probably need to do this in the CLI

erezrokah: Also maybe you can share how a validation error looks like from the console

hermanschaaf:
To fix the test error, I think you need to run `go mod tidy` in the example plugin directory: https://github.com/cloudquery/plugin-sdk/blob/main/examples/simple_plugin/go.mod

hermanschaaf:
For this to be useful in the cloud version, or if we wanted to generate docs from it like the `cloudquery tables` command, or do basic validation in the CLI before starting a sync, I would have expected this to be added to the gRPC protocol in plugin-pb? Is that still part of the plan?

candiduslynx (approved):

yevgenypats:
> For this to be useful in the cloud version, or if we wanted to generate docs from it like the `cloudquery tables` command, or do basic validation in the CLI before starting a sync, I would have expected this to be added to the gRPC protocol in plugin-pb? Is that still part of the plan?

I thought initially that we don't need to expose it via gRPC because we can upload it to the hub as json - this way it's very easy for the frontend to use this JSONschema .

erezrokah (approved):

PR #1216, opened 2023-09-18
feat: Add message to package command

hermanschaaf:
This adds a required "message" paramater to the package command, which then gets added to package.json. This is for summarizing the changes in a version.

At the moment this flag is required, but we could potentially loosen that restriction if we had a way to use git to track changes between commits.

### DIFF
--- a/serve/package.go
+++ b/serve/package.go
@@ -31,6 +31,7 @@ This creates a directory with the plugin binaries, package.json and documentatio
 type PackageJSON struct {
 	SchemaVersion    int                `json:"schema_version"`
 	Name             string             `json:"name"`
+	Message          string             `json:"message"`
 	Version          string             `json:"version"`
 	Protocols        []int              `json:"protocols"`
 	SupportedTargets []TargetBuild      `json:"supported_targets"`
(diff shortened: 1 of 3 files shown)


serve/package.go
```diff
@@ -206,7 +208,7 @@ func copyFile(src, dst string) error {
 
 func (s *PluginServe) newCmdPluginPackage() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "package <plugin_directory> <version>",
+		Use:   "package <plugin_directory> <version> -m <message>",
```
disq, line 211: Ideally arguments should come later than parameters... `package -m <message> <dir> <version>`

disq (approved):

serve/package.go
```diff
@@ -31,6 +31,7 @@ This creates a directory with the plugin binaries, package.json and documentatio
 type PackageJSON struct {
 	SchemaVersion    int                `json:"schema_version"`
 	Name             string             `json:"name"`
+	Message          string             `json:"message"`
```
erezrokah, line 34: Would this be the equivalent of a changelog? If so can we call it `Changelog`?

hermanschaaf: It is equivalent to a changelog entry, yes. It's currently also known as `message` in the API and DB layers, associated with a particular plugin version. https://github.com/cloudquery/cloud/blob/main/internal/servergen/openapi/components/schemas/PluginVersion.yaml#L26-L29

I chose `message` because of its analogue with a commit message. Do you think it's worth the effort of updating?

erezrokah: Ah if it's a single message no need to change it

PR #1217, opened 2023-09-18
feat: Add Checksums to package.json format

hermanschaaf:
Adds zip file checksums to the package.json format.

Includes some commits from https://github.com/cloudquery/plugin-sdk/pull/1216 to avoid merge conflicts later.

### DIFF
--- a/serve/package.go
+++ b/serve/package.go
@@ -4,6 +4,7 @@ import (
 	"archive/zip"
 	"bytes"
 	"context"
+	"crypto/sha256"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -39,9 +40,10 @@ type PackageJSON struct {
 }
 
 type TargetBuild struct {
-	OS   string `json:"os"`
-	Arch string `json:"arch"`
-	Path string `json:"path"`
+	OS       string `json:"os"`
+	Arch     string `json:"arch"`
+	Path     string `json:"path"`
+	Checksum string `json:"checksum"`
 }
 
 func (s *PluginServe) writeTablesJSON(ctx context.Context, dir string) error {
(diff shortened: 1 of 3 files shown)


erezrokah (approved):

serve/package.go
```diff
@@ -31,16 +32,18 @@ This creates a directory with the plugin binaries, package.json and documentatio
...
 }
 
 type TargetBuild struct {
-	OS   string `json:"os"`
-	Arch string `json:"arch"`
-	Path string `json:"path"`
+	OS       string `json:"os"`
+	Arch     string `json:"arch"`
+	Path     string `json:"path"`
+	Checksum string `json:"checksum"`
```
erezrokah, line 46: How are we going to use the `Checksum`? Shouldn't we also sign the checksum file

disq: @erezrokah they will be sent with HTTP PATCH after the upload.

disq (approved):

serve/package.go
erezrokah: > @erezrokah they will be sent with HTTP PATCH after the upload.

Thanks @disq, but what do we do with them? Do we validate it?

disq: @erezrokah We pass them on to the client and after download the client should validate them (soon)

hermanschaaf: > How are we going to use the Checksum?

We will store it in package.json here temporarily, and then read and upload it along with the binaries via the `cloudquery publish` command (which reads the package.json file). Then it'll be stored on the backend, and used to verify downloads of binaries from the CLI. 

> Shouldn't we also sign the checksum file

I think we can add this on the backend at a later stage. The only case I can think of is if an attacker were to do a MITM-type attack between us and the user downloading the plugin. To mitigate this I think the CLI will need to ship with our public key, to make sure the checksum file came from us.

erezrokah: Yeah, mostly mentioning since we don't do any kind of checksum validation at the moment so maybe we can drop it altogether

PR #1218, opened 2023-09-18
feat: Add shuffle scheduler

hermanschaaf:
Falling somewhere between DFS and Round Robin, the `shuffle` scheduler uses randomization with a fixed seed to spread the load across tables and clients more evenly.

### DIFF
--- a/scheduler/scheduler.go
+++ b/scheduler/scheduler.go
@@ -34,6 +34,7 @@ const (
 const (
 	StrategyDFS Strategy = iota
 	StrategyRoundRobin
+	StrategyShuffle
 )
 
 type Strategy int
@@ -84,10 +85,11 @@ func (s *Strategy) Validate() error {
 	return fmt.Errorf("unknown scheduler strategy: %d", s)
 }
 
-var AllStrategies = Strategies{StrategyDFS, StrategyRoundRobin}
+var AllStrategies = Strategies{StrategyDFS, StrategyRoundRobin, StrategyShuffle}
 var AllStrategyNames = [...]string{
 	StrategyDFS:        "dfs",
 	StrategyRoundRobin: "round-robin",
+	StrategyShuffle:    "shuffle",
 }
 
 func StrategyForName(s string) (Strategy, error) {
--- a/scheduler/scheduler.go
+++ b/scheduler/scheduler.go
@@ -252,6 +252,8 @@ func (s *Scheduler) Sync(ctx context.Context, client schema.ClientMeta, tables s
 			syncClient.syncDfs(ctx, resources)
 		case StrategyRoundRobin:
 			syncClient.syncRoundRobin(ctx, resources)
+		case StrategyShuffle:
+			syncClient.syncShuffle(ctx, resources)
 		default:
 			panic(fmt.Errorf("unknown scheduler %s", s.strategy.String()))
 		}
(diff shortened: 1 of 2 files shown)


scheduler/scheduler_shuffle.go
```diff
@@ -0,0 +1,64 @@
...
+			wg.Wait()
+			return
+		}
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			defer s.scheduler.tableSems[0].Release(1)
+			// not checking for error here as nothing much to do.
+			// the error is logged and this happens when context is cancelled
+			// Round Robin currently uses the DFS algorithm to resolve the tables, but this
```
candiduslynx, line 47: Not round robin.

candiduslynx (changes requested):

scheduler/scheduler.go
```diff
@@ -34,6 +34,7 @@ const (
 const (
 	StrategyDFS Strategy = iota
 	StrategyRoundRobin
+	StrategyShuffle
```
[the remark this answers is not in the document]
hermanschaaf: I don't see anything that needs to be specifically added for that, other than having to add it to `AllStrategyNames` and `AllStrategies` which I already did

scheduler/scheduler_shuffle.go
```diff
@@ -0,0 +1,64 @@
...
+		}()
+	}
+
+	// Wait for all the worker goroutines to finish
+	wg.Wait()
+}
+
+func shuffle(tableClients []tableClient) {
+	// use a fixed seed so that runs with the same tables and clients perform similarly across syncs
+	r := rand.New(rand.NewSource(99))
```
erezrokah, line 59: Doesn't this mean we always randomize it the same way 🙃 ? Ideally this would be configurable but maybe we can use something from the spec so if the initial shuffle is bad users can modify it. Maybe hash on table/client IDs or something

scheduler/scheduler_shuffle.go
```diff
@@ -0,0 +1,64 @@
...
+	for i := range tableClients {
```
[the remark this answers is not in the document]
hermanschaaf: Good call! Updated

scheduler/scheduler_shuffle.go
hermanschaaf: > Doesn't this mean we always randomize it the same way 

First the obligatory https://xkcd.com/221/

>  Maybe hash on table/client IDs or something

Yeah we can do that, so then changing the ordering of tables in your config will cause a different shuffle.

hermanschaaf: I don't think it was really necessary, the odds of getting a bad shuffle (at either end of the DFS->Round Robin spectrum) if there are enough table-clients is astronomically small, but in [27c4e22](https://github.com/cloudquery/plugin-sdk/pull/1218/commits/27c4e22fc4dd2db79b2e7ede35ecd1c135afe28c) I added the ability to control it a little bit by changing the ordering of the tables in the config.

scheduler/scheduler.go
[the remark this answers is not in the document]
candiduslynx: I must've missed it on mobile

candiduslynx (approved):

PR #1223, opened 2023-09-20
fix: Validate spec only when connection is established

candiduslynx:
Documentation generation will fail otherwise.
Follow-up for #1214

### DIFF
--- a/plugin/plugin.go
+++ b/plugin/plugin.go
@@ -141,7 +141,7 @@ func (p *Plugin) Init(ctx context.Context, spec []byte, options NewClientOptions
 	defer p.mu.Unlock()
 	var err error
 
-	if len(spec) > 0 && p.schemaValidator != nil {
+	if !options.NoConnection && p.schemaValidator != nil {
 		var v any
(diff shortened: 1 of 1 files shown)


plugin/plugin.go
```diff
@@ -141,7 +141,7 @@ func (p *Plugin) Init(ctx context.Context, spec []byte, options NewClientOptions
 	defer p.mu.Unlock()
 	var err error
 
-	if p.schemaValidator != nil {
+	if len(spec) > 0 && p.schemaValidator != nil {
```
hermanschaaf, line 144: Shouldn't we still validate the spec even if it is nil?

hermanschaaf: I think we may need to move the validation somewhere else, or pass in an option to skip validation, if we need to init a plugin without running validation

candiduslynx: I've moved the check to `NoConnection` option setting, as it's the sole case for our docs gen

hermanschaaf (approved):
We could probably do with a better name for `NoConnection` but this works

PR #1226, opened 2023-09-20
fix(scalar): Don't pass typed nils in list values

disq:
Fixes issues like https://github.com/cloudquery/cloudquery-issues/issues/595

### DIFF
--- a/scalar/list.go
+++ b/scalar/list.go
@@ -94,7 +94,7 @@ func (s *List) Set(val any) error {
 		for i := 0; i < length; i++ {
 			s.Value[i] = NewScalar(s.Type.(*arrow.ListType).Elem())
 			iVal := reflectedValue.Index(i)
-			if (iVal.Kind() == reflect.Ptr || iVal.Kind() == reflect.Map || iVal.Kind() == reflect.Slice) && iVal.IsNil() {
+			if isReflectValueNil(iVal) {
 				continue
 			}
 			if err := s.Value[i].Set(iVal.Interface()); err != nil {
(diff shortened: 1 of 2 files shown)


erezrokah (approved):

scalar/list.go
```diff
@@ -93,7 +93,11 @@ func (s *List) Set(val any) error {
 		s.Value = make(Vector, length)
 		for i := 0; i < length; i++ {
 			s.Value[i] = NewScalar(s.Type.(*arrow.ListType).Elem())
-			if err := s.Value[i].Set(reflectedValue.Index(i).Interface()); err != nil {
+			iVal := reflectedValue.Index(i)
+			if (iVal.Kind() == reflect.Ptr || iVal.Kind() == reflect.Map || iVal.Kind() == reflect.Slice) && iVal.IsNil() {
+				continue
+			}
```
candiduslynx, line 99:
```suggestion
			switch iVal.Kind() {
			case reflect.Pointer, reflect.UnsafePointer, reflect.Map, reflect.Slice, reflect.Interface:
				if iVal.IsNil() {
					continue
				}
			}
```

or, better yet, define a helper
```go
func isReflectValueNil(v reflect.Value) bool {
	switch iVal.Kind() {
	case reflect.Pointer,
		reflect.UnsafePointer,
		reflect.Map,
		reflect.Slice,
		reflect.Interface:
	default:
		return false
	}
	return v.IsNil()
}
```

candiduslynx (approved):

PR #1238, opened 2023-09-21
fix: Skip tables.json when packaging destinations

hermanschaaf:
This adds a new `type` (`source` or `destination`) argument to the `package` command, and switches the ordering of arguments bit (since we're making a breaking change to it anyway). The `type` is written to package.json as well, and if it is `destination`, we don't create a `tables.json`.

### DIFF
--- a/serve/package.go
+++ b/serve/package.go
@@ -32,6 +32,7 @@ This creates a directory with the plugin binaries, package.json and documentatio
 type PackageJSON struct {
 	SchemaVersion    int                `json:"schema_version"`
 	Name             string             `json:"name"`
+	Type             plugin.Type        `json:"type"`
 	Message          string             `json:"message"`
 	Version          string             `json:"version"`
 	Protocols        []int              `json:"protocols"`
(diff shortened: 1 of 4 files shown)


serve/package.go
```diff
@@ -304,8 +310,10 @@ func (s *PluginServe) newCmdPluginPackage() *cobra.Command {
 			}); err != nil {
 				return err
 			}
-			if err := s.writeTablesJSON(cmd.Context(), distPath); err != nil {
-				return err
+			if pluginType == "source" {
```
disq, line 313:
```suggestion
			if pluginType == plugin.TypeSource {
```

erezrokah (approved):

erezrokah:
This means the CLI should not fail if `tables.json` doesn't exist correct?

hermanschaaf:
> This means the CLI should not fail if tables.json doesn't exist correct?

Right, it should skip the tables upload step. I'll update the CLI next

PR #1240, opened 2023-09-21
fix: Rename plugin type -> kind for consistency with existing configs

hermanschaaf:
I realized existing plugin configs use the word `kind` (e.g. `kind: source`) and not `type`. Let's just stick to one word everywhere for consistency.


disq (approved):

disq:
Good catch!

erezrokah (approved):

PR #1243, opened 2023-09-22
feat: Provide User with actionable error message when no tables are configured for syncing

bbernays:
#### Summary

---

- [ ] Update or add tests 🧪

### DIFF
--- a/scheduler/scheduler.go
+++ b/scheduler/scheduler.go
@@ -31,6 +31,8 @@ const (
 	otelName               = "schedule"
 )
 
+var ErrNoTables = errors.New("no tables specified for syncing, review `tables` and `skip_tables` in the your config file to specify at least one table to sync")
+
 const (
 	StrategyDFS Strategy = iota
 	StrategyRoundRobin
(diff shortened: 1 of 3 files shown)


scheduler/scheduler.go
```diff
@@ -31,6 +31,8 @@ const (
 	otelName               = "schedule"
 )
 
+var ErrNoTables = errors.New("no tables specified for syncing, review `tables` and `skip_tables` in the your config file to specify at least one table to sync")
```
hermanschaaf, line 34:
```suggestion
var ErrNoTables = errors.New("no tables specified for syncing, review `tables` and `skip_tables` in your config and specify at least one table to sync")
```

schema/table.go
```diff
@@ -520,6 +520,9 @@ func (t *Table) TableNames() []string {
 }
 
 func (t *Table) Copy(parent *Table) *Table {
+	if t == nil {
+		return nil
+	}
```
hermanschaaf, line 525: Is this change for testing? Could we move the `if t == nil` logic to the test instead? I'm just wary that adding this here might hide future bugs

bbernays: Definitely, we can move it, but currently it panics if you try to copy a nil table...

hermanschaaf: > currently it panics if you try to copy a nil table...

Sounds reasonable to me, you shouldn't be trying to copy a nil table :)

Issue #1248, opened 2023-09-27
bug: Package command doesn't work for destinations


erezrokah:
### Are you positive this is an SDK bug?

- [X] I'm positive this is an SDK bug

### Current Behavior

`go run main.go package -m "test" destination v2.2.8 .` fails if you run it from https://github.com/cloudquery/cloudquery/tree/c7c9d74a2d8e0c3b075bcdba2e8bb6ab5cef70c9/plugins/destination/test

See https://github.com/cloudquery/cloudquery/actions/runs/6323548104/job/17171329300#step:6:78

### Expected Behavior

`package` command should finish successfully

### Steps To Reproduce

Clone the monorepo https://github.com/cloudquery/cloudquery, then

```bash
cd plugins/destination/test
mkdir -p stubdocs
touch stubdocs/README.md
go run main.go package --docs-dir stubdocs -m "test" destination v2.2.8 .
```

See failure

### Pull request (optional)

- [X] I can submit a pull request

PR #1249, opened 2023-09-27
fix(package): Don't init destinations during package

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/plugin-sdk/issues/1248

Another option that we could do (or do both) is handle `NoConnection` in all destinations.

---

- [ ] Update or add tests 🧪


hermanschaaf (approved):
Ok let's go with this for now, but I'd suggest adding opening an issue to support NoConnection in destinations

erezrokah:
Opened in https://github.com/cloudquery/cloudquery/issues/14059

PR #1254, opened 2023-09-29
feat: Add JSON schema to scheduler strategy

candiduslynx:
This allows using centralized JSON schema definition instead of tagging every `scheduler.Strategy` entry in plugin specs


internal/random/random.go
```diff
@@ -15,9 +15,9 @@ var letters = []rune("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ")
 
 func randSeq(n int) string {
 	b := make([]rune, n)
-	rand.Seed(time.Now().UnixNano())
+	rnd := rand.New(rand.NewSource(time.Now().UnixNano()))
```
candiduslynx, line 18: If this is merged after https://github.com/cloudquery/plugin-sdk/pull/1255 the diff goes away

hermanschaaf (approved):
Was a bit hard to see what changed due to the file contents moving, but seems like it's mainly the addition of `func (Strategy) JSONSchema() *jsonschema.Schema` and a test for it?

PR #1267, opened 2023-10-03
fix: Add `linux_arm64` to default build targets

disq:


hermanschaaf (approved):

erezrokah:
After this is released we might want to set the targets explicly in the plugins that use cross compile like:
https://github.com/cloudquery/cloudquery/blob/9eedc4cc6b44eb37d26c250fdefa51d2469b6558/.github/workflows/publish_plugin_to_hub_duckdb.yml
https://github.com/cloudquery/cloudquery/blob/9eedc4cc6b44eb37d26c250fdefa51d2469b6558/.github/workflows/publish_plugin_to_hub_snowflake.yml
https://github.com/cloudquery/cloudquery/blob/9eedc4cc6b44eb37d26c250fdefa51d2469b6558/.github/workflows/publish_plugin_to_hub_sqlite.yml

As `linux_arm64` is excluded in Go Releaser config for those, but the our own publishing workflow relies on it not being the default

PR #1292, opened 2023-10-16
feat: Add support for conditional static linking of C lib to builds

maaarcelino:
This adds support for conditionally linking C libs when building plugin executables. CGO_ENABLED plugins (sqlite and duckdb) fail to run on Alpine Linux distributions because of missing C libraries needed to execute the code. See https://github.com/cloudquery/cloudquery/issues/14465.

The fix is to include those missing libraries into the executable that is created by `go build`:

`CGO_ENABLED=1 go build -ldflags="-linkmode external -extldflags=-static"`

### DIFF
--- a/serve/package.go
+++ b/serve/package.go
@@ -110,7 +110,7 @@ func (s *PluginServe) build(pluginDirectory, goos, goarch, distPath, pluginVersi
 		return nil, err
 	}
 	ldFlags := fmt.Sprintf("-s -w -X %s/plugin.Version=%s", importPath, pluginVersion)
-	if getEnvOrDefault("STATIC_LINKING_ENABLED", "0") == "1" {
+	if s.plugin.IsStaticLinkingEnabled() {
 		ldFlags += " -linkmode external -extldflags=-static"
 	}
 	args := []string{"build", "-o", pluginPath}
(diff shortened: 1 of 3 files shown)


serve/package.go
```diff
@@ -105,13 +105,17 @@ func (s *PluginServe) writeTablesJSON(ctx context.Context, dir string) error {
 func (s *PluginServe) build(pluginDirectory, goos, goarch, distPath, pluginVersion string) (*TargetBuild, error) {
 	pluginName := fmt.Sprintf("plugin-%s-%s-%s-%s", s.plugin.Name(), pluginVersion, goos, goarch)
 	pluginPath := path.Join(distPath, pluginName)
-	args := []string{"build", "-o", pluginPath}
 	importPath, err := s.getModuleName(pluginDirectory)
 	if err != nil {
 		return nil, err
 	}
+	ldFlags := fmt.Sprintf("-s -w -X %s/plugin.Version=%s", importPath, pluginVersion)
+	if getEnvOrDefault("STATIC_LINKING_ENABLED", "0") == "1" {
```
erezrokah, line 113: Can we make this into a plugin option? See example https://github.com/cloudquery/cloudquery/blob/9e916090240a33ddf5a0163b950bf8fc1a8873ed/plugins/destination/sqlite/main.go#L23

Not sure if we want `withStaticLinking` bool option or `withAdditionalLDFlags` `[]string` flag. Maybe `withStaticLinking` is good for now

erezrokah (changes requested):
The fix looks good, can we use a plugin option instead of an env variable?

erezrokah (approved):
Thanks for the follow up 🚀

PR #1303, opened 2023-10-17
feat: Add JSON schema for `configtype.Duration`

candiduslynx:


configtype/duration.go
```diff
@@ -19,6 +21,15 @@ func NewDuration(d time.Duration) Duration {
 	}
 }
 
+func (Duration) JSONSchema() *jsonschema.Schema {
+	// provide a pattern for only non-negative durations
```
hermanschaaf, line 25: The regex doesn't look like it only supports non-negative durations?

candiduslynx: yeah, I'll remove the comment (as we do support negative durations even in tests)

candiduslynx: removed in df0b27e

configtype/duration_test.go
```diff
@@ -52,3 +56,74 @@ func TestComparability(t *testing.T) {
...
+			Err:  true,
+			Spec: `false`,
+		},
+		{
+			Name: "bad format",
+			Err:  true,
+			Spec: `false`,
+		},
+	},
+		func() []testCase {
```
hermanschaaf, line 96: nit: this syntax of appending a function that calls itself and appends the results in-line feels unnecessarily complicated, it'd be more readable to introduce a variable first and append that, like:

```
generatedCases := make([]testcase, 20)
for i := 0; i < 20; i++ {
     generatedCases[i] = generateTestCase()
}

...
cases := append([]testCase{
		{
			Name: "empty",
			Err:  true,
			Spec: `""`,
		},
                ...,
                generatedCases...
}
```

or have a few manually created test cases, there doesn't seem to be enough complexity here to warrant generating random cases

hermanschaaf (approved):

PR #1313, opened 2023-10-18
feat: Support publishing plugins with team and kind metadata set

hermanschaaf:
This allows us to package the plugin with team and kind metadata set, needed for publishing to the new registry and checking usage and limits against the new cloud API.

The package command no longer needs to take a `kind` argument.

Also adds an `info` command to print this information, useful for debugging, otherwise it won't be visible anywhere. We can consider adding this to the `--version` output instead, but that would technically be breaking?


examples/simple_plugin/plugin/plugin.go
```diff
@@ -4,14 +4,20 @@ import (
 	"github.com/cloudquery/plugin-sdk/v4/plugin"
 )
 
+// These variables are set as part of the `package` command
```
mnorbury, line 7: Is this comment correct now?

hermanschaaf: Oh no, thanks

mnorbury (approved):

disq (approved):

PR #1319, opened 2023-10-19
feat: Adding `IsPaid` fields to `Table` and `SyncMessage`

mnorbury:
Adding an `IsPaid` field to both the `Table` and `SyncMessage` structs to allow tables and event based syncs to charge for sync'd resources.

fixes: https://github.com/cloudquery/cloudquery-issues/issues/689


schema/table.go
```diff
@@ -76,6 +76,9 @@ type Table struct {
 	// documented.
 	IsIncremental bool `json:"is_incremental"`
 
+	// IsPaid is a flag that indicates if the table is a paid resource or not.
+	IsPaid bool `json:"-"`
```
mnorbury, line 80: This field will be used by the plugin before sending the sync message, so doesn't need to be passed as metadata.

hermanschaaf: I think we do want it to be part of the JSON output, so should include ``json:"is_paid"`` here

For the `cloudquery tables` command to use it it will also need to be part of the Arrow metadata, but this is not too important right now

mnorbury: Good point, modified in https://github.com/cloudquery/plugin-sdk/pull/1319/commits/ca9cedf2a4fe2adf189af1551b9a6c7cd744ea23 and tested locally using `cloudquery cli` - which needs changes in a follow-up PR

mnorbury:
We are currently pursuing sync.Message before send hooks, which would replace this approach.

PR #1321, opened 2023-10-19
fix(package): Only return one level down of relations when writing `tables.json`

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery-issues/issues/690 (internal issue).

`Tables.TableNames()` is recursive and gets all nested names so we can't use it.

Without the fix each table will have all of descendants under `relations` instead of only the children (1 level descendants)

---

- [ ] Update or add tests 🧪


internal/memdb/memdb.go
```diff
@@ -165,12 +181,12 @@ func (c *client) Sync(_ context.Context, options plugin.SyncOptions, res chan<-
...
 }
 
-func (c *client) Tables(context.Context, plugin.TableOptions) (schema.Tables, error) {
+func (c *client) Tables(ctx context.Context, opts plugin.TableOptions) (schema.Tables, error) {
 	tables := make(schema.Tables, 0, len(c.tables))
 	for _, table := range c.tables {
 		tables = append(tables, table)
 	}
-	return tables, nil
+	return tables.FilterDfs(opts.Tables, opts.SkipTables, opts.SkipDependentTables)
```
erezrokah, line 189: This is needed so we set parents implicitly via `Copy` (see also https://github.com/cloudquery/cloudquery/issues/14637)

disq (approved):

hermanschaaf:
@erezrokah There's a test failure FYI (expected 2 tables but got 3)

PR #1325, opened 2023-10-20
feat: Add OnBeforeSend hook

hermanschaaf:
This adds an optional `OnBeforeSend` hook to plugin clients that choose to implement the interface. When implemented, the plugin client will receive a call before every message is sent, allowing it to modify the message, perform a side-effect, or return an error.

This keeps the potentially general useful for all plugins, with the first use case being to check whether a row should be counted as premium or not in upcoming paid plugins.


internal/servers/plugin/v3/plugin.go
```diff
@@ -130,8 +130,13 @@ func (s *Server) Sync(req *pb.Sync_Request, stream pb.Plugin_SyncServer) error {
 			syncErr = fmt.Errorf("failed to sync records: %w", err)
 		}
 	}()
-
+	var err error
 	for msg := range msgs {
+		msg, err = s.Plugin.OnBeforeSend(ctx, msg)
+		if err != nil {
+			syncErr = fmt.Errorf("failed before sending message: %w", err)
```
yevgenypats, line 137: I can't recall now but why we `break` here and not `return` like in all other places below

yevgenypats (approved):
Looks good. had one question ^

internal/servers/plugin/v3/plugin.go
hermanschaaf: The reason I did it this way was that I didn't want it to be treated as an internal error like in the other `return` cases, but I think it's all the same really

hermanschaaf: I think I'll change it to return just to be clear

PR #1326, opened 2023-10-20
feat: Adding a batch updater to allow usage updates to be batched

mnorbury:
This will allows a client to update the usage asynchronously. The updater will only call the API when a configurable number of rows have been updated or a timeout is reached. In addition the updater will flush any remaining rows before closing.

If an error is encountered then an exponential backoff up to a max number of retries or a maximum wait time (relative to the start time of the last query) will be followed.

In addition, if the server replies with a status of `429` and includes a `Retry-After` header, then the client will wait that number of seconds before retrying.

It might be worth considering just using a retry library also e.g. https://github.com/avast/retry-go.

### DIFF
--- a/premium/usage.go
+++ b/premium/usage.go
@@ -227,7 +227,7 @@ func (u *BatchUpdater) updateUsageWithRetryAndBackoff(ctx context.Context, numbe
 		if err != nil {
 			return fmt.Errorf("failed to update usage: %w", err)
 		}
-		if resp.StatusCode() == http.StatusOK {
+		if resp.StatusCode() >= 200 && resp.StatusCode() < 300 {
 			u.lastUpdateTime = time.Now().UTC()
 			return nil
 		}
--- a/premium/usage.go
+++ b/premium/usage.go
@@ -7,7 +7,6 @@ import (
 	"github.com/google/uuid"
 	"github.com/rs/zerolog/log"
 	"net/http"
-	"sync"
 	"sync/atomic"
 	"time"
 )
@@ -73,7 +72,7 @@ type BatchUpdater struct {
 	rowsToUpdate   atomic.Uint32
 	triggerUpdate  chan struct{}
 	done           chan struct{}
-	wg             *sync.WaitGroup
+	closeError     chan error
 	isClosed       bool
 }
 
@@ -92,7 +91,7 @@ func NewUsageClient(ctx context.Context, apiClient *cqapi.ClientWithResponses, t
 		maxWaitTime:    defaultMaxWaitTime,
 		triggerUpdate:  make(chan struct{}),
 		done:           make(chan struct{}),
-		wg:             &sync.WaitGroup{},
+		closeError:     make(chan error),
 	}
 	for _, op := range ops {
 		op(u)
@@ -139,20 +138,17 @@ func (u *BatchUpdater) Close(_ context.Context) error {
 	u.isClosed = true
 
 	close(u.done)
-	u.wg.Wait()
 
-	return nil
+	return <-u.closeError
 }
 
 func (u *BatchUpdater) backgroundUpdater(ctx context.Context) {
 	started := make(chan struct{})
-	u.wg.Add(1)
 
 	duration := time.Duration(u.tickerDuration) * time.Millisecond
 	ticker := time.NewTicker(duration)
 
 	go func() {
-		defer u.wg.Done()
 		started <- struct{}{}
 		for {
 			select {
@@ -176,7 +172,7 @@ func (u *BatchUpdater) backgroundUpdater(ctx context.Context) {
 				}
 				if err := u.updateUsageWithRetryAndBackoff(ctx, rowsToUpdate); err != nil {
 					log.Error().Err(err).Msg("failed to update usage")
-					// TODO: what to do with an update error
+					// TODO: what to do with a timer error
 					continue
 				}
 				u.rowsToUpdate.Add(-rowsToUpdate)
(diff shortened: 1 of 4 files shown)


premium/usage.go
```diff
@@ -0,0 +1,240 @@
...
+
+// WithBatchLimit sets the maximum number of rows to update in a single request
+func WithBatchLimit(batchLimit uint32) UpdaterOptions {
+	return func(updater *BatchUpdater) {
+		updater.batchLimit = batchLimit
+	}
+}
+
+// WithTickerDuration sets the duration between updates if the number of rows to update have not reached the batch limit
+func WithTickerDuration(durationms int) UpdaterOptions {
```
hermanschaaf, line 40: Maybe we can make this `WithFlushEvery(d time.Duration)` ?

premium/usage.go
```diff
@@ -0,0 +1,240 @@
...
+	"github.com/google/uuid"
+	"github.com/rs/zerolog/log"
+	"net/http"
+	"sync/atomic"
+	"time"
+)
+
+const (
+	defaultBatchLimit  = 1000
+	defaultDurationMS  = 10000
```
hermanschaaf, line 16: Let's use a `time.Duration` here

premium/usage.go
```diff
@@ -0,0 +1,240 @@
...
+	cqapi "github.com/cloudquery/cloudquery-api-go"
+	"github.com/google/uuid"
+	"github.com/rs/zerolog/log"
+	"net/http"
+	"sync/atomic"
+	"time"
+)
+
+const (
+	defaultBatchLimit  = 1000
```
hermanschaaf, line 15: This might be a bit tricky to implement, but I think we may also want to have a minimum time between flushes, probably around 10 or 20 seconds. This way we don't risk sending a thousand requests in all quick succession if we're able to sync 1M rows in a short period.

premium/usage.go
```diff
@@ -0,0 +1,264 @@
...
+	"sync/atomic"
+	"time"
+)
+
+const (
+	defaultBatchLimit            = 1000
+	defaultMaxRetries            = 5
+	defaultMaxWaitTime           = 60 * time.Second
+	defaultMinimumUpdateDuration = 10 * time.Second
+	defaultFlushDuration         = 30 * time.Second
```
hermanschaaf, line 19:
```suggestion
	defaultMaxTimeBetweenFlushes         = 30 * time.Second
```

(nit)

premium/usage.go
```diff
@@ -0,0 +1,264 @@
...
+	"net/http"
+	"sync/atomic"
+	"time"
+)
+
+const (
+	defaultBatchLimit            = 1000
+	defaultMaxRetries            = 5
+	defaultMaxWaitTime           = 60 * time.Second
+	defaultMinimumUpdateDuration = 10 * time.Second
```
hermanschaaf, line 18:
```suggestion
	defaultMinTimeBetweenFlushes = 10 * time.Second
```

(nit)

And we'd update the `With...` functions accordingly

premium/usage.go
mnorbury: Added a comparison to the last update time here 59e4ab28f23cc93ab2acc49bdb2024d3c969dab2, which will prevent updates happening quicker than a minimumUpdateDuration (default 10s).

premium/usage.go
```diff
@@ -0,0 +1,264 @@
...
+	if statusCode == http.StatusTooManyRequests {
+		retryAfter := headers.Get("Retry-After")
+		if retryAfter != "" {
+			retryDelay, err := time.ParseDuration(retryAfter + "s")
+			if err != nil {
+				return 0, fmt.Errorf("failed to parse retry-after header: %w", err)
+			}
+			if retryDelay > u.maxWaitTime {
+				return 0, fmt.Errorf("retry-after header exceeds max wait time: %s > %s", retryDelay, u.maxWaitTime)
+			}
```
hermanschaaf, line 257: I'm not sure this should be an error. Maybe we should respect whatever delay the server asks us to add?

premium/usage.go
```diff
@@ -0,0 +1,264 @@
...
+			RequestId:  uuid.New(),
+			PluginTeam: u.pluginTeam,
+			PluginKind: cqapi.PluginKind(u.pluginKind),
+			PluginName: u.pluginName,
+			Rows:       int(numberToUpdate),
+		})
+		if err != nil {
+			return fmt.Errorf("failed to update usage: %w", err)
+		}
+		if resp.StatusCode() == http.StatusOK {
```
hermanschaaf, line 230: I think this will be `201` or `204` going forward (I'm changing it in an upcoming PR). Maybe just accept any 2xx code to be safe.

hermanschaaf (commented):
Looks great! Left a couple of comments, nothing major

premium/usage.go
```diff
@@ -0,0 +1,264 @@
...
+		if retryDuration > 0 {
+			time.Sleep(retryDuration)
+		}
+	}
+	return fmt.Errorf("failed to update usage: max retries exceeded")
+}
+
+// calculateRetryDuration calculates the duration to sleep relative to the query start time before retrying an update
+func (u *BatchUpdater) calculateRetryDuration(statusCode int, headers http.Header, queryStartTime time.Time, retry int) (time.Duration, error) {
+	if statusCode == http.StatusTooManyRequests {
```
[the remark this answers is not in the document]
mnorbury: We should add `503` as a possible return [here](https://github.com/cloudquery/cloud/blob/main/internal/servergen/openapi/paths/teams_%7Bteam_name%7D_usage.yaml)

hermanschaaf (approved):

PR #1327, opened 2023-10-20
feat: Add IsPaid flag to table definition

hermanschaaf:
This adds an `IsPaid` flag to the table definition. This can be used by premium plugins to indicate that a certain table is paid.


yevgenypats (changes requested):
It looks good - one thing that Im afraid with tags is that if we use it for things like isPaid we loose the whole Golang/gRPC static/api checks. I feel like tags are just for things like docs and metadata and not for things that are part of a protocol essentially.

hermanschaaf:
@yevgenypats The reason to have tags as part of the Arrow metadata, even isPaid, is essentially only for docs. Other than for docs generation, the CLI won't use this. There aren't plans for destinations to use this either.

However we do plan to use it to check whether we should check the quota for a given table before proceeding with the sync. The `premium` package will define the name of the tag to set and then use that to check. I think that's okay, WDYT?

yevgenypats (approved):

PR #1330, opened 2023-10-25
feat: Adding premium plugin support for updating usage and monitoring quota

mnorbury:
- Fixing API to match the example usage
- Add a quota monitoring function

### DIFF
--- a/premium/monitor.go
+++ b/premium/monitor.go
@@ -1,0 +1,80 @@
+package premium
+
+import (
+	"context"
+	"errors"
+	"time"
+)
+
+var ErrNoQuota = errors.New("no remaining quota for the month, please increase your usage limit if you want to continue syncing this plugin")
+
+const DefaultQuotaCheckInterval = 30 * time.Second
+
+type quotaChecker struct {
+	qm       QuotaMonitor
+	duration time.Duration
+}
+
+type QuotaCheckOption func(*quotaChecker)
+
+// WithQuotaCheckPeriod the time interval between quota checks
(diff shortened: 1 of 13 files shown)


premium/monitor.go
```diff
@@ -0,0 +1,80 @@
...
+
+	if !hasQuota {
+		return ErrNoQuota
+	}
+
+	return nil
+}
+
+func (qc quotaChecker) startQuotaMonitor(ctx context.Context) (context.Context, func()) {
+	newCtx, cancel := context.WithCancel(ctx)
```
mnorbury, line 60: Typically a cancelable context is created by the caller, but in this case I didn't want to make the caller (plugin developer) have to write this code.

premium/monitor.go
```diff
@@ -0,0 +1,80 @@
...
+	go func() {
+		defer cancel()
+		ticker := time.NewTicker(qc.duration)
+		for {
+			select {
+			case <-newCtx.Done():
+				return
+			case <-ticker.C:
+				hasQuota, err := qc.qm.HasQuota(newCtx)
```
mnorbury, line 69: Need to add retries to the `HasQuota` call.

premium/monitor.go
```diff
@@ -0,0 +1,80 @@
...
+				if err != nil {
+					continue
```
mnorbury, line 71: What should we do with an error determining the quota - since returning an error here would cancel the sync.

mnorbury:
Split into #1333 and #1334

PR #1333, opened 2023-10-27
feat: Adding quota monitoring for premium plugins

mnorbury:
This adds a background monitoring process that will periodically check the remaining quota for a premium plugin. If the quota is exceeded then a context cancellation is triggered, forcing the sync process to stop.

fixes: https://github.com/cloudquery/cloudquery-issues/issues/749


premium/monitor.go
```diff
@@ -0,0 +1,80 @@
...
+		op(&m)
+	}
+
+	if err := m.checkInitialQuota(ctx); err != nil {
+		return ctx, nil, err
+	}
+
+	ctx, cancel := m.startQuotaMonitor(ctx)
+
+	return ctx, cancel, nil
```
hermanschaaf, line 43: It seems to me like we don't need to return `cancel` as part of the function signature here, as the cancellation is handled internally

premium/monitor.go
```diff
@@ -0,0 +1,80 @@
...
+		ticker := time.NewTicker(qc.duration)
+		for {
+			select {
+			case <-newCtx.Done():
+				return
+			case <-ticker.C:
+				hasQuota, err := qc.qm.HasQuota(newCtx)
+				if err != nil {
+					continue
+				}
```
hermanschaaf, line 72: I think we need to set some limit on how long the sync can continue without knowing the remaining quota.

hermanschaaf (approved):

PR #1334, opened 2023-10-27
feat: Allow sync to be cancelled when in progress

mnorbury:
In some cases, when the resource channel is processing a large collection of objects, the sync process can continue even if the context cancel has been triggered. This is problematic for billing and usage since a sync can significantly overshoot the remaining rows quota. To allow an in-progress sync to abort quicker, we need to modify a number of channel patterns from using `range` to using `select` - which will allow the `context.Done` channel to be monitored.

This PR also adds the `OnSyncFinisher` hook which is called when the sync has finished adding messages. This allows the client code to call the batch updater close method without missing any updates.

fixes: https://github.com/cloudquery/cloudquery-issues/issues/750


scheduler/scheduler.go
```diff
@@ -182,15 +183,23 @@ func (s *Scheduler) Sync(ctx context.Context, client schema.ClientMeta, tables s
...
 	for resource := range resources {
-		vector := resource.GetValues()
-		bldr := array.NewRecordBuilder(memory.DefaultAllocator, resource.Table.ToArrowSchema())
-		scalar.AppendToRecordBuilder(bldr, vector)
-		rec := bldr.NewRecord()
-		res <- &message.SyncInsert{Record: rec}
+		select {
+		case res <- &message.SyncInsert{Record: resourceToRecord(resource)}:
+		case <-ctx.Done():
+			return nil
```
disq, line 189: why not return ctx.Err()?

mnorbury: No good reason, I'll look at doing that 🙏

disq (approved):

hermanschaaf (approved):
LGTM

PR #1359, opened 2023-11-09
feat(package): Check for Version variable

disq:
Also adds another ldflag path (`resources/plugin`) to set `Version` which is what we already have in github releases.

### DIFF
--- a/serve/package.go
+++ b/serve/package.go
@@ -2,10 +2,12 @@ package serve
 
 import (
 	"archive/zip"
+	"bufio"
 	"bytes"
 	"context"
 	"crypto/sha256"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -111,7 +113,7 @@ func (s *PluginServe) build(pluginDirectory, goos, goarch, distPath, pluginVersi
 	if err != nil {
 		return nil, err
 	}
-	ldFlags := fmt.Sprintf("-s -w -X %s/plugin.Version=%s", importPath, pluginVersion)
+	ldFlags := fmt.Sprintf("-s -w -X %s/plugin.Version=%s -X %s/resources/plugin.Version=%s", importPath, pluginVersion, importPath, pluginVersion)
 	if s.plugin.IsStaticLinkingEnabled() && strings.EqualFold(goos, plugin.GoOSLinux) {
 		ldFlags += " -linkmode external -extldflags=-static"
 	}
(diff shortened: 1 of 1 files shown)


serve/package.go
```diff
@@ -262,6 +264,64 @@ func (*PluginServe) copyDocs(distPath, docsPath string) error {
...
+			if !strings.HasSuffix(strings.ToLower(d.Name()), ".go") {
+				return nil
+			}
+
+			f, err := os.Open(path)
+			if err != nil {
+				return err
+			}
+			defer f.Close()
+			if ok, err := fileContains(f, "Version"); err != nil {
```
disq, line 308: could also have a regex that requires whitespace around `Version`...

serve/package.go
```diff
@@ -111,7 +113,7 @@ func (s *PluginServe) build(pluginDirectory, goos, goarch, distPath, pluginVersi
 	if err != nil {
 		return nil, err
 	}
-	ldFlags := fmt.Sprintf("-s -w -X %s/plugin.Version=%s", importPath, pluginVersion)
+	ldFlags := fmt.Sprintf("-s -w -X %s/plugin.Version=%s -X %s/resources/plugin.Version=%s", importPath, pluginVersion, importPath, pluginVersion)
```
hermanschaaf, line 116:
```suggestion
	ldFlags := fmt.Sprintf("-s -w -X %[1]s/plugin.Version=%[2]s -X %[1]s/resources/plugin.Version=%[2]s", importPath, pluginVersion)
```

(untested but should work)

serve/package.go
hermanschaaf: Or maybe a regex that checks for `Version\s*=`?

hermanschaaf (approved):

serve/package.go
disq: Will add regex

disq: Regex in https://github.com/cloudquery/plugin-sdk/pull/1359/commits/cdb8b66c842e44d72d723fe6a0b79511dc745858

https://regex101.com/r/7Hm4PT/1

serve/package.go
```diff
@@ -262,6 +264,66 @@ func (*PluginServe) copyDocs(distPath, docsPath string) error {
...
+		s, err := os.Stat(p)
+		if err == nil && s.IsDir() {
+			foundDirs = append(foundDirs, dir)
+		}
+	}
+	if len(foundDirs) == 0 {
+		return fmt.Errorf("plugin directory must contain at least one of the following directories: %s", strings.Join(checkRelativeDirs, ", "))
+	}
+
+	versionRegex := regexp.MustCompile(`^\s?Version\s*=`)
```
hermanschaaf, line 289:
```suggestion
	versionRegex := regexp.MustCompile(`Version\s*=`)
```

otherwise `var Version = "development"` would fail

disq: Fixed by adding `(var)?`

` var Version = ` wouldn't work (preceding space before `var` but I think that's fine, or I could add another `\s?`

https://regex101.com/r/7Hm4PT/3

disq: will add regex tests.

serve/package.go
```diff
@@ -262,6 +264,66 @@ func (*PluginServe) copyDocs(distPath, docsPath string) error {
...
+		s, err := os.Stat(p)
+		if err == nil && s.IsDir() {
+			foundDirs = append(foundDirs, dir)
+		}
+	}
+	if len(foundDirs) == 0 {
+		return fmt.Errorf("plugin directory must contain at least one of the following directories: %s", strings.Join(checkRelativeDirs, ", "))
+	}
+
+	versionRegex := regexp.MustCompile(`^(var)?\s?Version\s*=`)
```
hermanschaaf, line 289:
```suggestion
	versionRegex := regexp.MustCompile(`^\s*(var\s+)?Version\s*=`)
```

disq: do we really want to allow that though? Would lead to more false matches (inside a function... I know others also do)

hermanschaaf: oh yeah I guess it would be a false match inside a function. fair enough

hermanschaaf (approved):

hermanschaaf (approved):

yevgenypats (approved):

PR #1366, opened 2023-11-14
fix: Mark relations as paid as well

candiduslynx:


premium/tables.go
```diff
@@ -23,5 +23,6 @@ func MakeAllTablesPaid(tables schema.Tables) schema.Tables {
 // MakeTablePaid sets the table to paid
 func MakeTablePaid(table *schema.Table) *schema.Table {
 	table.IsPaid = true
+	MakeAllTablesPaid(table.Relations)
```
hermanschaaf, line 26: Not sure `MakeTablePaid` should also affect child tables, that's a bit surprising. Maybe only add this behavior to `MakeAllTablesPaid`?

candiduslynx: I think we should just document this (as the easiest way to make a single table paid without its children would be just to flip the flag `table.IsPaid = true`).

hermanschaaf: It's not necessary to change this function to fix the bug though?

hermanschaaf (approved):

PR #1372, opened 2023-11-16
fix: Retrieve team for api key

mnorbury:
Retreive the team associated with an API key to allow usage tracking against a team name

fixes: https://github.com/cloudquery/cloudquery-issues/issues/893

### DIFF
--- a/premium/usage.go
+++ b/premium/usage.go
@@ -377,7 +377,7 @@ func (u *BatchUpdater) getTeamNameByTokenType(tokenType auth.TokenType) (string,
 			return "", fmt.Errorf("failed to list teams for API key: %w", err)
 		}
 		if resp.StatusCode() != http.StatusOK {
-			return "", fmt.Errorf("failed to list teams for API key status code: %s", resp.Status())
+			return "", fmt.Errorf("failed to list teams for API key, status code: %s", resp.Status())
 		}
 		if len(resp.JSON200.Items) != 1 {
 			return "", fmt.Errorf("expected to find exactly one team for API key, found %d", len(resp.JSON200.Items))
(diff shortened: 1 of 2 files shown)


premium/usage.go
```diff
@@ -361,3 +357,33 @@ func (u *BatchUpdater) calculateRetryDuration(statusCode int, headers http.Heade
...
+			return "", fmt.Errorf("team name not set. Hint: use `cloudquery switch <team>`")
+		}
+		return teamName, nil
+	case auth.APIKey:
+		resp, err := u.apiClient.ListTeamsWithResponse(context.Background(), &cqapi.ListTeamsParams{})
+		if err != nil {
+			return "", fmt.Errorf("failed to list teams for API key: %w", err)
+		}
+		if resp.StatusCode() != http.StatusOK {
+			return "", fmt.Errorf("failed to list teams for API key status code: %s", resp.Status())
```
hermanschaaf, line 380:
```suggestion
			return "", fmt.Errorf("failed to list teams for API key, status code: %s", resp.Status())
```

hermanschaaf (approved):

premium/usage.go
```diff
@@ -361,3 +357,33 @@ func (u *BatchUpdater) calculateRetryDuration(statusCode int, headers http.Heade
...
+			return "", fmt.Errorf("config file for reading team name not found (%w). Hint: use `cloudquery login` and/or `cloudquery switch <team>`", err)
+		} else if err != nil {
+			return "", fmt.Errorf("failed to get team name from config: %w", err)
+		}
+		if teamName == "" {
+			return "", fmt.Errorf("team name not set. Hint: use `cloudquery switch <team>`")
+		}
+		return teamName, nil
+	case auth.APIKey:
+		resp, err := u.apiClient.ListTeamsWithResponse(context.Background(), &cqapi.ListTeamsParams{})
```
disq, line 375: The use of `context.Background()` isn't very nice here, but then it's in a constructor so the constructor would have to get a ctx otherwise.

mnorbury: Agreed, but considered it better than changing the constructor in this case.

disq (approved):

PR #1388, opened 2023-11-30
feat: Add NoOpUsageClient

disq:
To simplify usage usage.


erezrokah (commented):
What's the use case for this? Where should we use it?

disq:
@erezrokah This would shorten/simplify all the OnBeforeSend/OnSyncFinish logic, but performance wise I'm not sure if it's a right step for larger plugins.

hermanschaaf:
Sorry, first time I'm seeing this, but I also don't understand the use case 🤔

PR #1403, opened 2023-12-05
feat: Add `GetPaidTables()` and `HasPaidTables()` to `schema.Tables`

disq:


schema/table.go
```diff
@@ -366,6 +366,21 @@ func (tt Tables) ValidateDuplicateTables() error {
 	return nil
 }
 
+func (tt Tables) GetPaidTables() []string {
```
hermanschaaf, line 369: Maybe this should return `Tables` rather?

disq: yes in https://github.com/cloudquery/plugin-sdk/pull/1403/commits/ae9e8ff90eff7b0b6de2029c6ef89634cd9220a3

disq: Use `GetPaidTables().TableNames()` to get names if necessary.

hermanschaaf (approved):

erezrokah (approved):

PR #1410, opened 2023-12-10
feat: Add Scheduler benchmark back

bbernays:
#### Summary

---

- [ ] Update or add tests 🧪


hermanschaaf:
@bbernays Looks like unit tests are failing right now: 

```
2023/12/10 22:30:58 failed to write .delta.BenchmarkTablesWithGlobalRateLimiting/dfs-10000-8_successful-api-calls_per_s: open .delta.BenchmarkTablesWithGlobalRateLimiting/dfs-10000-8_successful-api-calls_per_s: no such file or directory
```

bbernays:
This is failing because of the benchmark takes too long to run... Need to figure out how to make it shorter and test be an effective benchmark

erezrokah:
> This is failing because of the benchmark takes too long to run... Need to figure out how to make it shorter and test be an effective benchmark

Can we run it on a schedule instead, then send a slack message on failure? Similar to https://github.com/cloudquery/cloudquery/blob/4328ead968995569dc0a4f955c22ed8382af8e2c/.github/workflows/broken_links_scheduled.yml

bbernays:
The benchmark is now running. The issue was that the new benchmarks I added have a `/` in them and that was introducing errors with the the processing of the benchmark data because it was expecting a file name not a path to a file.

That being said. I am not sure if the verbosity and the time it takes to run these benchmarks is worth adding to the repo. each of the metrics is important as it sheds light into the performance of schedulers in different scenarios.

I especially am weary of merging this as it run on *every* single pr not just prs that touch the scheduler

Thoughts @hermanschaaf @erezrokah

erezrokah:
> The benchmark is now running. The issue was that the new benchmarks I added have a `/` in them and that was introducing errors with the the processing of the benchmark data because it was expecting a file name not a path to a file.
> 
> That being said. I am not sure if the verbosity and the time it takes to run these benchmarks is worth adding to the repo. each of the metrics is important as it sheds light into the performance of schedulers in different scenarios.
> 
> I especially am weary of merging this as it run on _every_ single pr not just prs that touch the scheduler
> 
> Thoughts @hermanschaaf @erezrokah?

Yeah, I think we should revise how we report the performance measurements. Ideally they will fail a PR if there's a drop in performance but I don't think they're achieving it now.

hermanschaaf:
Yeah I think we don't have to run these on every PR. In fact, the benchmark reporter on PRs has probably outlived its usefuness, maybe we should retire it for now. We can try something else in the future if it becomes necessary. 

One idea would be a command like `/benchmarks` that we can run on a PR, to trigger a benchmark run. It could then run the benchmarks twice: once on `main` branch and once on the branch and compare results. We can then run this command on any relevant PRs.

Anyway, not for now. Maybe we can just remove the GItHub action in this PR?

erezrokah:
Closing as stale, please re-open if still relevant

PR #1411, opened 2023-12-10
feat: Individual Table and Client rate limit

bbernays:
#### Summary

This pr introduces a rate semaphore that limits the number of simultaneous go routines that can be resolving the same table + client to just 10. Prior to this a single table could be using all of the available concurrency to resolve a single table. This alleviates the problem where nested tables get throttled because the load is not spread across all client/table pairs but is concentrated on a single client and table.

I will update this PR with a benchmark case once #1410 gets merged

### DIFF
--- a/scheduler/scheduler.go
+++ b/scheduler/scheduler.go
@@ -67,9 +67,9 @@ func WithStrategy(strategy Strategy) Option {
 	}
 }
 
-func WithSingleTableMaxConcurrency(concurrency int) Option {
+func WithSingleTableMaxConcurrency(concurrency int64) Option {
 	return func(s *Scheduler) {
-		s.singleTableMaxConcurrency = int64(concurrency)
+		s.singleTableMaxConcurrency = concurrency
 	}
 }
 
--- a/scheduler/scheduler.go
+++ b/scheduler/scheduler.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"runtime/debug"
+	"sync"
 	"sync/atomic"
 	"time"
 
@@ -65,6 +66,12 @@ func WithStrategy(strategy Strategy) Option {
 	}
 }
 
+func WithSingleTableMaxConcurrency(concurrency int) Option {
+	return func(s *Scheduler) {
+		s.singleTableMaxConcurrency = concurrency
+	}
+}
+
 type SyncOption func(*syncClient)
 
 func WithSyncDeterministicCQID(deterministicCQID bool) SyncOption {
@@ -86,8 +93,10 @@ type Scheduler struct {
 	// tableSem is a semaphore that limits the number of concurrent tables being fetched
 	tableSems []*semaphore.Weighted
 	// Logger to call, this logger is passed to the serve.Serve Client, if not defined Serve will create one instead.
-	logger      zerolog.Logger
-	concurrency int
+	logger                    zerolog.Logger
+	concurrency               int
+	singleTableConcurrency    sync.Map
+	singleTableMaxConcurrency int
 }
 
 type syncClient struct {
--- a/scheduler/scheduler.go
+++ b/scheduler/scheduler.go
@@ -72,12 +72,6 @@ func WithSingleTableMaxConcurrency(concurrency int64) Option {
 	}
 }
 
-func WithGlobalRateLimiting() Option {
-	return func(s *Scheduler) {
-		s.globalRateLimiting = true
-	}
-}
-
 type SyncOption func(*syncClient)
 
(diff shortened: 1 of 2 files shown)


scheduler/scheduler.go
```diff
@@ -86,8 +94,10 @@ type Scheduler struct {
 	// tableSem is a semaphore that limits the number of concurrent tables being fetched
 	tableSems []*semaphore.Weighted
 	// Logger to call, this logger is passed to the serve.Serve Client, if not defined Serve will create one instead.
-	logger      zerolog.Logger
-	concurrency int
+	logger                    zerolog.Logger
+	concurrency               int
+	singleTableConcurrency    sync.Map
+	singleTableMaxConcurrency int
```
candiduslynx, line 100: just use int64 here?

scheduler/scheduler_dfs.go
```diff
@@ -198,10 +199,19 @@ func (s *syncClient) resolveResourcesDfs(ctx context.Context, table *schema.Tabl
 				wg.Wait()
 				return
 			}
+			tableSemVal, _ := s.scheduler.singleTableConcurrency.LoadOrStore(table.Name+"-"+client.ID(), semaphore.NewWeighted(int64(s.scheduler.singleTableMaxConcurrency)))
```
candiduslynx, line 202: It looks to me that we'll create a semaphore for each client/table pair, effectively doing nothing.

Shouldn't we instead use either only the table name or `s.scheduler.tableSems[depth]`?

scheduler/scheduler_dfs.go
```diff
@@ -198,10 +199,19 @@ func (s *syncClient) resolveResourcesDfs(ctx context.Context, table *schema.Tabl
 				wg.Wait()
 				return
 			}
+			tableSemVal, _ := s.scheduler.singleTableConcurrency.LoadOrStore(table.Name+"-"+client.ID(), semaphore.NewWeighted(s.scheduler.singleTableMaxConcurrency))
```
candiduslynx, line 202: maybe it should be performed before depth semaphore acquire?

bbernays: We can investigate and test that out in a followup PR once #1410 is merged.

erezrokah (commented):
Added a few comments/questions mostly. The blocking one is on making this a non breaking change

scheduler/scheduler_dfs.go
```diff
@@ -193,14 +194,30 @@ func (s *syncClient) resolveResourcesDfs(ctx context.Context, table *schema.Tabl
 		resolvedResources <- resource
 		for _, relation := range resource.Table.Relations {
 			relation := relation
+			// Depending on the source, rate limiting might be done on the table basis (GCP) or client + table (AWS)
```
erezrokah, line 197: Can you explain the reason to chose between the options? Why GCP is table only and AWS both table + client?

bbernays: GCP applies the rate limit to the Project where the callers identity resides. So if a single caller is calling the same API across 50 projects it all counts towards a single rate limit.

In AWS the rate limit applies to the account+region where the resource resides and doesn't matter about the callers identity

erezrokah: Maybe the API should be pass a function that generates the key with the default to use table name? I'm wondering what would be a good default here for other plugins

erezrokah (approved):

candiduslynx (approved):

scheduler/scheduler_dfs.go
```diff
@@ -193,14 +194,26 @@ func (s *syncClient) resolveResourcesDfs(ctx context.Context, table *schema.Tabl
 		resolvedResources <- resource
 		for _, relation := range resource.Table.Relations {
 			relation := relation
+			tableConcurrencyKey := table.Name + "-" + client.ID()
```
hermanschaaf, line 197: Unless I'm missing something, it seems like right now the `single table max concurrency` setting only applies to relational (ie. child) tables, and not top-level ones. As the name implies it would apply to all tables, I think we should either change the name or make it apply at the top-level as well.

My vote would go for making it apply to all tables. For top-level tables, it could still be useful. Imagine you have concurrency of 100, 10 tables and 100 clients. Without a per-table limit, the DFS scheduler would assign all 100 slots to the first table, which might not be desirable, so having a lever there would be useful.

bbernays: But at the top level aren't all top level tables a unique pair of `table.Name` and `client.ID`?

hermanschaaf (approved):

PR #1418, opened 2023-12-15
fix(tests): Find empty columns for JSON types

disq:
I'm not sure if there's a better way, but this seems to work for me.

~Still testing cases.~

Fixes https://github.com/cloudquery/cloudquery/issues/15707

### DIFF
--- a/schema/validators.go
+++ b/schema/validators.go
@@ -1,7 +1,10 @@
 package schema
 
 import (
+	"encoding/json"
+
 	"github.com/apache/arrow/go/v14/arrow"
+	"github.com/cloudquery/plugin-sdk/v4/types"
 )
 
 func FindEmptyColumns(table *Table, records []arrow.Record) []string {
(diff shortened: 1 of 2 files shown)


schema/validators.go
```diff
@@ -12,6 +15,14 @@ func FindEmptyColumns(table *Table, records []arrow.Record) []string {
 		for colIndex, arr := range resource.Columns() {
 			for i := 0; i < arr.Len(); i++ {
 				if arr.IsValid(i) {
+					if arrow.TypeEqual(arr.DataType(), types.ExtensionTypes.JSON) {
+						// JSON column shouldn't be empty
+						val := arr.GetOneForMarshal(i).(json.RawMessage)
+						if len(val) == 0 || string(val) == "null" || string(val) == "{}" || string(val) == "[]" {
```
candiduslynx, line 21: impl a separate `isEmptyJSON`:
```go
func isEmptyJSON(msg json.RawMessage) bool {
	if len(msg) == 0 {
		return true
	}
	switch string(msg) {
	case "null", "{}", "[]":
		return true
	default:
		return false
	}
}
```

disq: sure why not https://github.com/cloudquery/plugin-sdk/pull/1418/commits/f4c7cb6c4bdd0d4a42e18cd8f4907d4725ab85d7

candiduslynx (approved):
LGTM

erezrokah (approved):

PR #1433, opened 2023-12-27
feat: Resource rate limit

bbernays:
#### Summary

This introduces a rate limit on a resource + table + client basis. Prior to this we only had rate limits on a per table+ client basis and all resources shared the same concurrency bucket


scheduler/scheduler.go
```diff
@@ -132,6 +141,11 @@ func NewScheduler(opts ...Option) *Scheduler {
 	}
 	s.resourceSem = semaphore.NewWeighted(int64(resourceConcurrency))
 
+	// To preserve backwards compatibility, if singleResourceMaxConcurrency is not set, set it to the max concurrency
+	if s.singleResourceMaxConcurrency == 0 {
+		s.singleResourceMaxConcurrency = int64(resourceConcurrency)
+	}
```
hermanschaaf, line 147: I think we don't need to preserve backwards-compatibility for behavior that was not good before, let's rather pick a reasonable default

scheduler/scheduler.go
```diff
@@ -25,11 +25,13 @@ import (
 )
 
 const (
-	DefaultConcurrency     = 50000
-	DefaultMaxDepth        = 4
-	minTableConcurrency    = 1
-	minResourceConcurrency = 100
-	otelName               = "schedule"
+	singleResourceMaxConcurrency    = 5
+	singleNestedTableMaxConcurrency = 5
```
hermanschaaf, line 29: nit: these should be called `default...`

hermanschaaf (approved):

PR #1436, opened 2023-12-28
feat: Offline licensing support

disq:
BEGIN_COMMIT_OVERRIDE
feat: Offline licensing support

Includes breaking changes to the `premium.NewUsageClient` interface. Save value of `opts.PluginMeta` from `Configure()` and pass it to `premium.NewUsageClient` as the first argument.
END_COMMIT_OVERRIDE


erezrokah (approved):

premium/offline.go
```diff
@@ -0,0 +1,91 @@
...
+	l, err := UnpackLicense(licenseContents)
+	if err != nil {
+		return err
+	}
+
+	return l.IsValid(logger)
+}
+
+func UnpackLicense(lic []byte) (*License, error) {
+	const publicKey = "96e4749b7550d33bd776cb7fb056d74cb16ed69dfe8e59c16e8c2500a94162b1"
```
candiduslynx, line 47: move this to separate file & use go:embed

premium/usage.go
```diff
@@ -161,14 +153,14 @@ type BatchUpdater struct {
...
-func NewUsageClient(pluginTeam cqapi.PluginTeam, pluginKind cqapi.PluginKind, pluginName cqapi.PluginName, ops ...UsageClientOptions) (*BatchUpdater, error) {
+func NewUsageClient(meta plugin.Meta, ops ...UsageClientOptions) (UsageClient, error) {
 	u := &BatchUpdater{
 		logger: zerolog.Nop(),
 		url:    defaultAPIURL,
 
-		pluginTeam: pluginTeam,
-		pluginKind: pluginKind,
-		pluginName: pluginName,
+		pluginTeam: meta.Team,
```
candiduslynx, line 161: just use meta instead?

premium/usage.go
```diff
@@ -161,14 +153,14 @@ type BatchUpdater struct {
 	isClosed       bool
 }
 
-func NewUsageClient(pluginTeam cqapi.PluginTeam, pluginKind cqapi.PluginKind, pluginName cqapi.PluginName, ops ...UsageClientOptions) (*BatchUpdater, error) {
+func NewUsageClient(meta plugin.Meta, ops ...UsageClientOptions) (UsageClient, error) {
```
hermanschaaf, line 156: I agree this signature change is an improvement, but it's technically a breaking change, and we'll have to update all plugins using the usage package. Maybe we should do a v5 release for this?

disq: `NewUsageClient` is used only in CQ-authored, effectively private plugins.

hermanschaaf: That might be true but we have community plugins that are premium as well. And if they're not using it, they should be

premium/usage.go
disq: Done in [`27b0d28` (#1436)](https://github.com/cloudquery/plugin-sdk/pull/1436/commits/27b0d28d55e054d001f1e1332d150f5f6fec8e39)

premium/usage.go
disq: I don't think that warrants an upgrade of ~7000 lines in the public cq/cq repo.

hermanschaaf: Can't we just avoid this by keeping the signature the same? It looks like we can adopt `UsageClientOptions` for the additional option we need

hermanschaaf: Oh I guess the `*BatchUpdater` return might still be an issue. Well let's at least make a clear note in the changelog that this is a breaking change to the premium package interface, and explain what plugin authors need to do to handle it.

disq: There are ways to avoid this by shifting the burden to the plugin side, which is not favourable.

hermanschaaf: Yeah we don't need to go that far; let's just add an explanation to the changelog

disq: I've updated the PR desc. Adding a `BREAKING-CHANGE` there might bump the version, but further edits or suggestions are welcome.

hermanschaaf (approved):

premium/offline.go
disq: Done in [`07fd136` (#1436)](https://github.com/cloudquery/plugin-sdk/pull/1436/commits/07fd13685b0a0fd9ea82bd28ae93d69ea3d03c06)

yevgenypats (approved):

PR #1461, opened 2024-01-05
CQ ID Changes

bbernays:
#### Summary

These DBs support a `Unique` Clause and therefore need to identify how this change will impact them:
```
plugins/destination/duckdb
plugins/destination/duckdb
plugins/destination/mongodb
plugins/destination/mssql
plugins/destination/mssql
plugins/destination/mssql
plugins/destination/mssql
plugins/destination/mssql
plugins/destination/mysql
plugins/destination/postgresql
plugins/destination/snowflake
```

### DIFF
--- a/scheduler/scheduler.go
+++ b/scheduler/scheduler.go
@@ -189,6 +189,14 @@ func (s *Scheduler) Sync(ctx context.Context, client schema.ClientMeta, tables s
 
 	// send migrate messages first
 	for _, table := range tables.FlattenTables() {
+		if syncClient.deterministicCQID {
+			for i, c := range table.Columns {
+				if c.PrimaryKey && c.Name != schema.CqIDColumn.Name {
+					table.Columns[i].PrimaryKey = false
+				}
+			}
+		}
+
 		res <- &message.SyncMigrateTable{
 			Table: table,
 		}
--- a/schema/meta.go
+++ b/schema/meta.go
@@ -18,7 +18,7 @@ var CqIDColumn = Column{
 	Type:        types.ExtensionTypes.UUID,
 	Description: "Internal CQ ID of the row",
 	NotNull:     true,
-	Unique:      true,
+	Unique:      false,
 }
 
 var CqParentIDColumn = Column{
--- a/scheduler/scheduler_dfs.go
+++ b/scheduler/scheduler_dfs.go
@@ -208,7 +208,12 @@ func (s *syncClient) resolveResourcesDfs(ctx context.Context, table *schema.Tabl
 		resolvedResources <- resource
 		for _, relation := range resource.Table.Relations {
 			relation := relation
+
 			tableConcurrencyKey := table.Name + "-" + client.ID()
+			if table.TableConcurrencyKey != nil {
+				tableConcurrencyKey = *table.TableConcurrencyKey
+			}
+
 			// Acquire the semaphore for the table
 			tableSemVal, _ := s.scheduler.singleTableConcurrency.LoadOrStore(tableConcurrencyKey, semaphore.NewWeighted(s.scheduler.singleNestedTableMaxConcurrency))
 			tableSem := tableSemVal.(*semaphore.Weighted)


scheduler/scheduler.go
```diff
@@ -189,6 +189,14 @@ func (s *Scheduler) Sync(ctx context.Context, client schema.ClientMeta, tables s
 
 	// send migrate messages first
 	for _, table := range tables.FlattenTables() {
+		if syncClient.deterministicCQID {
+			for i, c := range table.Columns {
+				if c.PrimaryKey && c.Name != schema.CqIDColumn.Name {
+					table.Columns[i].PrimaryKey = false
+				}
+			}
+		}
```
bbernays, line 198: If Deterministic_CQ_ID is enabled, then that should be the only PK that is being used. Update the Table definition prior to sending the SyncMessage

schema/meta.go
```diff
@@ -18,7 +18,7 @@ var CqIDColumn = Column{
 	Type:        types.ExtensionTypes.UUID,
 	Description: "Internal CQ ID of the row",
 	NotNull:     true,
-	Unique:      true,
+	Unique:      false,
```
bbernays, line 21: Make CQ ID not have a Unique Key on it

scheduler/scheduler.go
```diff
@@ -189,6 +189,21 @@ func (s *Scheduler) Sync(ctx context.Context, client schema.ClientMeta, tables s
...
 	// send migrate messages first
 	for _, table := range tables.FlattenTables() {
+		if syncClient.deterministicCQID {
+			for i, c := range table.Columns {
+				if !c.PrimaryKey {
+					continue
+				}
+				if c.Name == schema.CqIDColumn.Name {
+					// CQ_ID should be PK and not have a Unique Clause on it
+					table.Columns[i].PrimaryKey = true
```
disq, line 199: This line is unnecessary

scheduler/scheduler.go
```diff
@@ -189,6 +189,21 @@ func (s *Scheduler) Sync(ctx context.Context, client schema.ClientMeta, tables s
...
+					table.Columns[i].Unique = false
```
disq, line 200: Modifying the result of `FlattenTables` won't make it true for the real tables list so they'll still 'see it' as `Unique: true`. I have no recommendation on how to solve this off the top of my head.

bbernays: But only the table definition in the migration event  actually changes the schema... This enable all of the docs to remain unchanged

disq: _but_ in the docs `cq_id` will show up as unique, which would be incorrect (since the source itself always specifies DetCQId:true now...)
Still I don't think this is an issue

bbernays: Just looked at the `plugin-sdk/v4/docs` package and we only show in our docs columns that are incremental or that have a PK so the `Unique` clause is never actually shown to the user unless they go to the schema of one of the 4 dbs that actually supports it

scheduler/scheduler.go
```diff
@@ -189,6 +189,25 @@ func (s *Scheduler) Sync(ctx context.Context, client schema.ClientMeta, tables s
 
 	// send migrate messages first
 	for _, table := range tables.FlattenTables() {
+		if syncClient.deterministicCQID {
+			// No PK adjustment should occur if `_cq_id` is not present in the table
+			cqIDCol := table.Columns.Get(schema.CqIDColumn.Name)
+			if cqIDCol == nil || !cqIDCol.PrimaryKey {
```
erezrokah, line 195: I'm missing something here about the `!cqIDCol.PrimaryKey` condition.
Shouldn't we set cqID to be the PK if it's present and not the PK? Then it will be calculated from the other PKs?

> This change updates the MigrateMessage so that only the _cq_id will be a PK and it will not have a Unique constraint on it

Basing on ⬆️ 

Is the purpose to do this here so we can remove the `pk_mode` on the destinations?

scheduler/scheduler.go
```diff
@@ -189,6 +189,25 @@ func (s *Scheduler) Sync(ctx context.Context, client schema.ClientMeta, tables s
...
+				continue
+			}
+			for i, c := range table.Columns {
+				if !c.PrimaryKey {
+					continue
+				}
+				if c.Name == schema.CqIDColumn.Name {
+					// If CQ_ID is a PK and it should not have a Unique Clause on it
+					table.Columns[i].Unique = false
```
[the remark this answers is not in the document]
bbernays: `_cq_id` has a unique constraint on it that hinders using deterministic cq id in append mode (where there are no pks).

scheduler/scheduler.go
bbernays: If a user specifically generates a `_cq_id` column that is not a PK, I don't think we should override it...

scheduler/scheduler.go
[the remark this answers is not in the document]
bbernays: We can do it there as well, but overall why do we want to include the CLI at all when we can do it all in the source and not include the CLI at all?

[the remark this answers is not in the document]
erezrokah: If unique columns are not supported for append mode, we should disable them only for append mode, same as we do for PKs. That way I don't think it will be a breaking schema change (or maybe only breaking if someone's using append mode)

scheduler/scheduler.go
erezrokah: > If a user specifically generates a `_cq_id` column that is not a PK

What do you mean by that? Aren't we the ones adding the `_cq_id` column

I'll clarify, what happens when a table has 3 PKs that neither are the CQ ID column? Shouldn't we make CQ ID the PK and remove the others?

hermanschaaf (approved):
I agree with Erez's comments; I think the code can be simplified a bit by assuming we always control `_cq_id` and it's a reserved column, but overall it seems like this should work and not cause issues for existing installations, while allowing us to move forward with the default deterministic CQ ID change.

scheduler/scheduler.go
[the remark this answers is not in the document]
bbernays: Discussed with @hermanschaaf... the `_cq_id` is a reserved column, if it is present it should be considered a PK and we should override it.

Changed in f50a58781e03b04f4c5fa0edaf587f3691cca936

scheduler/scheduler.go
bbernays: Discussed with @hermanschaaf... the `_cq_id` is a reserved column, if it is present it should be considered a PK and we should override it.

Changed in f50a58781e03b04f4c5fa0edaf587f3691cca936

scheduler/scheduler.go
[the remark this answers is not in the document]
bbernays: Unique constraints on the `_cq_id` do nothing for us in any mode other than adding the amount of data we write...PG destination also can't detect if a unique constraint has changed.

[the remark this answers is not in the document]
erezrokah: OK so looks like we don't consider `unique` here https://github.com/cloudquery/plugin-sdk/blob/a1857c340c77fc55c8e82a129cff81fa7757e5e1/schema/table.go#L458 when detecting changes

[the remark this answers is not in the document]
erezrokah: I still find it confusing that we set `Unique = false` and `PrimaryKey = true`.
It seems the only reason we do it is due to how we implemented append mode, by the CLI removing the primary key constraint and not the unique one. The latter seems like an oversight and we should remove unique from all columns, `_cq_id` or not, when in append mode, otherwise the sync will fail on duplicate values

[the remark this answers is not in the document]
erezrokah: Other than that I went over our destinations and setting `unique` to `false` seems safe when the column is a primary as DBs treat PKs as unique regardless

erezrokah (approved):

PR #1470, opened 2024-01-17
feat: Expose special migration paths

bbernays:
#### Summary

This pr allows destinations to handle some of the more safe but unique migrations. This will assist with the changes that we have plan with regards to _cq_id and uniqueness changes


schema/table.go
```diff
@@ -443,6 +457,15 @@ func (t *Table) ToArrowSchema() *arrow.Schema {
 // GetChanges returns changes between two tables when t is the new one and old is the old one.
 func (t *Table) GetChanges(old *Table) []TableColumnChange {
 	var changes []TableColumnChange
+	var movingToCQOnly bool
```
erezrokah, line 460: This change means users have to drop the table correct?
Maybe we can return immediately after it instead of returning all changes so destinations don't print non relevant ones here https://github.com/cloudquery/cloudquery/blob/6889eed19136fe8e63ed712212bb271964aaee24/plugins/destination/postgresql/client/migrate.go#L33

Also you might need to implement `String()` for these here https://github.com/cloudquery/plugin-sdk/blob/f583ceabc11cab9be2371b90ce1c7f44e17f8ca4/schema/table.go#L185

erezrokah: We can also return all changes and let the destination decide what to do with them, but then I'll remove `!movingToCQOnly` from https://github.com/cloudquery/plugin-sdk/pull/1470/files#diff-abecd8d935ca8952a3554e15fa6989972dd0cea6de3410b6541931883b0d6d84R482

bbernays: This change means that destinations can choose if they want to support this use case... Prior to this change, it was automatically a non-automigratable change.

bbernays: Removed `!movingToCQOnly` in 5094499f2e584d7ad83afc7687fdf4abd136ef21

PR #1476, opened 2024-01-22
fix: Proper `nil` check in `scalar`

candiduslynx:
Closes https://github.com/cloudquery/plugin-sdk/issues/1475
The `typed nil` also can be observed in https://github.com/cloudquery/cloudquery/pull/16221

### DIFF
--- a/scalar/interval.go
+++ b/scalar/interval.go
@@ -39,6 +39,12 @@ func (s *MonthInterval) String() string {
 }
 
 func (s *MonthInterval) Set(value any) error {
+	// this will check for typed nils as well, so no need to check below
+	if IsNil(value) {
+		s.Valid = false
+		return nil
+	}
+
 	if mi, ok := value.(arrow.MonthInterval); ok {
 		return s.Int.Set(int32(mi))
 	}
@@ -49,7 +55,7 @@ func (s *MonthInterval) Set(value any) error {
 			s.Valid = false
 			return nil
 		}
-		return s.Int.Set(value)
+		return s.Int.Set(v)
 	case []byte:
 		if len(v) == 0 {
 			s.Valid = false
@@ -64,15 +70,12 @@ func (s *MonthInterval) Set(value any) error {
 		s.Value = int64(mi.Months)
 		return nil
 	case *string:
-		if v == nil {
-			return s.Int.Set(nil)
-		}
-		return s.Set(value)
+		return s.Set(*v)
 	case map[string]any:
 		b, _ := json.Marshal(v)
 		return s.Set(b)
 	default:
-		return s.Int.Set(value)
+		return s.Int.Set(v)
 	}
 }
 
@@ -109,7 +112,8 @@ func (s *DayTimeInterval) Equal(rhs Scalar) bool {
 }
 
 func (s *DayTimeInterval) Set(value any) error {
-	if value == nil {
+	// this will check for typed nils as well, so no need to check below
+	if IsNil(value) {
 		s.Valid = false
 		return nil
 	}
@@ -148,10 +152,6 @@ func (s *DayTimeInterval) Set(value any) error {
 		s.Value = dti
 		return nil
 	case *string:
-		if v == nil {
-			s.Valid = false
-			return nil
-		}
 		return s.Set(*v)
 	case map[string]any:
 		b, _ := json.Marshal(v)
@@ -202,7 +202,8 @@ func (s *MonthDayNanoInterval) Equal(rhs Scalar) bool {
 }
 
 func (s *MonthDayNanoInterval) Set(value any) error {
-	if value == nil {
+	// this will check for typed nils as well, so no need to check below
+	if IsNil(value) {
 		s.Valid = false
 		return nil
 	}
@@ -241,10 +242,6 @@ func (s *MonthDayNanoInterval) Set(value any) error {
 		s.Value = dti
 		return nil
 	case *string:
-		if v == nil {
-			s.Valid = false
-			return nil
-		}
 		return s.Set(*v)
 	case map[string]any:
 		b, _ := json.Marshal(v)
--- a/scalar/nil_test.go
+++ b/scalar/nil_test.go
@@ -1,0 +1,18 @@
+package scalar
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestIsNil(t *testing.T) {
+	s := make([]*string, 1)
+	s[0] = nil
+	require.True(t, IsNil(s[0]))
+	require.False(t, isNilTrivial(s[0]))
+}
+
+func isNilTrivial(a any) bool {
+	return a == nil
+}
--- a/scalar/binary.go
+++ b/scalar/binary.go
@@ -38,7 +38,8 @@ func (s *Binary) Get() any {
 }
 
 func (s *Binary) Set(val any) error {
-	if val == nil {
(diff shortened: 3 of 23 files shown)


disq (approved):

hermanschaaf:
Has this changed anything, or is it a refactor? If it's not a refactor, can we add a test for the fix? (Other than testing `IsNil` directly)

scalar/nil_test.go
```diff
@@ -0,0 +1,18 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestIsNil(t *testing.T) {
+	s := make([]*string, 1)
+	s[0] = nil
+	require.True(t, IsNil(s[0]))
+	require.False(t, isNilTrivial(s[0]))
```
hermanschaaf, line 13: Seems like we could make this a table-driven test with some more types and assertions?

hermanschaaf: Also line 13 seems to be testing the test itself 🙃

candiduslynx: Added a test in 346bd33

candiduslynx: do we want to?

scalar/interval.go
```diff
@@ -64,15 +70,12 @@ func (s *MonthInterval) Set(value any) error {
 		s.Value = int64(mi.Months)
 		return nil
 	case *string:
-		if v == nil {
-			return s.Int.Set(nil)
```
erezrokah, line 68: Can we split this fix into a separate PR?

erezrokah (commented):
Added a comment on the interval fix, also won't this allows setting `nil` slices/maps/etc. to the bool type for example? Before it would return a validation error.

scalar/interval.go
candiduslynx: @erezrokah https://github.com/cloudquery/plugin-sdk/pull/1477

candiduslynx:
> Added a comment on the interval fix, also won't this allows setting `nil` slices/maps/etc. to the bool type for example? Before it would return a validation error.

This would allow setting `(map[string]string)(nil)` to the `scalar.Bool`, but I think we're better off allowing that than getting NPE.

erezrokah:
> Won't it make more sense to return a validation error? If you try to assign a map to a bool I'm not sure we should allow it

This https://github.com/cloudquery/plugin-sdk/blob/346bd334ee4af39544536ccdd9d136b26dbfee5b/scalar/bool.go#L65

candiduslynx:
Once again: if we go the "validation err" for such cases, we'll be once again susceptible to similar bugs. Especially, when we (or Apache Arrow) would introduce new type.

erezrokah:
> we'll be once again susceptible to similar bugs. Especially, when we (or Apache Arrow) would introduce new type.

Can't we `isReflectValueNil` and then return a validation error if the conversion doesn't make sense, for example in `bool, int, etc.` types?

candiduslynx:
> Can't we `isReflectValueNil` and then return a validation error if the conversion doesn't make sense, for example in `bool, int, etc.` types?

We absolutely can, but then again, it will be an overkill to do this separately for each type:
1. check for `nil`
2. check for typed nil + match for appropriate type

I can't see how this can be done in a shorter & idiomatic way when compared to the current codebase in `main` branch.

If you insist on erroring-out on type mismatch then we need to stick to the current pattern & just add `nil` checks to the proper  type `switch` branches in the places it applies to. I'm a bit opposed to it because of:
1. Larger codebase
2. Susceptibility to the same in the future

However, I can implement this as a side PR to compare the outcomes

erezrokah:
Looking at the stack trace from the issue https://github.com/cloudquery/plugin-sdk/issues/1475 looks like this was fixed in https://github.com/cloudquery/plugin-sdk/pull/1226 by not passing the typed nil value to the underlying non list type.

The error is from SDK version `v4.5.0` and the fix was released in `v4.10.1`. Maybe we can wait with this and see if it happens on new SDK versions?

candiduslynx:
> The error is from SDK version `v4.5.0` and the fix was released in `v4.10.1`. Maybe we can wait with this and see if it happens on new SDK versions?

It won't be reproduced by the new SDK (as the list will sanitize typed nils into untyped ones per https://github.com/cloudquery/plugin-sdk/pull/1226) as the issue was with `[]*string` slice passed with nil values & `Inet` not handling them properly.
However, the issue would be possible to reproduce via a `nil` field of `*net.HardrareAddr` type.

erezrokah:
> However, the issue would be possible to reproduce via a `nil` field of `*net.HardrareAddr` type.

Can you write a test that demonstrate this issue? Which scalars are relevant for this type? I don't think we want to try and translate `net.HardrareAddr` to bool

candiduslynx:
I've added a test line to inet test

erezrokah:
> I've added a test line to inet test

Got it, but it seems the issue is that `inet` is missing the handling for nil pointers in the respective case clauses, so https://github.com/cloudquery/plugin-sdk/compare/main...erezrokah:fix/inet_ptr?expand=1 should be a sufficient fix for that issue.

My point is that we probably want to handle `typed nil` in a specific way for each scalar. `inet` should not allow setting `type nil` of a type that is doesn't know how to set given a non nil value, it currently supports setting only from `net.IPNet, net.IP, string, TextMarshaler, Stringer`.

candiduslynx:
> My point is that we probably want to handle `typed nil` in a specific way for each scalar. `inet` should not allow setting `type nil` of a type that is doesn't know how to set given a non nil value, it currently supports setting only from `net.IPNet, net.IP, string, TextMarshaler, Stringer`.

Although it might be the case, we still allow for `scalar.Bool{Valid:false}` to successfully set `scalar.Int` to the `Valid:false` state. Either we need to rethink the scalar strategy in this place or we shouldn't bother a lot about typed nil of not-so-proper type (it should still result in error for the unsupported non-nil value, so that's basically covers the issue we're discussing ATM).

erezrokah:
> Although it might be the case, we still allow for `scalar.Bool{Valid:false}` to successfully set `scalar.Int` to the `Valid:false` state. Either we need to rethink the scalar strategy in this place or we shouldn't bother a lot about typed nil of not-so-proper type (it should still result in error for the unsupported non-nil value, so that's basically covers the issue we're discussing ATM).

OK so I think we're getting a bit out of scope of the PR.
I see two discussions:

1. Don't crash when setting a `typed nil` value for `inet` scalar (issue with the current code, not the issue linked in this PR). Here I would chose the solution that's most consistent with the current code
2. What do we consider as "valid" for scalar - I agree we are not consistent there is some places, but I feel choosing a strategy here should be discussed in an issue (and prioritized) before we change all scalars.

candiduslynx:
> 1. Don't crash when setting a `typed nil` value for `inet` scalar (issue with the current code, not the issue linked in this PR which was fixed already I think). Here I would chose the solution that's most consistent with the current code

Additionally, we have an issue when setting `scalar.Inet` with `fmt.Stringer` `nil` value (e.g., `nil` `*net.HardwareAddr` that isn't explicitly covered by the current code in `main` but will still get to the `String()` method call & NPE).

> 2. What do we consider as "valid" for scalar - I agree we are not consistent there is some places, but I feel choosing a strategy here should be discussed in an issue (and prioritized) before we change all scalars.

💯 agree

candiduslynx:
Overtaken by https://github.com/cloudquery/plugin-sdk/pull/1478

PR #1479, opened 2024-01-22
feat: Make UUID in `testdata` always deterministic like all other columns

bbernays:
#### Summary

`UUID` columns now follow the same pattern as all other columns

This could also be considered a `refactor`... I am open to changing it if people want

### DIFF
--- a/schema/testdata.go
+++ b/schema/testdata.go
@@ -1,6 +1,7 @@
 package schema
 
 import (
+	"crypto/sha256"
 	"encoding/base64"
 	"fmt"
 	"sort"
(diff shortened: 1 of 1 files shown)


schema/testdata.go
```diff
@@ -301,10 +300,10 @@ func (tg TestDataGenerator) getExampleJSON(colName string, dataType arrow.DataTy
 	}
 	// handle extension types
 	if arrow.TypeEqual(dataType, types.ExtensionTypes.UUID) {
-		u := uuid.New()
-		if opts.StableUUID != uuid.Nil {
-			u = opts.StableUUID
-		}
+		// This will make UUIDs deterministic like all other types
```
erezrokah, line 303: This looks good, but maybe we have some tests relying on the fact this is a random value?

erezrokah (approved):

hermanschaaf (approved):

PR #1480, opened 2024-01-22
feat: Expose Migration test for new special case for moving to `_cq_id` as only primary key

bbernays:
#### Summary

Destinations can now verify the new `MoveToCQIDPrimaryKey` change type that was introduced in #1470 

Depends on #1479

### DIFF
--- a/schema/testdata.go
+++ b/schema/testdata.go
@@ -3,7 +3,6 @@ package schema
 import (
 	"encoding/base64"
 	"fmt"
-	"sort"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


schema/testdata.go
```diff
@@ -252,16 +251,7 @@ func (tg *TestDataGenerator) Generate(table *Table, opts GenTestDataOptions) arr
...
 		bldr.Release()
 	}
-	if indices := sc.FieldIndices(CqIDColumn.Name); len(indices) > 0 {
-		cqIDIndex := indices[0]
-		sort.Slice(records, func(i, j int) bool {
-			firstUUID := records[i].Column(cqIDIndex).(*types.UUIDArray).Value(0).String()
-			secondUUID := records[j].Column(cqIDIndex).(*types.UUIDArray).Value(0).String()
-			return strings.Compare(firstUUID, secondUUID) < 0
-		})
-	}
```
bbernays, line 262: This code seems to be no longer used as we now use the `plugin.sortRecords` function

hermanschaaf (approved):
👍

erezrokah (approved):

PR #1481, opened 2024-01-23
chore: Remove Codecov Action

bbernays:
#### Summary

We do not use Codecov anymore, so no reason to run the action every commit

### DIFF
--- a/.github/workflows/unittest.yml
+++ b/.github/workflows/unittest.yml
@@ -44,9 +44,3 @@ jobs:
           path: |
             .delta.*
             pr_number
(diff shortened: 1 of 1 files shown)


erezrokah (approved):

erezrokah (changes requested):
I think we could simplify the workflow by removing the `matrix.os != 'large-ubuntu-plugin-sdk'` and `matrix.os == 'large-ubuntu-plugin-sdk'` conditions, WDYT?

erezrokah (approved):

PR #1489, opened 2024-01-25
fix: When `_cq_id` SyncMigrateMessage not sent

bbernays:
#### Summary

Fixed a case where no sync message would be sent

### DIFF
--- a/scheduler/scheduler.go
+++ b/scheduler/scheduler.go
@@ -190,10 +190,10 @@ func (s *Scheduler) Sync(ctx context.Context, client schema.ClientMeta, tables s
 	// send migrate messages first
 	for _, tableOriginal := range tables.FlattenTables() {
 		migrateMessage := &message.SyncMigrateTable{
-			Table: tableOriginal,
+			Table: tableOriginal.Copy(nil),
 		}
 		if syncClient.deterministicCQID {
-			migrateMessage.Table = schema.CqIDAsPK(tableOriginal.Copy(nil))
+			schema.CqIDAsPK(migrateMessage.Table)
 		}
 		res <- migrateMessage
 	}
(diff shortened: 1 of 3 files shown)


scheduler/scheduler.go
```diff
@@ -189,30 +189,13 @@ func (s *Scheduler) Sync(ctx context.Context, client schema.ClientMeta, tables s
...
 	// send migrate messages first
 	for _, tableOriginal := range tables.FlattenTables() {
-		// var table *schema.Table
-		table := tableOriginal.Copy(nil)
-		if syncClient.deterministicCQID {
-			// No PK adjustment should occur if `_cq_id` is not present in the table
-			cqIDCol := table.Columns.Get(schema.CqIDColumn.Name)
-			if cqIDCol == nil {
-				continue
-			}
```
bbernays, line 199: When this occurred, no syncMigrateMessage would be sent

scheduler/scheduler.go
```diff
@@ -189,14 +189,13 @@ func (s *Scheduler) Sync(ctx context.Context, client schema.ClientMeta, tables s
...
-		table := tableOriginal.Copy(nil)
-		if syncClient.deterministicCQID {
-			table = schema.CqIDAsPK(table)
+		migrateMessage := &message.SyncMigrateTable{
+			Table: tableOriginal,
 		}
-		res <- &message.SyncMigrateTable{
-			Table: table,
+		if syncClient.deterministicCQID {
+			migrateMessage.Table = schema.CqIDAsPK(tableOriginal.Copy(nil))
```
erezrokah:
```suggestion
			migrateMessage.Table = schema.CqIDAsPK(tableOriginal)
```

Shouldn't the copy inside `CqIDAsPK` be enough?

bbernays: I think you reviewed an older version of the PR, Can you look at the latest version and see if you think this comment still applies?

erezrokah (approved):

PR #1492, opened 2024-01-29
fix: Better build overrides

candiduslynx:
Instead of passing `static` option we should allow for env var values overrides in the package call.

### DIFF
--- a/plugin/plugin_package.go
+++ b/plugin/plugin_package.go
@@ -36,14 +36,21 @@ const (
 type BuildTarget struct {
 	OS   string   `json:"os"`
 	Arch string   `json:"arch"`
+	CGO  bool     `json:"cgo"`
 	Env  []string `json:"env"`
 }
 
 func (t BuildTarget) GetEnvVariables() []string {
+	cgo := "CGO_ENABLED="
+	if t.CGO {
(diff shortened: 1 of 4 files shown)


plugin/plugin_package.go
```diff
@@ -34,14 +34,23 @@ const (
...
 
 type BuildTarget struct {
-	OS   string `json:"os"`
-	Arch string `json:"arch"`
+	OS   string   `json:"os"`
+	Arch string   `json:"arch"`
+	Env  []string `json:"env"`
+}
+
+func (t BuildTarget) GetEnvVariables() []string {
```
disq, line 42:
```suggestion
func (t BuildTarget) EnvVariables() []string {
```
Removing `Get` would be more Go-idiomatic

candiduslynx: a6d6867d8e184927b2aa49feb335e2f4db29657e

disq (approved):
How about adding `Tags` to `BuildTarget`?

candiduslynx:
> How about adding `Tags` to `BuildTarget`?

I think we'll add them when we'll really need them 😃

erezrokah (approved):

PR #1497, opened 2024-01-31
fix: Remove access to parent tests in test suite

disq:
Saving `*testing.T` inside the `Suite` to use it later in sub-tests results in this error if we have a failed test:

> testing.go:1490: test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test

We shouldn't get this error.

Other parts of the test suite return `error` (signature is `func (s *WriterTestSuite) testName(context.Context) error`) and we do a simple `t.Fatal` in the main test body, but for DeleteStale tests we use `testify/require` inside the test so we need to access `*testing.T`.

### DIFF
--- a/plugin/testing_write.go
+++ b/plugin/testing_write.go
@@ -143,10 +143,10 @@ func TestWriterSuiteRunner(t *testing.T, p *Plugin, tests WriterTestSuiteTests, 
 			t.Skip("skipping " + t.Name())
 		}
 		t.Run("Basic", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


hermanschaaf (approved):

candiduslynx (commented):
Do we have an option to switch to a proper test suites?

disq:
@candiduslynx wdym?

candiduslynx:
> @candiduslynx wdym?

Utilize https://pkg.go.dev/github.com/stretchr/testify/suite

PR #1530, opened 2024-02-20
fix: Support list scalars from JSON

disq:
Without this, lists from the test source (at least, could be more plugins/cases) get persisted as `[]`

I'm still not 100% sure how the lists end up getting encoded to JSON, probably we use Arrow's `.ValueStr` and/or `.GetOneForMarshal` somewhere.

Called from table resolver - source plugins need this fix.

### DIFF
--- a/scalar/list.go
+++ b/scalar/list.go
@@ -1,6 +1,7 @@
 package scalar
 
 import (
+	"encoding/json"
 	"reflect"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


candiduslynx:
Could you also verify that the same fix is present in other SDKs?

disq:
> Could you also verify that the same fix is present in other SDKs?

Opened issues, linked above (Java SDK doesn't support lists at all so added a general issue for that too)

erezrokah (approved):

PR #1560, opened 2024-03-14
feat: Enable destinations to completely skip migration that are not supported

bbernays:
#### Summary

Some destinations don't support all migrations types. This allows them to opt out

### DIFF
--- a/plugin/testing_write.go
+++ b/plugin/testing_write.go
@@ -48,6 +48,17 @@ type SafeMigrations struct {
 	MovePKToCQOnly         bool
 }
 
+// Migrations defines which migrations should be skipped completely
+type Migrations struct {
+	AddColumn              bool
+	AddColumnNotNull       bool
+	RemoveColumn           bool
+	RemoveColumnNotNull    bool
+	ChangeColumn           bool
+	RemoveUniqueConstraint bool
+	MovePKToCQOnly         bool
+}
+
 type WriterTestSuiteTests struct {
 	// SkipUpsert skips testing with message.Insert and Upsert=true.
 	// Usually when a destination is not supporting primary keys
@@ -68,6 +79,8 @@ type WriterTestSuiteTests struct {
 	// SafeMigrations defines which tests should work with force migration
 	// and which should pass with safe migration
 	SafeMigrations SafeMigrations
+
+	SkipSpecificMigrations Migrations
 }
 
 type NewPluginFunc func() *Plugin
--- a/plugin/testing_write.go
+++ b/plugin/testing_write.go
@@ -50,11 +50,6 @@ type SafeMigrations struct {
 
 // Migrations defines which migrations should be skipped completely
 type Migrations struct {
-	AddColumn              bool
-	AddColumnNotNull       bool
-	RemoveColumn           bool
-	RemoveColumnNotNull    bool
-	ChangeColumn           bool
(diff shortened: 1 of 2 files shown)


plugin/testing_write_migrate.go
```diff
@@ -149,6 +152,9 @@ func (s *WriterTestSuite) testMigrate(
 	})
 
 	t.Run("add_column_not_null"+suffix, func(t *testing.T) {
+		if s.tests.SkipSpecificMigrations.AddColumn {
```
erezrokah, line 155:
```suggestion
		if s.tests.SkipSpecificMigrations.AddColumnNotNull {
```

plugin/testing_write_migrate.go
```diff
@@ -200,6 +209,9 @@ func (s *WriterTestSuite) testMigrate(
 	})
 
 	t.Run("remove_column_not_null"+suffix, func(t *testing.T) {
+		if s.tests.SkipSpecificMigrations.RemoveColumn {
```
erezrokah, line 212:
```suggestion
		if s.tests.SkipSpecificMigrations.RemoveColumnNotNull {
```

plugin/testing_write.go
```diff
@@ -48,6 +48,17 @@ type SafeMigrations struct {
 	MovePKToCQOnly         bool
 }
 
+// Migrations defines which migrations should be skipped completely
+type Migrations struct {
+	AddColumn              bool
+	AddColumnNotNull       bool
+	RemoveColumn           bool
+	RemoveColumnNotNull    bool
+	ChangeColumn           bool
```
erezrokah, line 57:
```suggestion
```

Aren't all migrations support these? Maybe only allow the new ones?

bbernays: My reasoning was that new destinations might not support those migrations and so might as well expose them for future usage.

erezrokah: I think the baseline should be to support these basic changes in all destinations. I'd only open this for configuration if we run into a destination that really needs them

bbernays: fixed in https://github.com/cloudquery/plugin-sdk/pull/1560/commits/b774c094ad9d394062bdffca8afec5dc7d623cf0

plugin/testing_write_migrate.go
```diff
@@ -270,13 +270,16 @@ func (s *WriterTestSuite) testMigrate(
 				{Name: "uuid", Type: types.ExtensionTypes.UUID},
 				{Name: "bool", Type: arrow.FixedWidthTypes.Boolean, NotNull: true},
 			}}
-		require.NoError(t, s.migrate(ctx, target, source, s.tests.SafeMigrations.RemoveUniqueConstraint, forceMigrate))
+		require.NoError(t, s.migrate(ctx, target, source, s.tests.SafeMigrations.MovePKToCQOnly, forceMigrate))
```
erezrokah, line 273: Was this change intentional?

erezrokah:
For the destinations that don't support these migrations (safe or forced) how would that look like from a user perspective?

bbernays:
> For the destinations that don't support these migrations (safe or forced) how would that look like from a user perspective?

For a forced migration, if a destination cannot detect that there is a difference between the before and after then the destination will just insert the data rather than dropping first

erezrokah (approved):

PR #1571, opened 2024-03-18
feat: Expose InvocationID to Plugin Client

bbernays:
#### Summary

Depends On: https://github.com/cloudquery/plugin-pb/pull/21

This PR allows plugin clients to expose functionality that depends on the InvocationID

### DIFF
--- a/plugin/plugin.go
+++ b/plugin/plugin.go
@@ -18,6 +18,7 @@ var ErrNotImplemented = fmt.Errorf("not implemented")
 
 type NewClientOptions struct {
 	NoConnection bool
+	InvocationID string
 	PluginMeta   Meta
 }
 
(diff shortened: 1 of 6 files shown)


plugin/plugin.go
```diff
@@ -220,7 +228,10 @@ func (p *Plugin) Init(ctx context.Context, spec []byte, options NewClientOptions
 
 	options.PluginMeta = p.Meta()
 
-	p.client, err = p.newClient(ctx, p.logger, spec, options)
```
erezrokah, line 223: Do we want to add this field to the log only if it's not empty?

bbernays: https://github.com/cloudquery/cloudquery/issues/17322

erezrokah (approved):

PR #1572, opened 2024-03-20
feat: Handle unknown token types when getting team name in usage client

erezrokah:
#### Summary

Every time we add a new token type (e.g. test connection token type) the existing plugin versions that use the usage client can't use the new token type since the default on unknown token type is to fail when trying to get the team name.
This PR updates the code to fallback to `_CQ_TEAM_NAME` so even if a new token type is added we try to get the team name from `_CQ_TEAM_NAME`.
This will ensure that if we add a new token type, existing versions can still get the team name as long as `_CQ_TEAM_NAME` exists.

---

- [x] Update or add tests 🧪

### DIFF
--- a/premium/usage.go
+++ b/premium/usage.go
@@ -405,14 +405,16 @@ func (u *BatchUpdater) getTeamNameByTokenType(tokenType auth.TokenType) (string,
 			return "", fmt.Errorf("expected to find exactly one team for API key, found %d", len(resp.JSON200.Items))
 		}
 		return resp.JSON200.Items[0].Name, nil
-	case auth.SyncRunAPIKey, auth.SyncTestConnectionAPIKey:
+	default:
 		team := os.Getenv("_CQ_TEAM_NAME")
 		if team == "" {
-			return "", fmt.Errorf("_CQ_TEAM_NAME environment variable not set")
+			switch tokenType {
+			case auth.SyncRunAPIKey, auth.SyncTestConnectionAPIKey:
+				return "", fmt.Errorf("_CQ_TEAM_NAME environment variable not set")
+			}
+			return "", fmt.Errorf("unsupported token type: %v", tokenType)
 		}
 		return team, nil
-	default:
-		return "", fmt.Errorf("unsupported token type: %v", tokenType)
 	}
 }
 
--- a/premium/usage_test.go
+++ b/premium/usage_test.go
@@ -21,14 +21,19 @@ import (
 )
 
 type MockTokenClient struct {
+	tokenType auth.TokenType
 }
 
 func (*MockTokenClient) GetToken() (auth.Token, error) {
 	return auth.Token{}, nil
 }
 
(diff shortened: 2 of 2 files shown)


erezrokah:
> * **WriterMemory_BatchWriter_wide_batch10k_bytes100M-8 bytes/op**: 196,719 ⬆️ **14.56% increase** vs. [f8a8610](https://github.com/cloudquery/plugin-sdk/commit/f8a86104338da80d69ed4a8895b07d62c8e9d31a)

> * **WriterMemory_StreamingBatchWriter_wide_defaults-8 bytes/op**: 7,218 ⬇️ **43.05% decrease** vs. [f8a8610](https://github.com/cloudquery/plugin-sdk/commit/f8a86104338da80d69ed4a8895b07d62c8e9d31a)
> * **WriterMemory_StreamingBatchWriter_wide_defaults-8 ns/op**: 148,972 ⬇️ **8.65% decrease** vs. [f8a8610](https://github.com/cloudquery/plugin-sdk/commit/f8a86104338da80d69ed4a8895b07d62c8e9d31a)

Unrelated but some of these benchmarks seem noisy

premium/usage.go
```diff
@@ -405,14 +405,16 @@ func (u *BatchUpdater) getTeamNameByTokenType(tokenType auth.TokenType) (string,
 			return "", fmt.Errorf("expected to find exactly one team for API key, found %d", len(resp.JSON200.Items))
 		}
 		return resp.JSON200.Items[0].Name, nil
-	case auth.SyncRunAPIKey, auth.SyncTestConnectionAPIKey:
+	default:
```
disq, line 408: Could this have consequences in the future where it would allow other (potential) user-accessible keys to be used and spoof the team name, at least in older plugins?

erezrokah: If you run plugins/CLI locally you can always spoof the environment variables. The backend still enforces API keys can only access what they are permitted to do.

If this code runs in managed syncs we can control which environment variables to pass and decide not to pass `_CQ_TEAM_NAME` for other tokens

disq: CLI: Backend still enforces api keys but the team name could be different, and with an "unhandled" new type of user key (and an older plugin version, which doesn't explicitly support that key) this code would allow a different team name to be reported.

erezrokah: > this code would allow a different team name to be reported.

What would be the impact of that? API keys are tied to a team to they only have a single team so if the team is spoofed auth will fail. For user tokens the impact would be to report usage not on the team specified in the config, which they can modify regardless?
If we want to enforce usage reporting limitations it shouldn't be in the client code

disq: yeah the keys are still tied to a team so it wouldn't really do anything, ok good :)

disq (approved):

erezrokah:
Thanks for the review @disq I think this stills needs @hermanschaaf to approve

hermanschaaf (approved):

PR #1584, opened 2024-03-28
feat: Test duplicated primary key insertion

bbernays:
#### Summary

This PR adds a way for destinations to opt-in or out of testing what happens if a source sends duplicate data


schema/testdata.go
```diff
@@ -231,7 +233,10 @@ func (tg *TestDataGenerator) Generate(table *Table, opts GenTestDataOptions) arr
 
 	var records []arrow.Record
 	for j := 0; j < opts.MaxRows; j++ {
-		tg.counter++
+		if !opts.DuplicateData {
+			tg.counter++
```
disq, line 237: What kind of dup data is generated when`tg.counter` isn't incremented?

bbernays: All `id` columns are the same:
```
	// special case for auto-incrementing id column, used for to determine ordering in tests
	if arrow.IsInteger(dataType.ID()) && colName == "id" {
		return `` + strconv.Itoa(tg.counter) + ``
	}
```

disq (approved):

plugin/testing_write_upsert.go
```diff
@@ -132,3 +132,74 @@ func (s *WriterTestSuite) testUpsertAll(ctx context.Context) error {
 
 	return nil
 }
+
+func (s *WriterTestSuite) testInsertDuplicatePK(ctx context.Context) error {
```
erezrokah, line 136: Would it make sense to build the table and data manually instead of using `schema.TestTable` and `tg.Generate`?
This seems like a specific use case so not sure we need to expose another option to the test generator (also we override the ID column, so might as well build the table ourselves). See reference in `testUpsertBasic`

I think that would also make it easier to do the validation in the test, as we mostly want to validate the insert didn't fail and the most recent row was persisted

plugin/testing_write_upsert.go
```diff
@@ -132,3 +132,74 @@ func (s *WriterTestSuite) testUpsertAll(ctx context.Context) error {
...
+	if diff := RecordsDiff(table.ToArrowSchema(), records, []arrow.Record{extractLastRowFromRecord(table, normalRecord)}); diff != "" {
+		return fmt.Errorf("record differs after insert: %s", diff)
+	}
+
+	return nil
+}
+
+func extractLastRowFromRecord(table *schema.Table, existingRecord arrow.Record) arrow.Record {
+	sc := table.ToArrowSchema()
+	var lastRecord []arrow.Record
```
erezrokah, line 182:
```suggestion
```

We can probably move this closer to where it's used

plugin/testing_write_upsert.go
```diff
@@ -132,3 +132,74 @@ func (s *WriterTestSuite) testUpsertAll(ctx context.Context) error {
...
+	bldr := array.NewRecordBuilder(memory.DefaultAllocator, sc)
+	for i, c := range table.Columns {
+		col := existingRecord.Column(i)
+		err := bldr.Field(i).AppendValueFromString(col.ValueStr(int(existingRecord.NumRows()) - 1))
+		if err != nil {
+			panic(fmt.Sprintf("failed to unmarshal json `%v` for column %v: %v", col.ValueStr(int(existingRecord.NumRows())-1), c.Name, err))
+		}
+	}
+	lastRecord = append(lastRecord, bldr.NewRecord())
```
erezrokah, line 191:
```suggestion
	lastRecord := append([]arrow.Record, bldr.NewRecord())
```

Would this work?

plugin/testing_write_upsert.go
```diff
@@ -132,3 +132,74 @@ func (s *WriterTestSuite) testUpsertAll(ctx context.Context) error {
...
+	return nil
+}
+
+func extractLastRowFromRecord(table *schema.Table, existingRecord arrow.Record) arrow.Record {
+	sc := table.ToArrowSchema()
+	var lastRecord []arrow.Record
+	bldr := array.NewRecordBuilder(memory.DefaultAllocator, sc)
+	for i, c := range table.Columns {
+		col := existingRecord.Column(i)
+		err := bldr.Field(i).AppendValueFromString(col.ValueStr(int(existingRecord.NumRows()) - 1))
```
erezrokah, line 186: Could we extract `col.ValueStr(int(existingRecord.NumRows()) - 1`? It's used a bit below too

plugin/testing_write_upsert.go
```diff
@@ -132,3 +132,74 @@ func (s *WriterTestSuite) testUpsertAll(ctx context.Context) error {
...
+		if err != nil {
+			panic(fmt.Sprintf("failed to unmarshal json `%v` for column %v: %v", col.ValueStr(int(existingRecord.NumRows())-1), c.Name, err))
```
erezrokah, line 188: We should return an error instead of panicking, also not sure `failed to unmarshal json` is clear enough

plugin/testing_write_upsert.go
```diff
@@ -132,3 +132,74 @@ func (s *WriterTestSuite) testUpsertAll(ctx context.Context) error {
...
+	}
+	lastRecord = append(lastRecord, bldr.NewRecord())
+	bldr.Release()
+
+	arrowTable := array.NewTableFromRecords(sc, lastRecord)
+	columns := make([]arrow.Array, sc.NumFields())
+	for n := 0; n < sc.NumFields(); n++ {
+		concatenated, err := array.Concatenate(arrowTable.Column(n).Data().Chunks(), memory.DefaultAllocator)
+		if err != nil {
+			panic(fmt.Sprintf("failed to concatenate arrays: %v", err))
```
erezrokah, line 199: Same here we should return an error instead of panicking

erezrokah (changes requested):
This is indeed a good test to have, I think it can be simplified by dropping the test data generator usage and using an approach similar to `testUpsertBasic`, WDYT?

bbernays:
> This is indeed a good test to have, I think it can be simplified by dropping the test data generator usage and using an approach similar to `testUpsertBasic`, WDYT?

We can't use something as basic as `testUpsertBasic` because we need multiple rows per record otherwise the records won't be in the same batch, but I can look into something more basic

PR #1599, opened 2024-04-03
feat: Add State Client support to append only destinations

disq:
Enables the state client to load the ~~lexicographically~~ -latest value from the table. ~~Which works for most things (date strings) but not for unpadded integers (fixed in [`be1227d` (#1599)](https://github.com/cloudquery/plugin-sdk/pull/1599/commits/be1227d535c96dd45812172aa3f8efc1c90c227c) where we don't attempt any sorting if value lengths differ).~~ This way append-only destinations can also be used for state (with a bit of duplication) reliably.

Currently state client reads the whole table on init to memory, and writes everything back on Flush. Also a change list is added to avoid unnecessary writing.

It correctly handles zero-trimmed floats and timestamps.

### DIFF
--- a/internal/clients/state/v3/state.go
+++ b/internal/clients/state/v3/state.go
@@ -3,8 +3,8 @@ package state
 import (
 	"bytes"
 	"context"
-	"fmt"
 	"io"
+	"strings"
 	"sync"
 
 	"github.com/apache/arrow/go/v15/arrow"
@@ -114,8 +114,11 @@ func NewClientWithTable(ctx context.Context, pbClient pb.PluginClient, table *sc
 			keys := record.Columns()[0].(*array.String)
 			values := record.Columns()[1].(*array.String)
 			for i := 0; i < keys.Len(); i++ {
-				fmt.Println(keys.Value(i), values.Value(i))
-				c.mem[keys.Value(i)] = values.Value(i)
+				k, v := keys.Value(i), values.Value(i)
+				curVal, ok := c.mem[k]
+				if !ok || strings.Compare(curVal, v) == -1 { // set only if not exists or greater lexicographically
+					c.mem[k] = v
+				}
 			}
 		}
 	}
--- a/internal/clients/state/v3/state.go
+++ b/internal/clients/state/v3/state.go
@@ -19,12 +19,13 @@ const keyColumn = "key"
 const valueColumn = "value"
 
 type Client struct {
-	client pb.PluginClient
-	mem    map[string]string
-	mutex  *sync.RWMutex
-	keys   []string
-	values []string
-	schema *arrow.Schema
+	client  pb.PluginClient
+	mem     map[string]string
+	changes map[string]struct{}
+	mutex   *sync.RWMutex
+	keys    []string
+	values  []string
+	schema  *arrow.Schema
 }
 
 func Table(name string) *schema.Table {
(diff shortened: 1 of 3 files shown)


internal/clients/state/v3/state.go
```diff
@@ -119,16 +131,17 @@ func (c *Client) SetKey(_ context.Context, key string, value string) error {
...
 }
 
 func (c *Client) Flush(ctx context.Context) error {
-	c.mutex.RLock()
-	defer c.mutex.RUnlock()
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
 	bldr := array.NewRecordBuilder(memory.DefaultAllocator, c.schema)
-	for k, v := range c.mem {
+	for k := range c.changes {
```
disq, line 142: Write only the changed keys on flush.

internal/clients/state/v3/state.go
```diff
@@ -108,7 +116,11 @@ func NewClient(ctx context.Context, pbClient pb.PluginClient, tableName string)
 			keys := record.Columns()[0].(*array.String)
 			values := record.Columns()[1].(*array.String)
 			for i := 0; i < keys.Len(); i++ {
-				c.mem[keys.Value(i)] = values.Value(i)
+				k, v := keys.Value(i), values.Value(i)
+				curVal, ok := c.mem[k]
+				if !ok || len(curVal) != len(v) || strings.Compare(curVal, v) == -1 { // set only if not exists, of different length, or greater lexicographically
```
disq, line 121: Write only if it's a new key, or the length of value differs (because we give up sorting at this stage)
if both conditions don't suit, only write if we're writing a "larger" value with the same length (`value2` will overwrite `value1`, but `value0` won't)

hermanschaaf:
I'd like to take a look tomorrow, I haven't read the code here in great detail but I'm not sure this can be done without introducing (at least the potential of) edge cases for which it won't work correctly 🤔

disq:
> I'd like to take a look tomorrow, I haven't read the code here in great detail but I'm not sure this can be done without introducing (at least the potential of) edge cases for which it won't work correctly 🤔

Could add a config option to specify what the value format actually is (but it would have to be per-table so boo-hoo) or maybe try to auto detect values based on format (since we're reading all of them anyway)

erezrokah:
What's the use case for it? You could always add another destination only for storing the state, we even test for it here https://github.com/cloudquery/cloudquery/blob/e400adb35ff8331522d8eda4f7279f4d49191071/cli/cmd/testdata/different-backend-from-destination.yml#L9

disq:
@erezrokah As Aparture Science's motto goes, "We do what we must because we can". In the current state it "seems" to work with non-overwrite write modes, but not really as there's no ordering guarantee (and it will read all data and last row for each key "wins")

This way (still thinking about how to solve caveats...) it's a bit more polished as it will still work albeit writing duplicates (but storing few KVs many times is still cheap)

erezrokah:
> it "seems" to work with non-overwrite write modes

So maybe exit with an error when someone uses it in that way?

erezrokah:
> There's no way to detect which mode

We can detect in the CLI? Don't allow `backend_options.connection` to a plugin that has `write_mode: append`

disq:
some destinations don't care if it's `append` or `overwrite` but simply ignore PKs (clickhouse)

erezrokah:
> some destinations don't care if it's `append` or `overwrite` but simply ignore PKs (clickhouse)

Isn't this another issue to fix? If a destination doesn't support `overwrite` and someone uses it it should fail instead of pretending to work

hermanschaaf:
I think we shouldn't do this to be honest... what if the column's order needs to be in reverse? Or it's a UUID, and we need the last inserted value? Or the plugin actually wants to reset the state? The intention of the state interface was to be a very simple key-value store, leaving the logic of value updates to the client.

I think the only thing we could maybe consider doing with append-only destinations is to have them version every key, and then we always return the latest version. For this we'd need to be aware of whether we are working with an append-only destination or not.

PR #1604, opened 2024-04-04
feat: Add versioning to state client

disq:
Instead-of https://github.com/cloudquery/plugin-sdk/pull/1599

This way append only destinations can also be used as state clients.

Currently state client reads the whole table on init to memory, and writes everything back on Flush. Also a change list is added to avoid unnecessary writing.


erezrokah (commented):
Do we really need version? Can't consumers do `Set(v1-<key>, <value>`)?
They can embed the version in the key name no?

erezrokah:
Maybe an example how this would work with append only destinations can help

disq:
Screenshot 2024-04-04 at 15 51 10

Example screenshot from bigquery. The last two rows with `version` filled in are the recent syncs with this. (Previous data was synced with the latest BigQuery source v1.2.2)
On the state `GetKey` call, we return the latest version for the given key. If the state data is from an old client it would return the latest one depending on retrieval order (last retrieved value for each key would overwrite all previous values) but once it's migrated then we always return the latest version.

The current protocol's `Read` call dumps the whole table so there isn't an option to selectively get only the requested key anyway.

internal/clients/state/v3/state.go
```diff
@@ -107,28 +120,57 @@ func NewClient(ctx context.Context, pbClient pb.PluginClient, tableName string)
...
+				k, val := keys.Value(i), values.Value(i)
+
+				if _, ok := c.mem[k]; !ok {
+					c.mem[k] = make(map[uint64]string)
+				}
+				var ver uint64
+				if versions.IsValid(i) {
+					ver = versions.Value(i)
+				}
+				c.mem[k][ver] = val
```
hermanschaaf, line 134: Why do we need to keep all the versions in memory? We should be able to keep only the latest version, along with its version number? 

And maybe in flush we can only write and increment the version by one if the value of the key was updated. Otherwise we might be writing a lot of intermediate values that will never be used

erezrokah:
Can we make the version column optional? That way we can gradually add it to source plugins?
I would even say it should be the sync ID/invocation ID so you can trace state to syncs.

I think it would make the code simpler too instead of doing "automatic versioning" each time a new client is created

hermanschaaf:
> Can we make the version column optional? That way we can gradually add it to source plugins?

Agree, that would be nice.

> I would even say it should be the sync ID/invocation ID so you can trace state to syncs.

I don't think we can do this (or at least not only this, without another column), because we need the version to be sortable so ew can get the latest version in append-only destinations.

internal/clients/state/v3/state.go
disq: done in https://github.com/cloudquery/plugin-sdk/pull/1604/commits/12a143b3ada6d01b24be7982eea814a1e580223e

disq:
What would making version optional achieve? Currently it works whether the table has `version` column or not (but it will migrate the table to have it)

erezrokah:
> What would making version optional achieve?

It will allow us to gradually roll this change and test it instead of impacting all plugins instantly, reducing the risk.

disq:
> > What would making version optional achieve?
> 
> It will allow us to gradually roll this change and test it instead of impacting all plugins instantly, reducing the risk.

@erezrokah we can add it as a stateclient v4?

erezrokah:
I don't think we need a new protocol version if we make it backwards compatible/optional

disq:
> Can't we add `NewClientWithOps` instead of a new package?

Implementing it like that wouldn't reduce the risk completely.

erezrokah:
> Implementing it like that wouldn't reduce the risk completely.

I get that and ok with it. I'm mostly want to make sure the migration thing/additional column doesn't breaking existing states

erezrokah (approved):

disq:
> > What would making version optional achieve?
> 
> It will allow us to gradually roll this change and test it instead of impacting all plugins instantly, reducing the risk.

@erezrokah although... this is baked into the source plugin. So, destinations can't selectively opt-in into using this feature. If a source plugin uses it, it will effect all destinations in that config.

disq:
@erezrokah we could add it into:
```go
type BackendOptions struct {
	TableName  string
	Connection string
}
```
which would then make it available in the backend spec

hermanschaaf (approved):
Nice! 👍

PR #1605, opened 2024-04-04
fix: Update Otel Schema version

bbernays:
#### Summary

Keep resource version in line with the packages specified in go.mod

### DIFF
--- a/serve/opentelemetry.go
+++ b/serve/opentelemetry.go
@@ -4,7 +4,7 @@ import (
 	"github.com/cloudquery/plugin-sdk/v4/plugin"
 
 	"go.opentelemetry.io/otel/sdk/resource"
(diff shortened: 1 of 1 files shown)


erezrokah (approved):

erezrokah (commented):
Can we also disable automatic updates for this package?
See example in https://github.com/cloudquery/cloudquery/blob/4839d53643176cc23f69da3680d697b5257a1b8c/.github/renovate.json5#L28 (we don't need the `matchFileNames` part)

bbernays:
> Can we also disable automatic updates for this package? See example in https://github.com/cloudquery/cloudquery/blob/4839d53643176cc23f69da3680d697b5257a1b8c/.github/renovate.json5#L28 (we don't need the `matchFileNames` part)

@erezrokah - I will do that if I can't add tests that would catch this

PR #1611, opened 2024-04-05
feat: Support arbitrary map values for structs (scalar)

disq:


scalar/scalar.go
```diff
@@ -208,7 +209,23 @@ func AppendToBuilder(bldr array.Builder, s Scalar) {
...
+				panic(fmt.Sprintf("expected map[string]any, got %T", v))
+			}
+
+			m = make(map[string]any, reflectedMap.Len())
+			for _, key := range reflectedMap.MapKeys() {
+				if key.Kind() != reflect.String {
+					panic(fmt.Sprintf("expected map[string]any, got %T", v))
+				}
+				value := reflectedMap.MapIndex(key)
+				m[key.String()] = value.Interface()
```
candiduslynx, line 225: (nit) although it's fine here, the good rule of thumb is to check via `can interface` method before calling `interface

candiduslynx (approved):

erezrokah:
Can we add a test for this?

disq:
@erezrokah added in https://github.com/cloudquery/plugin-sdk/pull/1611/commits/05787e7f862107312d24a71ad9d6a17f7353b7b5

PR #1632, opened 2024-04-12
fix(test): Slice rows properly when reading in tests

candiduslynx:
We never actually required the read records to be a single-row ones, so we may as well slice them ourselves

### DIFF
--- a/plugin/plugin_read.go
+++ b/plugin/plugin_read.go
@@ -18,8 +18,16 @@ func (p *Plugin) readAll(ctx context.Context, table *schema.Table) ([]arrow.Reco
 	// nolint:prealloc
 	var records []arrow.Record
 	for record := range ch {
-		records = append(records, record)
+		records = append(records, sliceToSingleRowRecord(record)...)
 	}
 
 	return records, err
+}
+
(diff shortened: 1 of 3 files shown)


erezrokah (commented):
Can you explain what bug this PR fixes? Maybe via a unit test?

candiduslynx:
> Can you explain what bug this PR fixes? Maybe via a unit test?

During development of the Databricks destination I found the issue with returned records: in Databricks it's easy to have multirow arrow.Record returned directly from the database. If we pass this as is (with transformation) to the sdk, some of the methods we use in tests will fail:
* sorting records based on `id` or other fields
* skipping first rows (via skipping records)

Basically, it means that our SDK expects to receive a single-row record, but it's never documented.

We have a similar issue inDuckDB destination, where we opted to reslice the returned record manually, but I think the slicing should be handled by the SDK itself:
https://github.com/cloudquery/cloudquery/blob/de3297476bcc5e3aa71bc1b14c4fdf6cf5b49ba0/plugins/destination/duckdb/client/read.go#L68

erezrokah (approved):

PR #1642, opened 2024-04-22
feat: Batch resources into a single record on source side

candiduslynx:


scheduler/batch.go
```diff
@@ -0,0 +1,133 @@
...
+	for {
+		select {
+		case r, ok := <-w.ch:
+			if !ok {
+				w.sync()
+				return
+			}
+
+			w.rows = append(w.rows, r)
+			if size > 0 && len(w.rows) == size {
```
disq, line 67: After the `append()`, `len(w.rows)` can't be 0 any more, so `size > 0` check isn't needed.

scheduler/batch.go
```diff
@@ -0,0 +1,133 @@
...
+		case <-ticker.Chan():
+			flush()
+
+		case done := <-w.flush:
+			flush()
+			close(done)
+
+		case <-ctx.Done():
+			// this means the request was cancelled
+			return // after this NO other call will succeed
```
disq, line 80: Not flushing here. Intentional?

scheduler/batch.go
```diff
@@ -0,0 +1,133 @@
...
+			// discard newWorker
+			close(newWorker.ch)
+			close(newWorker.flush)
+			newWorker.builder.Release()
+		}
+	}
+	wr := v.(*worker)
+	if !loaded {
+		// now loaded basically determines if we need to call the goroutine.
+		// Note that loaded will be false oly for the goroutine that actually successfully stores the worker
```
disq, line 108:
```suggestion
		// Note that loaded will be false only for the goroutine that actually successfully stores the worker
```

scheduler/batch.go
```diff
@@ -0,0 +1,133 @@
...
+	w.res <- &message.SyncInsert{Record: w.builder.NewRecord()}
+}
+
+func (w *worker) work(ctx context.Context, size int, timeout time.Duration) {
+	ticker := writers.NewTicker(timeout)
+	defer ticker.Stop()
+	flush := func() {
+		w.sync()
+		ticker.Reset(timeout)
+		w.rows = w.rows[:0]
```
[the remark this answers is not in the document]
candiduslynx: I don't see significant improvement, have you tested with the following flags to ensure the proper results?
`-test.count 10 -test.benchmem -test.benchtime 1000x`

See https://github.com/cloudquery/plugin-sdk/pull/1172

scheduler/batch.go
candiduslynx: nice catch

scheduler/batch.go
candiduslynx: yep, as ctx was cancelled (copy-paste from batchwriter)

scheduler/batch.go
candiduslynx: 3b4291a0947c2015c95d88e6860e62ea3b5dbac6

disq (approved):
LGTM but needs a better PR title

scheduler/scheduler.go
```diff
@@ -210,16 +207,22 @@ func (s *Scheduler) Sync(ctx context.Context, client schema.ClientMeta, tables s
...
 		}
 	}()
-	for resource := range resources {
+
+	b := newBatcher(res, 50, 15*time.Second)
+	done := ctx.Done()
+	for {
 		select {
-		case <-ctx.Done():
+		case <-done:
```
disq, line 216: This optimization is probably unnecessary as `ctx.Done()` will also return the same member every time, and it's hopefully optimized

candiduslynx: it will return the same value, but the func will call atomic loads, which impacts performance (see cancelCtx implementation, for instance)

candiduslynx: https://cs.opensource.google/go/go/+/refs/tags/go1.22.2:src/context/context.go;l=438-451

scalar/inet.go
```diff
@@ -129,6 +129,12 @@ func (s *Inet) Set(val any) error {
 	return nil
 }
 
+func (s *Inet) ByteSize() int64 { return int64(len(s.Value.IP) + len(s.Value.Mask)) }
```
murarustefaan, line 132: I'm not sure here how these get serialized so feel free to resolve this if not applicable, but do we need to track the `/` separator as well?

scalar/time.go
```diff
@@ -94,3 +94,7 @@ func (s *Time) Set(value any) error {
 		return s.Int.Set(value)
 	}
 }
+
+var (
+	_ Scalar = (*Time)(nil)
+)
```
murarustefaan, line 100: Not adding ByteSize() to Time scalars? Is this on purpose?

scalar/inet.go
candiduslynx: we are looking only at the bytes being sent, so there's no real reason to account for the separator when it'll be tracked on the destination side once it decides on the destination value

scalar/time.go
candiduslynx: yes, as they really occupy only the underlying Int, which we get from embedding

scheduler/scheduler.go
```diff
@@ -207,22 +206,26 @@ func (s *Scheduler) Sync(ctx context.Context, client schema.ClientMeta, tables s
 			panic(fmt.Errorf("unknown scheduler %s", s.strategy.String()))
 		}
 	}()
+
+	const (
+		rows    = 50
```
erezrokah, line 211: Can we make these configurable by the source plugin?
I would also default the rows to 1 (e.g. no batching), so we can roll this out gradually while we test additional sources.
We tested AWS so far, so we could make 50 the default for AWS

candiduslynx: f01a1ff199b0d68a7136d5d14938083deb906e08

erezrokah (approved):
The code looks good, could we debug logs on when do the flushes similar to the destination batcher logs we have?

Can be done in a follow up PR since the batching is disabled by default

PR #1553, opened 2024-03-05
fix: MixedBatchWriter should nil the slice instead of zeroing

disq:
Helps with memory.

Benched with https://github.com/cloudquery/plugin-sdk/pull/1552, old:

```
BenchmarkWriterMemory/MixedBatchWriter_batch10k_bytes100M
BenchmarkWriterMemory/MixedBatchWriter_batch10k_bytes100M-10    	     826	   1445595 ns/op	      1393 bytes/op
BenchmarkWriterMemory/MixedBatchWriter_defaults
BenchmarkWriterMemory/MixedBatchWriter_defaults-10              	     670	   1504799 ns/op	      1393 bytes/op
BenchmarkWriterMemory/MixedBatchWriter_wide_batch10k_bytes100M
BenchmarkWriterMemory/MixedBatchWriter_wide_batch10k_bytes100M-10         	       8	 128966609 ns/op	    172785 bytes/op
BenchmarkWriterMemory/MixedBatchWriter_wide_defaults
BenchmarkWriterMemory/MixedBatchWriter_wide_defaults-10                   	       9	 118630315 ns/op	     60355 bytes/op
```

new:
```
BenchmarkWriterMemory/MixedBatchWriter_batch10k_bytes100M
BenchmarkWriterMemory/MixedBatchWriter_batch10k_bytes100M-10    	     690	   1476479 ns/op	      1401 bytes/op
BenchmarkWriterMemory/MixedBatchWriter_defaults
BenchmarkWriterMemory/MixedBatchWriter_defaults-10              	     687	   1539707 ns/op	      1401 bytes/op
BenchmarkWriterMemory/MixedBatchWriter_wide_batch10k_bytes100M
BenchmarkWriterMemory/MixedBatchWriter_wide_batch10k_bytes100M-10         	       8	 129154755 ns/op	    173592 bytes/op
BenchmarkWriterMemory/MixedBatchWriter_wide_defaults
BenchmarkWriterMemory/MixedBatchWriter_wide_defaults-10                   	       9	 111416773 ns/op	     44588 bytes/op
```

PR #1645, opened 2024-04-22
fix: Use `clear` for mixed batch writer

candiduslynx:
Fixes the issue with `cap` used on `nil` slice introduced in https://github.com/cloudquery/plugin-sdk/pull/1553

### DIFF
--- a/writers/mixedbatchwriter/mixedbatchwriter.go
+++ b/writers/mixedbatchwriter/mixedbatchwriter.go
@@ -192,7 +192,6 @@ func (m *batchManager[A, T]) flush(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
-	m.batch = nil
 	clear(m.batch) // GC can work
 	m.batch = m.batch[:0]
 	return nil


erezrokah (approved):

writers/mixedbatchwriter/mixedbatchwriter.go
```diff
@@ -193,6 +193,8 @@ func (m *batchManager[A, T]) flush(ctx context.Context) error {
 		return err
 	}
 	m.batch = nil
+	clear(m.batch) // GC can work
```
erezrokah, line 196: Aren't we setting `m.batch` to `nil` a bit before?

erezrokah (changes requested):

writers/mixedbatchwriter/mixedbatchwriter.go
candiduslynx: typo, thanks!

candiduslynx: 6ea725a8fafa3f38c1d07010395aa5e597e9c0fb

candiduslynx:
alt approach: to add `size` member to the batcher managers & compare this.
@erezrokah @disq I actually don't have preference on this

erezrokah (approved):

PR #1647, opened 2024-04-22
fix: Properly handle records with multiple rows in batching

candiduslynx:
Fixes https://github.com/cloudquery/cloudquery-issues/issues/1567

Out of scope: https://github.com/cloudquery/cloudquery-issues/issues/1655


writers/streamingbatchwriter/streamingbatchwriter.go
```diff
@@ -396,8 +396,10 @@ func (s *streamingWorkerManager[T]) run(ctx context.Context, wg *sync.WaitGroup,
 			}
 
 			var recSize int64
+			rows := int64(1) // at least 1 row for messages wothout records
```
disq, line 399:
```suggestion
			rows := int64(1) // at least 1 row for messages without records
```

writers/batchwriter/batchwriter.go
```diff
@@ -122,10 +122,15 @@ func (w *BatchWriter) Close(context.Context) error {
...
-	sizeBytes := int64(0)
-	resources := make([]*message.WriteInsert, 0, w.batchSize)
+	sizeBytes, sizeRows := int64(0), int64(0)
+	resources := make([]*message.WriteInsert, 0, w.batchSize) // at least we have 1 row per record
 	ticker := writers.NewTicker(w.batchTimeout)
 	defer ticker.Stop()
+
+	send := func() {
+		w.flushTable(ctx, tableName, resources)
+		resources, sizeBytes, sizeRows = resources[:0], 0, 0
```
disq, line 132: `clear(resources)`? (are we misusing `clear`?)

candiduslynx: We didn't have this in the simple batch writer, but sure, I'll add it

candiduslynx: 5107d0b15d1a31ff9fe8d2217f85eef66ea97e5c adds clear & also exposes `capped` to be reused

candiduslynx:
TODO: Extract slicer to handle batching by rows & size.
This should produce several records if the new appended record has too much data (e.g. batch size is 5, hut we get a record with 100 rows).

writers/cap.go
```diff
@@ -0,0 +1,16 @@
+package writers
```
erezrokah, line 1: Can we drop this abstraction?
I'm not sure it's necessary to wrap an integer this way (also it seems we don't need it to be generic as we always use `int64`).

writers/batchwriter/batchwriter.go
```diff
@@ -122,39 +122,46 @@ func (w *BatchWriter) Close(context.Context) error {
...
+				if rows.Current() > 0 {
 					w.flushTable(ctx, tableName, resources)
 				}
 				return
 			}
 
-			if (w.batchSize > 0 && len(resources) >= w.batchSize) || (w.batchSizeBytes > 0 && sizeBytes+util.TotalRecordSize(r.Record) >= int64(w.batchSizeBytes)) {
-				w.flushTable(ctx, tableName, resources)
+			// we append data first not to call calculations twice
+			resources = append(resources, r)
```
erezrokah, line 148: Doesn't this changes the logic to flush after the limit is reached, instead of before? i.e. we'll always exceed the limit before flushing? I think we want to flush before hitting it, no?

For BigQuery I think this means the write will fail (@hermanschaaf can confirm)

writers/streamingbatchwriter/streamingbatchwriter.go
```diff
@@ -396,25 +397,27 @@ func (s *streamingWorkerManager[T]) run(ctx context.Context, wg *sync.WaitGroup,
...
 			var recSize int64
+			rowSize := int64(1) // at least 1 row for messages without records
 			if ins, ok := any(r).(*message.WriteInsert); ok {
 				recSize = util.TotalRecordSize(ins.Record)
+				rowSize = ins.Record.NumRows()
 			}
 
-			if (s.batchSizeRows > 0 && sizeRows >= s.batchSizeRows) || (s.batchSizeBytes > 0 && sizeBytes+recSize >= s.batchSizeBytes) {
+			ensureOpened()
+			clientCh <- r
```
erezrokah, line 407: Same comment here about flushing after hitting the limit, not before

writers/streamingbatchwriter/streamingbatchwriter_test.go
```diff
@@ -197,7 +197,7 @@ func TestStreamingBatchSizeRows(t *testing.T) {
 	}()
 
 	table := schema.Table{Name: "table1", Columns: []schema.Column{{Name: "id", Type: arrow.PrimitiveTypes.Int64}}}
-	record := array.NewRecord(table.ToArrowSchema(), nil, 0)
+	record := getRecord(table.ToArrowSchema(), 1)
```
erezrokah, line 200: What's the reason we need to change this test? Do we not support records without rows now?

erezrokah: Ah maybe because we were always incrementing the row count regardless if the record has rows or not

writers/mixedbatchwriter/mixedbatchwriter.go
```diff
@@ -199,39 +199,41 @@ func (m *batchManager[A, T]) flush(ctx context.Context) error {
...
+	logger    zerolog.Logger
 }
 
 func (m *insertBatchManager) append(ctx context.Context, msg *message.WriteInsert) error {
-	if len(m.batch) == cap(m.batch) || m.curBatchSizeBytes+util.TotalRecordSize(msg.Record) > m.maxBatchSizeBytes {
-		if err := m.flush(ctx); err != nil {
-			return err
-		}
-	}
+	// we append data first not to call calculations twice
```
erezrokah, line 210: Same issue with flushing after the limit was reached

erezrokah (commented):
Added a few comments, the main ones are:
1. Don't think we need the abstraction over int64 as it's basically wrapping `+=`, `>` and `Get`
2. This seems to change the logic to flush after the limit had reached instead of before, it won't be an issue for limiting by rows I think (since it works the same for single row records, and is broken for multi row records anyway), but can be an issue for limiting by bytes. Regardless probably best to flush before going over the limit

Also per your comment in https://github.com/cloudquery/plugin-sdk/pull/1647#issuecomment-2105983280 if we roll out multi-rows this will become an issue for batching both by rows and by bytes, and with single rows records we're less likely to exceed the batch size.
Not sure what to do about that, and how slicing can impact the performance improvements we get from multi records

writers/batchwriter/batchwriter.go
candiduslynx: Initially `writers.Cap` included `OverflownBy(a A)` method that checked this, I can add it back

writers/streamingbatchwriter/streamingbatchwriter_test.go
candiduslynx: yeah, we had records with 0 rows in tests

writers/mixedbatchwriter/mixedbatchwriter.go
candiduslynx: 9dbac8d613ad90520e9c2daf08314d3528c00a72

writers/streamingbatchwriter/streamingbatchwriter.go
candiduslynx: 9dbac8d613ad90520e9c2daf08314d3528c00a72

writers/batchwriter/batchwriter.go
candiduslynx: 9dbac8d613ad90520e9c2daf08314d3528c00a72

writers/batchwriter/batchwriter.go
```diff
@@ -122,39 +122,45 @@ func (w *BatchWriter) Close(context.Context) error {
...
+				if rows.Current() > 0 {
 					w.flushTable(ctx, tableName, resources)
 				}
 				return
 			}
 
-			if (w.batchSize > 0 && len(resources) >= w.batchSize) || (w.batchSizeBytes > 0 && sizeBytes+util.TotalRecordSize(r.Record) >= int64(w.batchSizeBytes)) {
-				w.flushTable(ctx, tableName, resources)
+			dataBytes := util.TotalRecordSize(r.Record)
+			if rows.OverflownBy(r.Record.NumRows()) || bytes.OverflownBy(dataBytes) {
```
candiduslynx, line 148: btw, we had inconsintency here as we checked for bytes overflow in advance, but not for the rows overflow

writers/cap.go
erezrokah: > We can make it `int64` always, but we have a lot of duplication for this, so I'd rather leave this.

I don't think the abstraction removes the duplicate it just wraps it in `Add` instead of `+=`, `GetCurrent` instead of direct access

writers/batchwriter/batchwriter.go
erezrokah: > btw, we had inconsintency here as we checked for bytes overflow in advance, but not for the rows overflow

Not sure I understand this comment, we did `len(resources) >= w.batchSize)` then we appended so we checked the overflow for rows before appending

writers/cap.go
erezrokah: Also `OverflownBy` instead of `>`

candiduslynx: > Also `OverflownBy` instead of `>`

`max > 0 && curr + delta > max`, but yes.

Still, I feel like it's a good abstraction & I haven't seen any argument why not (we basically get to keep the limit & the curr value coupled together with distinct meaning).

writers/batchwriter/batchwriter.go
candiduslynx: we checked if the rows were already at the level (but yeah, it worked for the single row records & not for the multirow here).

writers/cap.go
erezrokah: > I haven't seen any argument why not

You have to learn the abstraction to understand how the code works

> we basically get to keep the limit & the curr value coupled together with distinct meaning

We already have them together in the function scope?

erezrokah: Also if you want to update the code you need to modify the abstraction so it breaks fast (e.g. `OverflownBy` to keep the original order of the operations)

candiduslynx: I've added the `OverflownBy` now & removed `ReachedLimit`

erezrokah: > I've added the `OverflownBy` now & removed `ReachedLimit`

Sorry for not being clear, that was my point. To fix the bug you had to break the abstraction on the first change. Which probably means it's not ready to be abstracted yet

candiduslynx: I'm not ready to die on the hill for this impl, so sure, I'll remove it.

erezrokah (approved):

writers/batchwriter/batchwriter_test.go
```diff
@@ -122,40 +122,44 @@ func TestBatchFlushDifferentMessages(t *testing.T) {
...
-	}); err != nil {
-		t.Fatal(err)
-	}
-	// we need to wait for the batch to be flushed
-	time.Sleep(time.Second * 2)
-
-	if testClient.InsertsLen() != 2 {
-		t.Fatalf("expected 2 insert messages, got %d", testClient.InsertsLen())
+	for i := 0; i < 5; i++ {
+		batchSize := rand.Intn(100) + 10
```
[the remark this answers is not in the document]
candiduslynx: we log via adding this to the test name (in the `t.Run` call)

PR #1649, opened 2024-04-26
feat: Add function to create state client and gRPC backend connection

maaarcelino:
I'd like to propose adding a function to the `state` package to make it easier to initialise the client along with a gRPC connection.

Currently our plugins that support incremental syncing have this code in the `plugin/client.go` file:

```
var stateClient state.Client
if options.BackendOptions == nil {
	c.logger.Info().Msg("No backend options provided, using no state backend")
	stateClient = &state.NoOpClient{}
	c.backendConn = nil
} else {
	c.backendConn, err = grpc.DialContext(ctx, options.BackendOptions.Connection,
		grpc.WithTransportCredentials(insecure.NewCredentials()),
		grpc.WithDefaultCallOptions(
			grpc.MaxCallRecvMsgSize(maxMsgSize),
			grpc.MaxCallSendMsgSize(maxMsgSize),
		),
	)
	if err != nil {
		return fmt.Errorf("failed to dial grpc source plugin at %s: %w", options.BackendOptions.Connection, err)
	}
	stateClient, err = state.NewClient(ctx, c.backendConn, options.BackendOptions.TableName)
	if err != nil {
		return fmt.Errorf("failed to create state client: %w", err)
	}
	c.logger.Info().Str("table_name", options.BackendOptions.TableName).Msg("Connected to state backend")
}
```

My solution would replace it with:
```
stateClient, backendConn, err := state.NewGrpcConnectedClient(ctx, options.BackendOptions)
```

### DIFF
--- a/state/client.go
+++ b/state/client.go
@@ -14,15 +14,21 @@ type ConnectionOptions struct {
 	MaxMsgSizeInBytes int
 }
 
-// NewGrpcConnectedClient returns a state client and a gRPC connection to the state backend with a 100MiB max message size.
+type NopCloser interface {
+	Close() error
+}
+
+// NewGrpcConnectedClient returns a state client and initialises the gRPC connection to the state backend with a 100MiB max message size.
 // The state client is guaranteed to be non-nil (it defaults to the NoOpClient).
-func NewGrpcConnectedClient(ctx context.Context, backendOpts *plugin.BackendOptions) (Client, *grpc.ClientConn, error) {
+// You must call Close() on the returned closer object.
+func NewGrpcConnectedClient(ctx context.Context, backendOpts *plugin.BackendOptions) (Client, NopCloser, error) {
 	return NewGrpcConnectedClientWithOptions(ctx, backendOpts, ConnectionOptions{MaxMsgSizeInBytes: defaultMaxMsgSizeInBytes})
 }
 
-// NewGrpcConnectedClientWithOptions returns a state client and a gRPC connection to the state backend.
+// NewGrpcConnectedClientWithOptions returns a state client and initialises the gRPC connection to the state backend.
 // The state client is guaranteed to be non-nil (it defaults to the NoOpClient).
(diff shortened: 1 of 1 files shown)


disq (approved):

state/client.go
```diff
@@ -0,0 +1,48 @@
...
+
+// NewGrpcConnectedClient returns a state client and a gRPC connection to the state backend with a 100MiB max message size.
+// The state client is guaranteed to be non-nil (it defaults to the NoOpClient).
+func NewGrpcConnectedClient(ctx context.Context, backendOpts *plugin.BackendOptions) (Client, *grpc.ClientConn, error) {
+	return NewGrpcConnectedClientWithOptions(ctx, backendOpts, ConnectionOptions{MaxMsgSizeInBytes: defaultMaxMsgSizeInBytes})
+}
+
+// NewGrpcConnectedClientWithOptions returns a state client and a gRPC connection to the state backend.
+// The state client is guaranteed to be non-nil (it defaults to the NoOpClient).
+func NewGrpcConnectedClientWithOptions(ctx context.Context, backendOpts *plugin.BackendOptions, opts ConnectionOptions) (Client, *grpc.ClientConn, error) {
```
candiduslynx, line 25: IMO we should provide a nop closer instead of `*grpc.ClientConn` or something similar (& also make sure that the clients calling this method will run `close` manually.
Otherwise, this returned value has no sense

candiduslynx (changes requested):

hermanschaaf (approved):
Yeah I think this is an improvement; I don't love the dependency on `github.com/cloudquery/plugin-sdk/v4/plugin` here but we can maybe shift it around to another package / do some refactoring in the future.

state/client.go
hermanschaaf: Yeah returning a Closer is probably a good call

maaarcelino:
Closing as I want to refactor this as per Alex's suggestions

PR #1650, opened 2024-04-29
feat: Add function to create state client and gRPC backend connection

maaarcelino:
This is the second iteration of https://github.com/cloudquery/plugin-sdk/pull/1649. Alex suggested that we add a `Close()` method to the state client itself and we then wouldn't have to return three objects.

Strictly speaking this is not a breaking change, but `Close()` must now be called on the state client which wasn't a requirement until now.


state/state.go
```diff
@@ -33,9 +36,9 @@ func NewClientWithOptions(ctx context.Context, conn *grpc.ClientConn, tableName
 	}
 	if slices.Contains(versions.Versions, 3) {
 		if opts.Versioned {
-			return stateV3.NewClientWithTable(ctx, pbPluginV3.NewPluginClient(conn), stateV3.VersionedTable(tableName))
+			return stateV3.NewClientWithTable(ctx, pbPluginV3.NewPluginClient(conn), stateV3.VersionedTable(tableName), conn)
```
erezrokah, line 39:
```suggestion
			return stateV3.NewClientWithTable(ctx, conn, stateV3.VersionedTable(tableName))
```

Maybe and then do `pbPluginV3.NewPluginClient(conn)` inside `NewClientWithTable`? Same comment a bit below

state/client.go
```diff
@@ -0,0 +1,50 @@
+package state
```
erezrokah, line 1: Maybe move the content of `state.go` to this file? Other we'd have two separate files, both have methods for creating a new client

erezrokah (commented):
Great improvement, added a couple of comments.
I don't think it's a breaking change since the requirement to close the state client is only when you use the new function no?

state/client.go
maaarcelino: Done in 9037389

state/state.go
maaarcelino: Done in 9037389

maaarcelino:
@erezrokah The existing calls to `NewClient` and `NewClientWithOptions` are affected by this change, because those constructors now return a client that must be closed

erezrokah:
> @erezrokah The existing calls to `NewClient` and `NewClientWithOptions` are affected by this change, because those constructors now return a client that must be closed

Aren't existing consumers of those calls pass in an existing connection and already close it themselves? Example in https://github.com/cloudquery/cloudquery-private/blob/c4d9fe8f1802def24da7af5c11ce76c02a00c833/plugins/source/stripe/resources/plugin/client.go#L105

erezrokah (approved):
Regardless of the breaking comment I approved the changes. We only mark protocol breaking changes as breaking in the SDK so no need to make this breaking anyway.

I think we should be able to gradually update plugins to use this approach without it breaking existing plugins (that init and close the connection outside `NewClient`). If that's not the case let me know

PR #1668, opened 2024-05-08
feat: Allow reporting usage as breakdown by table

murarustefaan:
Implements changes requested in [#1502](https://github.com/cloudquery/cloudquery-issues/issues/1502).

The solution consists of adding a new func `IncreaseWithTableBreakdown(table string, rows uint32)` to the existing client, which we can declaratively replace in required non-database plugins incrementally.

### DIFF
--- a/premium/usage.go
+++ b/premium/usage.go
@@ -28,6 +28,12 @@ const (
 	defaultMaxTimeBetweenFlushes = 30 * time.Second
 )
 
+const (
+	UsageIncreaseMethodUnset = iota
+	UsageIncreaseMethodTotal
+	UsageIncreaseMethodBreakdown
+)
+
 type TokenClient interface {
 	GetToken() (auth.Token, error)
 	GetTokenType() auth.TokenType
@@ -144,13 +150,14 @@ type BatchUpdater struct {
 
 	// State
 	sync.Mutex
-	rows           uint32
-	tables         map[string]uint32
-	lastUpdateTime time.Time
-	triggerUpdate  chan struct{}
-	done           chan struct{}
-	closeError     chan error
-	isClosed       bool
+	rows                uint32
+	tables              map[string]uint32
+	lastUpdateTime      time.Time
+	triggerUpdate       chan struct{}
+	done                chan struct{}
+	closeError          chan error
+	isClosed            bool
+	usageIncreaseMethod int
 }
 
 func NewUsageClient(meta plugin.Meta, ops ...UsageClientOptions) (UsageClient, error) {
@@ -222,6 +229,10 @@ func NewUsageClient(meta plugin.Meta, ops ...UsageClientOptions) (UsageClient, e
 }
 
 func (u *BatchUpdater) Increase(rows uint32) error {
+	if u.usageIncreaseMethod == UsageIncreaseMethodBreakdown {
+		return fmt.Errorf("mixing usage increase methods is not allowed, use IncreaseForTable instead")
+	}
+
 	if rows <= 0 {
 		return fmt.Errorf("rows must be greater than zero got %d", rows)
 	}
@@ -232,6 +243,10 @@ func (u *BatchUpdater) Increase(rows uint32) error {
 
 	u.Lock()
 	defer u.Unlock()
+
+	if u.usageIncreaseMethod == UsageIncreaseMethodUnset {
+		u.usageIncreaseMethod = UsageIncreaseMethodTotal
+	}
 	u.rows += rows
 
 	// Trigger an update unless an update is already in process
@@ -244,6 +259,10 @@ func (u *BatchUpdater) Increase(rows uint32) error {
 }
 
 func (u *BatchUpdater) IncreaseForTable(table string, rows uint32) error {
+	if u.usageIncreaseMethod == UsageIncreaseMethodTotal {
+		return fmt.Errorf("mixing usage increase methods is not allowed, use Increase instead")
+	}
+
 	if rows <= 0 {
 		return fmt.Errorf("rows must be greater than zero got %d", rows)
 	}
@@ -254,6 +273,10 @@ func (u *BatchUpdater) IncreaseForTable(table string, rows uint32) error {
 
 	u.Lock()
 	defer u.Unlock()
+
+	if u.usageIncreaseMethod == UsageIncreaseMethodUnset {
+		u.usageIncreaseMethod = UsageIncreaseMethodBreakdown
+	}
 
 	u.tables[table] += rows
 	u.rows += rows
--- a/premium/usage.go
+++ b/premium/usage.go
@@ -7,7 +7,7 @@ import (
 	"math/rand"
 	"net/http"
 	"os"
-	"sync/atomic"
+	"sync"
 	"time"
 
 	cqapi "github.com/cloudquery/cloudquery-api-go"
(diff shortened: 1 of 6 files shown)


erezrokah (approved):

premium/usage.go
```diff
@@ -226,7 +230,34 @@ func (u *BatchUpdater) Increase(rows uint32) error {
...
+	select {
+	case u.triggerUpdate <- struct{}{}:
+	default:
+		return nil
+	}
+
+	return nil
+}
+
+func (u *BatchUpdater) IncreaseWithTableBreakdown(table string, rows uint32) error {
```
hermanschaaf, line 247:
```suggestion
func (u *BatchUpdater) IncreaseForTable(table string, rows uint32) error {
```

I would suggest naming it something like this, since a breakdown (at least to me) would imply multiple tables being passed in to this function

premium/usage.go
```diff
@@ -264,6 +295,31 @@ func (u *BatchUpdater) Close() error {
...
+		usage = append(usage, cqapi.UsageIncreaseTablesInner{
+			Name: key,
+			Rows: int(value),
+		})
+	}
+
+	return usage, u.rows
+}
+
+func (u *BatchUpdater) chunkTableUsage(usage []cqapi.UsageIncreaseTablesInner, total uint32) {
```
hermanschaaf, line 312: what does `chunk` mean in this context?

premium/usage.go
```diff
@@ -321,23 +384,24 @@ func (u *BatchUpdater) backgroundUpdater() {
...
 		queryStartTime := time.Now()
 
 		resp, err := u.apiClient.IncreaseTeamPluginUsageWithResponse(ctx, u.teamName, cqapi.IncreaseTeamPluginUsageJSONRequestBody{
 			RequestId:  uuid.New(),
 			PluginTeam: u.pluginMeta.Team,
 			PluginKind: u.pluginMeta.Kind,
 			PluginName: u.pluginMeta.Name,
-			Rows:       int(numberToUpdate),
+			Rows:       int(rowsToUpdate),
+			Tables:     &tables,
```
hermanschaaf, line 398: We need to be mindful here that if the `tables` parameter is supplied, its total number of rows should equal the number given in the `rows` parameter. 

My reading of the code here is that this will not necessarily be the case, as if no table breakdowns were supplied, we will pass an empty array here, rather than nil. I haven't tested how the backend handles this, but probably we should explicitly not pass `tables` if it hasn't been populated

premium/usage.go
murarustefaan: We store total rows and table breakdowns on the usage client, so this function basically **chunks** the table usages and total rows that were sent to the API from the ones tracked on the usage client.

Does it make sense?

premium/usage.go
murarustefaan: Yeah I can do that, it makes sense.

premium/usage.go
hermanschaaf: Hmm, I've not heard of `chunk` being used this way; to me `chunk` means to divide something into parts, and this doesn't really seem to apply here. Maybe `subtractTableUsage`?

premium/usage.go
murarustefaan: That's what I initially started with, but then I went with this approach as there are a lot of functions exposed by the SDK that go for the `With` prefix, so I though I should not stray from that convention.
Wdyt @hermanschaaf, should I go ahead and rename to `IncreaseForTable` or leave it as it is?

hermanschaaf: `With` is often used for function Options, I don't think there's a hard rule it has to be used for function names other than in that case. I would say let's use a name that fits, and to me "breakdown" doesn't seem to fit here

premium/usage.go
murarustefaan: Done

premium/usage.go
murarustefaan: Done

premium/usage.go
murarustefaan: Done

hermanschaaf:
In addition to my [comment here](https://github.com/cloudquery/plugin-sdk/pull/1668#discussion_r1595269086), there's one more case to be mindful of: clients shouldn't mix calls to `IncreaseWithTableBreakdown` and `Increase`. If they do, again, the total of rows won't match the total number of rows reported by table.

I see two okay-ish options for solving this:
 1. create a new BatchUpdater client that specifically only allows updating by table
 2. if usage reporting was mixed, send two requests, one for the table-less rows, and one for the rest

I think I would lean toward option 1, so that the exposed interface enforces correct usage.

premium/usage.go
```diff
@@ -143,8 +143,10 @@ type BatchUpdater struct {
 	maxTimeBetweenFlushes time.Duration
 
 	// State
+	rows           uint32
+	tables         map[string]uint32
+	mutex          sync.Mutex
```
candiduslynx, line 148:
```suggestion
	sync.Mutex
```
will allow locking on `BatchUpdater` (`b.Lock()` essentially)

premium/usage.go
```diff
@@ -226,7 +230,34 @@ func (u *BatchUpdater) Increase(rows uint32) error {
...
-	u.rowsToUpdate.Add(rows)
+	u.mutex.Lock()
+	defer u.mutex.Unlock()
+	u.rows += rows
+
+	// Trigger an update unless an update is already in process
+	select {
+	case u.triggerUpdate <- struct{}{}:
+	default:
+		return nil
```
candiduslynx, line 241: can be left empty
```suggestion
```

premium/usage.go
```diff
@@ -226,7 +230,34 @@ func (u *BatchUpdater) Increase(rows uint32) error {
...
+	}
+
+	return nil
+}
+
+func (u *BatchUpdater) IncreaseForTable(table string, rows uint32) error {
+	if rows <= 0 {
```
candiduslynx, line 248: should be done under lock

candiduslynx (changes requested):

premium/usage.go
murarustefaan: @candiduslynx why though? Why should you lock before validating the input? Or did you mean the `isClosed()` check?

candiduslynx: ah, I overlooked that it's an external input, not the state
😵‍💫

murarustefaan:
> 1. create a new BatchUpdater client that specifically only allows updating by table 
> I think I would lean toward option 1, so that the exposed interface enforces correct usage.

@hermanschaaf how would that client be built? 
The way I see it would be creating another type of Updater, or providing a WithTableBreakdownUsage something as a Option when building the UsageClient and have it stored as a configuration field and checked.

candiduslynx (approved):

hermanschaaf:
> WithTableBreakdownUsage something as a Option when building the UsageClient and have it stored as a configuration field and checked.

@murarustefaan Yes, I think this could work. Or maybe we don't need an option, but we can internally keep track of two booleans (e.g. `increaseUsed`, `increaseForTableUsed`) and return an error if ever both these booleans were to be true

murarustefaan:
I'll go with the boolean approach as I think from a user / plugin developer perspective, using a Option would not be the best UX I think, as you need to do an extra step just to make sure this works the way we want it to. It would also require a new type of client and I do not think it is worth it.

murarustefaan:
@hermanschaaf what do you think of the implementation in the latest commit?

hermanschaaf (approved):
Nice, yeah looks good 👍

PR #1682, opened 2024-05-15
feat: Add test connection

disq:
wip

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -124,3 +124,5 @@ require (
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace github.com/cloudquery/plugin-pb-go => ../plugin-pb-go
--- a/plugin/plugin_testconn.go
+++ b/plugin/plugin_testconn.go
@@ -1,0 +1,70 @@
+package plugin
+
+import (
+	"context"
+	"fmt"
+)
+
+type TestConnFailureCode string
+
+const (
+	TestConnFailureCodeUnknown            TestConnFailureCode = "UNKNOWN"
+	TestConnFailureCodeUnimplemented      TestConnFailureCode = "UNIMPLEMENTED"
+	TestConnFailureCodeInvalidSpec        TestConnFailureCode = "INVALID_SPEC"
+	TestConnFailureCodeInvalidCredentials TestConnFailureCode = "INVALID_CREDENTIALS"
+)
+
+type TestConnError struct {
+	Code    TestConnFailureCode
+	Message error
+}
+
+func NewTestConnError(code TestConnFailureCode, err error) *TestConnError {
+	if code == "" {
+		code = TestConnFailureCodeUnknown
+	}
+	return &TestConnError{
+		Code:    code,
+		Message: err,
+	}
+}
+
+var ErrTestConnUnimplemented = &TestConnError{
+	Code: TestConnFailureCodeUnimplemented,
+}
+
+func (e *TestConnError) Error() string {
+	return e.Message.Error()
+}
+
+func (e *TestConnError) Unwrap() error {
+	return e.Message
+}
+
+func (e *TestConnError) Is(err error) bool {
+	if err2, ok := err.(*TestConnError); ok {
+		return e.Code == err2.Code
+	}
+	return false
+}
+
+// ConnectionTester is an interface that can be implemented by a plugin client to enable explicit connection testing.
+type ConnectionTester interface {
+	TestConnection(context.Context, []byte) *TestConnError
+}
+
+func (p *Plugin) TestConnection(ctx context.Context, spec []byte) *TestConnError {
+	if !p.mu.TryLock() {
+		return &TestConnError{
+			Code:    TestConnFailureCodeUnknown,
+			Message: fmt.Errorf("plugin already in use"),
+		}
+	}
+	defer p.mu.Unlock()
+
+	if v, ok := p.client.(ConnectionTester); ok {
+		return v.TestConnection(ctx, spec)
+	}
+
+	return ErrTestConnUnimplemented
+}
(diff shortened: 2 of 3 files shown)


plugin/plugin_testconn.go
```diff
@@ -0,0 +1,70 @@
...
+func (e *TestConnError) Is(err error) bool {
+	if err2, ok := err.(*TestConnError); ok {
+		return e.Code == err2.Code
+	}
+	return false
+}
+
+// ConnectionTester is an interface that can be implemented by a plugin client to enable explicit connection testing.
+type ConnectionTester interface {
+	TestConnection(context.Context, []byte) *TestConnError
```
disq, line 53: Having this return `error` makes things weirder when we want to extract the failure code. Could be an interface that embeds `error`, and a `Code() string`...

plugin/plugin_testconn.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func (p *Plugin) TestConnection(ctx context.Context, spec []byte) *TestConnError {
+	if !p.mu.TryLock() {
+		return &TestConnError{
+			Code:    TestConnFailureCodeUnknown,
```
disq, line 59: This is an example, I think we can safely remove this locking mechanism from TestConnection.

examples/simple_plugin/plugin/client.go
```diff
@@ -91,6 +91,19 @@ func Configure(_ context.Context, logger zerolog.Logger, spec []byte, opts plugi
 	}, nil
 }
 
+func TestConnection(_ context.Context, _ zerolog.Logger, spec []byte) *plugin.TestConnError {
```
mnorbury, line 94:
```suggestion
func TestConnection(_ context.Context, _ zerolog.Logger, spec []byte) error {
```

plugin/options.go
```diff
@@ -46,6 +46,13 @@ func WithTeam(team string) Option {
 	}
 }
 
+// WithConnectionTester can be specified by a plugin to enable explicit connection testing, given a spec.
+func WithConnectionTester(tester ConnectionTester) Option {
```
disq, line 50: Having this as an option (rather than an optional interface under `plugin.Client`, like how `OnBeforeSender` was implemented) enables us to directly call the given function without creating/instantiating the plugin with `Configure` (or `client.New`) and makes things much simpler.

erezrokah: Don't we want to fallback to `Init` when the test connection is not implemented? How would we do this in this approach?

plugin/plugin_testconn.go
```diff
@@ -0,0 +1,62 @@
...
+	if err2, ok := err.(*TestConnError); ok {
+		return e.Code == err2.Code
+	}
+	return false
+}
+
+type ConnectionTester func(ctx context.Context, logger zerolog.Logger, spec []byte) error
+
+func (p *Plugin) TestConnection(ctx context.Context, logger zerolog.Logger, spec []byte) error {
+	return p.testConnFn(ctx, logger, spec)
```
erezrokah, line 55: What will happen if `WithConnectionTester` was not called?

erezrokah (commented):
Added a couple of comments/questions

plugin/options.go
disq: This would be done in the CLI code once it detects the return value from this function is either an empty falsy value (in the case of an outdated SDK in the plugin) or `ErrTestConnUnimplemented`, to keep this code here simple and hassle free.

erezrokah: OK then I'll take a look at the CLI code once it's ready

hermanschaaf (approved):
I had a read through and it looks good to me, though I haven't read the CLI parts yet

examples/simple_plugin/go.mod
```diff
@@ -125,3 +125,5 @@ require (
 )
 
 replace github.com/cloudquery/plugin-sdk/v4 => ../../
+
+replace github.com/cloudquery/plugin-pb-go => github.com/cloudquery/plugin-pb-go v1.19.13-0.20240516105108-3fdf6ac15ea2
```
disq, line 129: To be removed before merge when `plugin-pb-go` is released.

go.mod
```diff
@@ -124,3 +124,5 @@ require (
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace github.com/cloudquery/plugin-pb-go => github.com/cloudquery/plugin-pb-go v1.19.13-0.20240516105108-3fdf6ac15ea2
```
disq, line 128: To be removed before merge when `plugin-pb-go` is released.

plugin/plugin_testconn.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"context"
+
+	"github.com/rs/zerolog"
+)
+
+type TestConnFailureCode string
+
+const (
+	TestConnFailureCodeUnknown            TestConnFailureCode = "UNKNOWN"
```
erezrokah, line 12: Shouldn't the error codes be part of the protocol in https://github.com/cloudquery/plugin-pb/pull/22?
So we don't have to reference the SDK directly in https://github.com/cloudquery/cloudquery/pull/17993/files#diff-9f63834674c5ec7944d55a63556099b99c2ce34c0633f8bad046241910e35acbR18

erezrokah: Otherwise we'd have to duplicate them in each SDK

disq: Empty string also falls back to UNKNOWN code, maybe we can do that? Do we want a strict list of error codes? Then we need a task force to determine the list.

erezrokah: > Do we want a strict list of error codes?

Don't we already have it in the SDK? I'm suggesting the enum for the codes should be defined at the protocol level. We can still have unknown errors

erezrokah: Maybe we don't need a code and only have a message

disq: Removed consts, keeping the code.

erezrokah (approved):

PR #1684, opened 2024-05-16
fix: Unwrap nested embedded structs

candiduslynx:
fixes https://github.com/cloudquery/cloudquery-issues/issues/1695

### DIFF
--- a/scalar/date32_test.go
+++ b/scalar/date32_test.go
@@ -5,6 +5,7 @@ import (
 	"time"
 
 	"github.com/apache/arrow/go/v16/arrow"
+	"github.com/stretchr/testify/assert"
 )
 
 func TestDate32Set(t *testing.T) {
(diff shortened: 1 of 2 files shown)


transformers/struct.go
```diff
@@ -228,9 +225,10 @@ func (t *structTransformer) ignoreField(field reflect.StructField) bool {
 	switch {
 	case len(field.Name) == 0,
 		slices.Contains(t.skipFields, field.Name),
-		!field.IsExported(),
 		isTypeIgnored(field.Type):
 		return true
+	case !field.IsExported():
```
candiduslynx, line 230: this is required as the embedded structs might contain other embedded structs and so on

scalar/date32_test.go
```diff
@@ -5,29 +5,27 @@ import (
...
 func TestDate32Set(t *testing.T) {
 	successfulTests := []struct {
 		source any
 		result Date32
 	}{
-		{source: time.Date(1900, 1, 1, 0, 0, 0, 0, time.Local), result: Date32{Value: arrow.Date32FromTime(time.Date(1900, 1, 1, 0, 0, 0, 0, time.Local)), Valid: true}},
-		{source: time.Date(1999, 12, 31, 12, 59, 59, 0, time.Local), result: Date32{Value: arrow.Date32FromTime(time.Date(1999, 12, 31, 0, 0, 0, 0, time.Local)), Valid: true}},
+		{source: time.Date(1900, 1, 1, 0, 0, 0, 0, time.UTC),
+			result: Date32{Value: arrow.Date32FromTime(time.Date(1900, 1, 1, 0, 0, 0, 0, time.UTC)), Valid: true}},
+		{source: time.Date(1999, 12, 31, 12, 59, 59, 0, time.UTC),
```
candiduslynx, line 18: this was failing on my machine/locality, so moved to utc

disq:
This could be opt-in in the first release. Can you check if there are high number of plugins affected by this change, other than Slack?

candiduslynx:
> This could be opt-in in the first release. Can you check if there are high number of plugins affected by this change, other than Slack?

@disq 
https://github.com/cloudquery/cloudquery-private/pull/2489

erezrokah:
Can we close this PR until we can find the time to discuss the default behavior we want via the issue?

PR #1691, opened 2024-05-17
fix: Handle marshaling errors for `scalar.JSON`

candiduslynx:
Instead of https://github.com/cloudquery/cloudquery-private/pull/2500

Fixes a case with some types that implement `json.Marshaler` interface but don't validate the resulting output and may return empty/nil `[]byte` array with `nil` error when marshaling.

### DIFF
--- a/scalar/json.go
+++ b/scalar/json.go
@@ -3,6 +3,7 @@ package scalar
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"reflect"
 
 	"github.com/apache/arrow/go/v16/arrow"
@@ -104,13 +105,16 @@ func (s *JSON) Set(val any) error {
 		buffer := &bytes.Buffer{}
 		encoder := json.NewEncoder(buffer)
 		encoder.SetEscapeHTML(false)
-		err := encoder.Encode(value)
+		err := validateJSONMarshalError(encoder.Encode(value))
 		if err != nil {
 			return err
 		}
 
 		// JSON encoder adds a newline to the end of the output that we don't want.
 		buf := bytes.TrimSuffix(buffer.Bytes(), []byte("\n"))
+		if len(buf) == 0 {
+			return nil // just means that the underlying error was set to nil by validateJSONMarshalError
+		}
 		// For map and slice jsons, it is easier for users to work with '[]' or '{}' instead of JSON's 'null'.
 		if bytes.Equal(buf, []byte(`null`)) {
 			if isEmptyStringMap(value) {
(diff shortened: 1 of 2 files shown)


scalar/json.go
```diff
@@ -168,3 +172,15 @@ func jsonBytesEqual(a, b []byte) (bool, error) {
 	}
 	return reflect.DeepEqual(j2, j), nil
 }
+
+func validateJSONMarshalError(err error) error {
+	var mErr *json.MarshalerError
+	var sErr *json.SyntaxError
+	if errors.As(err, &mErr) {
+		if errors.As(mErr.Err, &sErr) {
+			// means that we encountered the issue with marshaller interface, possibly empty result
```
erezrokah, line 181: Won't this set to `nil` for any syntax error? Not just `nil`/empty result? Not sure we'd want that

candiduslynx: Don't we fail the entire table resolving when such error happens?
I think storing `nil` is better than storing anything else.

Additionally, this double nested check is actually more or less the check for the invalid data returned by `MarshalJSON` implementation if the marshaled value.

erezrokah: > Don't we fail the entire table resolving when such error happens?
> I think storing `nil` is better than storing anything else.

But what if we got the type wrong for the column and it's not really a JSON, and it should be a string?
There's a different between not handling a `nil` value because we don't support `type RawMessage []byte` (which should be handled in the plugin), to accepting all syntax errors.

See example in https://github.com/cloudquery/cloudquery/pull/10826. With this SDK change we would set `nil` on the column without knowing it should be a text column

candiduslynx: In this case we'll get the user report that they don't see the column that's available in the API & we'll be able to address this instead of not seeing the whole resource.

erezrokah: > In this case we'll get the user report that they don't see the column that's available in the API & we'll be able to address this instead of not seeing the whole resource.

Harder to understand why the data is `nil` - is it missing because it's truly `nil` or because it errored out?
If the whole resource is missing it's more likely that the error will surface.

Also, since this issue is specific to the GCP plugin, I'm not sure we should change the SDK default behavior for all plugins to accept invalid JSON as a value and implicitly convert it to `nil`.

erezrokah: > instead of assuming `interface{} -> JSON` as we do now,

The JSON type is there not only for structs that can be marshelled to JSON, but also for any "complex" type

candiduslynx: > Harder to understand why the data is nil - is it missing because it's truly nil or because it errored out?
> If the whole resource is missing it's more likely that the error will surface.

That's why I created a double-layered check: it checks that the outer is caused by the struct implementing `json.Marshaler` while the inner is about incorrect JSON produced (namely, empty one).

erezrokah: > while the inner is about incorrect JSON produced

And still, we should change the default behavior for all plugins to accept invalid JSON. We don't do it for other types as well (e.g. set `nil` when we fail to parse a timestamp)

PR #1700, opened 2024-05-23
feat: Enable `NewConnectedClientWithOptions` to set `ClientOptions`

bbernays:
#### Summary

Currently source plugins using the `ConnectedClient` cannot be used with `append` only destinations as the `ClientOption` cannot be set. This PR exposes that option. This won't have any impact on CQ plugins as none of our plugins use `NewConnectedClientWithOptions`


bbernays:
This is the initial step in fixing https://github.com/cloudquery/cloudquery-issues/issues/1733

erezrokah (approved):

PR #1719, opened 2024-06-05
fix: Account for bytes limit properly when batching records for writing

candiduslynx:
Fixes https://github.com/cloudquery/cloudquery-issues/issues/1567
Fixes https://github.com/cloudquery/cloudquery-issues/issues/1655

The new `batch` package provides a reusable logic for batching, accounting both for the bytes & rows limits.


writers/batchwriter/batchwriter.go
```diff
@@ -6,7 +6,7 @@ import (
 	"sync"
 	"time"
 
-	"github.com/apache/arrow/go/v16/arrow/util"
+	"github.com/cloudquery/plugin-sdk/v4/batch"
```
disq, line 9: Let's do `internal/batch` so we don't have to worry about maintaining API compatibility

candiduslynx: a6e73a10440d2f50ffca1e945f69464a19a2b39f

internal/batch/slice_test.go
```diff
@@ -0,0 +1,122 @@
...
+	"time"
+
+	"github.com/apache/arrow/go/v16/arrow/util"
+	"github.com/cloudquery/plugin-sdk/v4/schema"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestSliceRecord(t *testing.T) {
+	for run := 0; run < 10; run++ {
+		rows := rand.Intn(200) + 5
```
candiduslynx, line 17:
```suggestion
	for run := 0; run < 5; run++ {
		rows := rand.Intn(100) + 5
```

internal/batch/slice.go
```diff
@@ -0,0 +1,143 @@
...
+
+func newSlicedRecord(r arrow.Record) *SlicedRecord {
+	if r.NumRows() == 0 {
+		return nil
+	}
+	res := SlicedRecord{
+		Record: r,
+		Bytes:  util.TotalRecordSize(r),
+	}
+	res.bytesPerRow = res.Bytes / r.NumRows()
```
erezrokah, line 132: Still going through this PR, but is it true that we're assuming each row has the same number of bytes?
Can't we get the number of bytes per row?

internal/batch/slice.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	switch {
+	case rows == 0:
+		// not even a single row fits
+		// we still need to process this, so slice by single row
+		return s.slice()
+	case rows < 0:
+		// as s.Record != nil we know that the limits are there in place & the s.Record.NumRows() > 0
+		panic("should never be here")
```
erezrokah, line 90: If this can never happen what's the reason we have this case?

internal/batch/slice.go
candiduslynx: we can't get amount of bytes per row (we could get it for fixed-size ones, but for variable-sized ones it'll create too much overhead)

internal/batch/slice.go
candiduslynx: we could remove it 💯 , just wanted to cover all possible cases (so if & when some code in-between changes we won't end up in mem corruption state)

internal/batch/slice.go
erezrokah: Doesn't this mean we can't reliably enforce the bytes limits? Let me try this out on a destination that needs the bytes limit to be a hard one, and not a soft one.
Maybe we can find a solution for those type of destinations

candiduslynx: We can't make a 100% correct implementation without making a huge toll on CPU, as currently even the sliced array will still have all buffers of the original mentioned & thus we would have the total bytes for the array equal to the value for slice of even a single row.
While we don't actually impose the "technically correct" limit in here, having this PR at least enables slicing huge records into smaller chunks to try to accommodate for the destinations that really require that.

In the current mainline the huge record with lots of rows that is x10 times more than the destination limit will be flushed all at once as a single write. This change addresses such issues.

erezrokah: > In the current mainline the huge record with lots of rows

But we only support records with multiple rows in a few DB plugins (and not as the default setting), and the intention with this change is to have all sources send multi row records by default.

We should first check how likely it is that we breach the limit, and the impact on destinations (not all destinations need a hard limit).

erezrokah (approved):

PR #1724, opened 2024-06-06
feat: Remove plugin option for logging error events to Sentry

murarustefaan:
Closes [#1809](https://github.com/cloudquery/cloudquery-issues/issues/1809).

### DIFF
--- a/serve/plugin.go
+++ b/serve/plugin.go
@@ -108,6 +108,7 @@ func (s *PluginServe) Serve(ctx context.Context) error {
 func (s *PluginServe) newCmdPluginServe() *cobra.Command {
 	var address string
 	var network string
+	var noSentry bool
 	var otelEndpoint string
 	var otelEndpointHeaders []string
 	var otelEndpointInsecure bool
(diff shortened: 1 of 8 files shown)


serve/plugin.go
```diff
@@ -306,12 +268,7 @@ func (s *PluginServe) newCmdPluginServe() *cobra.Command {
 	cmd.Flags().StringVar(&otelEndpointURLPath, "otel-endpoint-urlpath", "", "Open Telemetry HTTP collector endpoint URL path")
 	cmd.Flags().StringArrayVar(&otelEndpointHeaders, "otel-endpoint-headers", []string{}, "Open Telemetry HTTP collector endpoint headers")
 	cmd.Flags().BoolVar(&otelEndpointInsecure, "otel-endpoint-insecure", false, "use Open Telemetry HTTP endpoint (for development only)")
-	cmd.Flags().BoolVar(&noSentry, "no-sentry", false, "disable sentry")
```
murarustefaan, line 309: I'm not sure here tbh what would be the best approach, whether to remove this or have it do nothing and mark as deprecated. Wdyt?

disq: I would keep the option but make it deprecated and no-op.

murarustefaan: That makes sense, I wasn't sure if deleting it was the best decision as if it was like that, I'd have had to delete it from the other plugin sdks as well.

candiduslynx (approved):

erezrokah (approved):

serve/plugin.go
erezrokah: Yeah I think if we remove the flag new plugins might not work with older CLI versions (don't remember if we're doing strict arg parsing)

PR #1743, opened 2024-06-19
fix: Allow marshaling plain string values into JSON scalars

candiduslynx:
Fixes https://github.com/cloudquery/cloudquery-issues/issues/1875

### DIFF
--- a/scalar/json_test.go
+++ b/scalar/json_test.go
@@ -1,6 +1,12 @@
 package scalar
 
-import "testing"
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
 
 type Foo struct {
 	Num int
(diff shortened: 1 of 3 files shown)


disq (approved):

hermanschaaf (approved):

erezrokah:
Isn't a plain string not a valid JSON? Only a quoted one is a valid JSON?

candiduslynx:
> Isn't a plain string not a valid JSON? Only a quoted one is a valid JSON?

yes (as we check `a b c`) value instead of `"a b c"`
so when we do json.Marshal it properly escapes the string + quotes it

scalar/json_test.go
```diff
@@ -15,6 +21,7 @@ func TestJSONSet(t *testing.T) {
 		{source: "{}", result: JSON{Value: []byte("{}"), Valid: true}},
 		{source: `"test"`, result: JSON{Value: []byte(`"test"`), Valid: true}},
 		{source: "1", result: JSON{Value: []byte("1"), Valid: true}},
+		{source: `json some data`, result: JSON{Value: []byte(`"json some data"`), Valid: true}},
```
erezrokah, line 24: The source value is an unquoted string so not sure why we're treating it as a valid JSON

erezrokah:
> yes (as we check `a b c`) value instead of `"a b c"`
> so when we do json.Marshal it properly escapes the string + quotes it

But this PR makes it so that any unquoted string value is treated as a valid JSON no?

PR #1751, opened 2024-06-19
feat: Better OTEL traces, add metrics

erezrokah:
#### Summary

Trying to follow https://www.honeycomb.io/blog/opentelemetry-best-practices-naming to have some convention and namespaces for traces.
Also removed `syncDfs` as it doesn't really add additional information.

I've added `invocationID` to traces, but not metrics as using a random value for a metric attribute will cause high cardinality (that will increase over time)

---

- [ ] Update or add tests 🧪

### DIFF
--- a/plugin/plugin_source.go
+++ b/plugin/plugin_source.go
@@ -88,12 +88,10 @@ func (p *Plugin) Sync(ctx context.Context, options SyncOptions, res chan<- messa
 	if p.client == nil {
 		return fmt.Errorf("plugin not initialized. call Init() first")
 	}
-	// startTime := time.Now()
 
 	if err := p.client.Sync(ctx, options, res); err != nil {
 		return fmt.Errorf("failed to sync unmanaged client: %w", err)
 	}
 
-	// p.logger.Info().Uint64("resources", p.metrics.TotalResources()).Uint64("errors", p.metrics.TotalErrors()).Uint64("panics", p.metrics.TotalPanics()).TimeDiff("duration", time.Now(), startTime).Msg("sync finished")
 	return nil
 }
--- a/serve/plugin.go
+++ b/serve/plugin.go
@@ -8,18 +8,11 @@ import (
 	"os/signal"
 	"strings"
 	"syscall"
-	"time"
 
 	"github.com/cloudquery/plugin-sdk/v4/helpers/grpczerolog"
 	"github.com/cloudquery/plugin-sdk/v4/plugin"
 	"github.com/cloudquery/plugin-sdk/v4/premium"
 	"github.com/cloudquery/plugin-sdk/v4/types"
-	"go.opentelemetry.io/otel"
-	"go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetrichttp"
-	"go.opentelemetry.io/otel/exporters/otlp/otlptrace"
-	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp"
-	"go.opentelemetry.io/otel/sdk/metric"
-	"go.opentelemetry.io/otel/sdk/trace"
 
 	pbDestinationV0 "github.com/cloudquery/plugin-pb-go/pb/destination/v0"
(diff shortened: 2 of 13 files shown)


serve/plugin.go
```diff
@@ -142,11 +145,15 @@ func (s *PluginServe) newCmdPluginServe() *cobra.Command {
 
 			if otelEndpoint != "" {
 				resource := newResource(s.plugin)
-				opts := []otlptracehttp.Option{
+				traceOptions := []otlptracehttp.Option{
```
candiduslynx, line 148: I think `if otelEndpoint != "" {` should be extracted into a separate`setupOTel` func

serve/plugin.go
```diff
@@ -140,45 +136,12 @@ func (s *PluginServe) newCmdPluginServe() *cobra.Command {
 				logger = log.Output(zerolog.ConsoleWriter{Out: os.Stdout}).Level(zerologLevel)
 			}
 
-			if otelEndpoint != "" {
```
erezrokah, line 143: Moved in to `opentelemetry.go`

serve/plugin.go
```diff
@@ -140,45 +136,12 @@ func (s *PluginServe) newCmdPluginServe() *cobra.Command {
...
-					if err := tp.Shutdown(context.Background()); err != nil {
-						logger.Error().Err(err).Msg("failed to shutdown OTLP trace exporter")
-					}
-				}()
-				otel.SetTracerProvider(tp)
+			shutdown, err := setupOtel(cmd.Context(), logger, s.plugin, otelEndpoint, otelEndpointInsecure, otelEndpointHeaders, otelEndpointURLPath)
+			if err != nil {
+				return fmt.Errorf("failed to setup OpenTelemetry: %w", err)
 			}
+			defer shutdown()
```
marianogappa, line 143: I think if `err != nil` this will blow up at defer time (e.g. https://go.dev/play/p/dOvHV41W1CE)

plugin/plugin_source.go
```diff
@@ -88,12 +88,10 @@ func (p *Plugin) Sync(ctx context.Context, options SyncOptions, res chan<- messa
 	if p.client == nil {
 		return fmt.Errorf("plugin not initialized. call Init() first")
 	}
-	// startTime := time.Now()
 
 	if err := p.client.Sync(ctx, options, res); err != nil {
 		return fmt.Errorf("failed to sync unmanaged client: %w", err)
 	}
 
-	// p.logger.Info().Uint64("resources", p.metrics.TotalResources()).Uint64("errors", p.metrics.TotalErrors()).Uint64("panics", p.metrics.TotalPanics()).TimeDiff("duration", time.Now(), startTime).Msg("sync finished")
```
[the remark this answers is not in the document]
marianogappa: Ah, you moved this, and it's one per table apparently.

marianogappa (approved):
Some comments.

plugin/plugin_source.go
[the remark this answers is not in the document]
erezrokah: Ah I just removed this commented line since it has been commented out for a while

serve/plugin.go
erezrokah: Shouldn't returning early when `err != nil` prevent us from reaching `defer shutdown()` ? 
https://github.com/cloudquery/plugin-sdk/pull/1751/files#diff-9d26691c28ec41ba1e4872ba4df8c1574c2ffc6582c6b82a65914f4e301a7517R141

marianogappa: Ah, I didn't expect that a failure to setup otel would cause the serve plugin to fail. Is that ok?

erezrokah: So we fail on setting OTEL when you explicitly pass the endpoint (same behavior as before) https://github.com/cloudquery/plugin-sdk/blob/e5e8e7ea650862e53214f381db6c22173fb04edb/serve/plugin.go#L144 and https://github.com/cloudquery/plugin-sdk/blob/e5e8e7ea650862e53214f381db6c22173fb04edb/serve/plugin.go#L169

PR #1757, opened 2024-06-20
feat: Add `duration_ms` to `table sync finished` log message

erezrokah:
#### Summary

Part of https://github.com/cloudquery/cloudquery-issues/issues/1778 (internal issue). Adds the duration in milliseconds to the table finished log message.
Tested this with open telemetry data that we send and the values match

---

- [ ] Update or add tests 🧪


scheduler/metrics.go
```diff
@@ -18,6 +18,7 @@ type TableClientMetrics struct {
 	Panics    uint64
 	StartTime time.Time
```
erezrokah, line 19: Looks like we can remove `StartTime` and `EndTime`, verifying this now

candiduslynx (approved):

PR #1759, opened 2024-06-21
fix: Don't allocate many loggers for source batching

candiduslynx:


scheduler/batch.go
```diff
@@ -53,7 +53,7 @@ func (s *BatchSettings) getBatcher(ctx context.Context, res chan<- message.SyncM
 			res:     res,
 			maxRows: s.MaxRows,
 			timeout: s.Timeout,
-			logger:  logger.With().Int("max_rows", s.MaxRows).Dur("timeout", s.Timeout).Logger(),
```
erezrokah, line 56: Maybe we should keep it as `Dur` to avoid the `.String()` allocation? I think the logger uses Milliseconds, so we could change the field name to `timeout_ms`

scheduler/batch.go
```diff
@@ -53,7 +53,7 @@ func (s *BatchSettings) getBatcher(ctx context.Context, res chan<- message.SyncM
...
+			logger:  logger.With().Int("max_rows", s.MaxRows).Str("timeout", s.Timeout.String()).Logger(),
```
candiduslynx, line 56:
```suggestion
			logger:  logger.With().Int("max_rows", s.MaxRows).Dur("timeout_ms", s.Timeout).Logger(),
```

erezrokah (approved):

PR #1768, opened 2024-06-21
fix: Skip printing irrelevant metrics

candiduslynx:


erezrokah (commented):
OK so I'm not sure this is a bug.
We print `table resolver started` only for top level tables (actually we print `top level table resolver started`).
And we print `table sync finished` for all tables (including relations), regardless if we actually resolved them (we skip resolving if the parent doesn't have resources).

For users relying on the `table sync finished` this might be breaking actually. I would say it's still useful to know since those tables were referenced in the spec

PR #1770, opened 2024-06-21
feat: Support AWS usage marketplace

bbernays:
This adds support for using CloudQuery plugins as part of the AWS Marketplace by changing the metering implementation to use the Marketplace instead, if specified via environment variables.


premium/usage.go
```diff
@@ -200,6 +206,20 @@ func NewUsageClient(meta plugin.Meta, ops ...UsageClientOptions) (UsageClient, e
 			TeamNameValue: u.teamName,
 		}, nil
 	}
+	// If user wants to use the AWS Marketplace for billing, don't even try to communicate with CQ API
+	if isAWSMarketplace() {
+		cfg, err := awsConfig.LoadDefaultConfig(context.TODO())
```
erezrokah, line 211: What's the reason for using `context.TODO()` in favor of `context.Background()`? Maybe we should add timeout as well for loading the config?

Also, can we put this logic in a `setupAWSMarketplace` function as `NewUsageClient` is getting quite long

premium/usage.go
```diff
@@ -240,6 +260,21 @@ func NewUsageClient(meta plugin.Meta, ops ...UsageClientOptions) (UsageClient, e
 	return u, nil
 }
 
+func isAWSMarketplace() bool {
+	return os.Getenv("CQ_AWS_MARKETPLACE_CONTAINER") == "true" || os.Getenv("CQ_AWS_MARKETPLACE_AMI") == "true"
```
erezrokah, line 264: Who's responsible for setting these? AWS?

premium/usage.go
```diff
@@ -240,6 +260,21 @@ func NewUsageClient(meta plugin.Meta, ops ...UsageClientOptions) (UsageClient, e
...
+}
+
+func awsMarketplaceProductCode() string {
+	if os.Getenv("CQ_AWS_MARKETPLACE_CONTAINER") == "true" {
+		return "PLACEHOLDER"
+	}
+	if os.Getenv("CQ_AWS_MARKETPLACE_AMI") == "true" {
+		// TODO: This is a placeholder product code, we need to get the real one from AWS Marketplace
```
erezrokah, line 272: Maybe read the placeholder from an env variable so we can test this without modifying the code? After we finish testing we can hard code our own value

premium/usage.go
```diff
@@ -307,6 +342,9 @@ func (u *BatchUpdater) TeamName() string {
 }
 
 func (u *BatchUpdater) HasQuota(ctx context.Context) (bool, error) {
+	if u.awsMarketPlaceClient != nil {
```
erezrokah, line 345: In AWS marketplace there's no concept of reaching a quota or a limit? You just get billed for more 💵 ?

premium/usage.go
```diff
@@ -420,6 +458,35 @@ func (u *BatchUpdater) updateUsageWithRetryAndBackoff(ctx context.Context, rows
 		u.logger.Debug().Str("url", u.url).Int("try", retry).Int("max_retries", u.maxRetries).Uint32("rows", rows).Msg("updating usage")
 		queryStartTime := time.Now()
 
+		// If the AWS Marketplace client is set, use it to track usage
+		if u.awsMarketPlaceClient != nil {
```
erezrokah, line 462: Maybe move this to `reportUsageToAWSMarketplace` as `updateUsageWithRetryAndBackoff` is getting quite long

premium/usage.go
```diff
@@ -420,6 +458,35 @@ func (u *BatchUpdater) updateUsageWithRetryAndBackoff(ctx context.Context, rows
...
+			// Timestamp + UsageDimension + UsageQuantity are required fields and must be unique
+			// since Timestamp only maintains a granularity of seconds, we need to ensure our batch size is large enough
+			_, err := u.awsMarketPlaceClient.MeterUsage(ctx, &marketplacemetering.MeterUsageInput{
```
erezrokah, line 465: Do we want to report usage per table (maybe put the table in the tags?) Not sure if that makes sense for AWS Marketplace

premium/usage.go
```diff
@@ -147,6 +151,8 @@ type BatchUpdater struct {
 	apiClient   *cqapi.ClientWithResponses
 	tokenClient TokenClient
 
+	awsMarketPlaceClient *marketplacemetering.Client
```
erezrokah, line 154:
```suggestion
	awsMarketplaceClient *marketplacemetering.Client
```

erezrokah (commented):
Looks good, added a questions and comments.
How can I test this?

premium/usage.go
```diff
@@ -200,6 +206,20 @@ func NewUsageClient(meta plugin.Meta, ops ...UsageClientOptions) (UsageClient, e
...
+	if isAWSMarketplace() {
+		cfg, err := awsConfig.LoadDefaultConfig(context.TODO())
+		if err != nil {
+			return nil, fmt.Errorf("failed to load AWS config: %w", err)
+		}
+
+		u.awsMarketPlaceClient = marketplacemetering.NewFromConfig(cfg)
+		u.teamName = "AWS_MARKETPLACE"
+		// This needs to be larger than normal, because we can only send a single usage record per second (from each compute node)
+		u.batchLimit = 1000000000
```
hermanschaaf, line 219: I'm not sure this is necessary, but would need to verify. I think how it should work is that there is a minimum flush time that should come into effect (by default this is 10s)

premium/usage.go
bbernays: Yes, there is no sort of concept of a limit or quota, it is a problem with AWS where you can't limit the spend of an account, which can lead to run away bills

premium/usage.go
```diff
@@ -240,6 +260,21 @@ func NewUsageClient(meta plugin.Meta, ops ...UsageClientOptions) (UsageClient, e
...
+
+func awsMarketplaceProductCode() string {
+	if os.Getenv("CQ_AWS_MARKETPLACE_CONTAINER") == "true" {
+		return "PLACEHOLDER"
+	}
+	if os.Getenv("CQ_AWS_MARKETPLACE_AMI") == "true" {
+		// TODO: This is a placeholder product code, we need to get the real one from AWS Marketplace
+		return "PLACEHOLDER"
+	}
+	return ""
```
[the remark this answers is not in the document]
bbernays: The AMI listing is taking longer, so I am just focusing on the Container to get it out as nearly all customers use containers already

premium/usage.go
bbernays: I think there is a race condition, if a batch is flushed and within the same second the sync ends then the timestamp will have the same value as timestamps are truncated to the second

premium/usage.go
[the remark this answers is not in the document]
bbernays: But I have added the product code for the container listing

premium/usage.go
bbernays: Done

premium/usage.go
bbernays: We are, we set it in the container we upload to the marketplace

premium/usage.go
bbernays: Have already updated it with the real code

premium/usage.go
bbernays: `context.TODO()` is supposed to be used when the surrounding function doesn't have a context available to use, so that is why I used it here. I am not sure the value of a timeout would add here.



I have moved

premium/usage.go
hermanschaaf: Ah, good catch... ok this should work for now but we should probably try and think of a better solution in the long term, otherwise we won't count any records if the sync fails

go.mod
```diff
@@ -2,8 +2,13 @@ module github.com/cloudquery/plugin-sdk/v4
 
 go 1.21.11
 
+replace github.com/cloudquery/plugin-sdk/v4 => ../../../../plugin-sdk
```
hermanschaaf, line 5: 🎗️ This will have to be removed before merging

erezrokah (commented):
Code looks great, how can I test this?

premium/usage.go
```diff
@@ -240,6 +251,31 @@ func NewUsageClient(meta plugin.Meta, ops ...UsageClientOptions) (UsageClient, e
...
+	u.awsMarketplaceClient = marketplacemetering.NewFromConfig(cfg)
+	u.teamName = "AWS_MARKETPLACE"
+	// This needs to be larger than normal, because we can only send a single usage record per second (from each compute node)
+	u.batchLimit = 1000000000
+	u.backgroundUpdater()
+	return nil
+}
+
+func isAWSMarketplace() bool {
+	return os.Getenv("CQ_AWS_MARKETPLACE_CONTAINER") == "true"
```
disq, line 269: Where these envs are being set? If it's something user settable (and not a published image of some sort) strconv.ParseBool would've been better, but I'm guessing it's an image?

premium/usage.go
```diff
@@ -41,6 +45,11 @@ const (
 	MaximumUpdateIntervalHeader = "x-cq-maximum-update-interval"
 )
 
+//go:generate mockgen -package=mocks -destination=../premium/mocks/marketplacemetering.go -source=usage.go AWSMarketplaceClientInterface
+type AWSMarketplaceClientInterface interface {
+	MeterUsage(ctx context.Context, params *marketplacemetering.MeterUsageInput, optFns ...func(*marketplacemetering.Options)) (*marketplacemetering.MeterUsageOutput, error)
+}
```
bbernays, line 51: This was added to ensure that we could mock the AWS API calls like we do the API Server cakks

premium/usage.go
```diff
@@ -240,6 +263,35 @@ func NewUsageClient(meta plugin.Meta, ops ...UsageClientOptions) (UsageClient, e
...
+	}
+	// This allows us to be able to inject a mock client for testing
+	if u.awsMarketplaceClient == nil {
+		u.awsMarketplaceClient = marketplacemetering.NewFromConfig(cfg)
+	}
+	u.teamName = "AWS_MARKETPLACE"
+	// This needs to be larger than normal, because we can only send a single usage record per second (from each compute node)
+	u.batchLimit = 1000000000
+
+	u.minTimeBetweenFlushes = 1 * time.Minute
```
hermanschaaf, line 279: Thinking about it now, I think a better solution than `u.batchLimit = 1000000000` would be to add a parameter like `minTimeSinceLastFlush`, set to `1s`, and then enforce this in the Close function, which should sleep until the min time threshold is reached.

This is non-blocking, but maybe we can open an issue for this kind of improvement

premium/usage.go
```diff
@@ -342,9 +397,28 @@ func (u *BatchUpdater) getTableUsage() (usage []cqapi.UsageIncreaseTablesInner,
...
 
+func (u *BatchUpdater) subtractTableUsageForAWSMarketplace(total uint32) {
+	for table := range u.tables {
+		tableTotal := u.tables[table]
+		if tableTotal < 1 {
+			continue
+		}
+		if tableTotal > total {
+			u.tables[table] -= total
+			total = 0
```
hermanschaaf, line 408: nit: we can add a `break` after setting `total = 0` and save some cycles

premium/usage.go
```diff
@@ -342,9 +397,28 @@ func (u *BatchUpdater) getTableUsage() (usage []cqapi.UsageIncreaseTablesInner,
 	return usage, u.rows
 }
 
+func (u *BatchUpdater) subtractTableUsageForAWSMarketplace(total uint32) {
+	for table := range u.tables {
+		tableTotal := u.tables[table]
+		if tableTotal < 1 {
+			continue
+		}
+		if tableTotal > total {
```
hermanschaaf, line 406:
```suggestion
		if tableTotal >= total {
```

premium/usage.go
```diff
@@ -575,3 +705,23 @@ func (NoOpUsageClient) IncreaseForTable(_ string, _ uint32) error {
...
+
+func roundUp(x, unit uint32) uint32 {
+	a := (x / unit) * unit
+	b := a + unit
+
+	if (x - a) < (b - x) {
+		return a
+	}
+	return b
+}
```
hermanschaaf, line 725:
```suggestion

func roundDown(x, unit uint32) uint32 {
	return x - (x % unit)
}

func roundUp(x, unit uint32) uint32 {
	if x%unit == 0 {
		return x
	}
	return x + (unit - x%unit)
}
```

I'm not sure why the `roundDown` and `roundUp` functions are so complicate? There also seems to be a bug in `roundUp`: for example, `roundUp(2001, 1000) == 2000`, but I'd expect `3000`

premium/usage.go
```diff
@@ -401,11 +481,19 @@ func (u *BatchUpdater) backgroundUpdater() {
 			case <-u.done:
```
hermanschaaf, line 481: Do we not need the `roundDown` logic in the `case <-u.flushDuration.C:` case as well?

premium/usage.go
bbernays: https://github.com/cloudquery/cloudquery-issues/issues/2177

PR #1796, opened 2024-07-17
feat: Add JSON type schema

erezrokah:
#### Summary

Part of https://github.com/cloudquery/cloudquery/issues/2023 ~~Still WIP and depends on https://github.com/cloudquery/cloudquery-api-go/pull/196~~

---

- [ ] Update or add tests 🧪

### DIFF
--- a/schema/column.go
+++ b/schema/column.go
@@ -84,6 +84,7 @@ func (c Column) ToArrowField() arrow.Field {
 		MetadataUnique:              MetadataFalse,
 		MetadataIncremental:         MetadataFalse,
 		MetadataPrimaryKeyComponent: MetadataFalse,
+		MetadataTypeSchema:          c.TypeSchema,
 	}
 	if c.PrimaryKey {
 		mdKV[MetadataPrimaryKey] = MetadataTrue
@@ -97,7 +98,6 @@ func (c Column) ToArrowField() arrow.Field {
 	if c.PrimaryKeyComponent {
 		mdKV[MetadataPrimaryKeyComponent] = MetadataTrue
 	}
-	mdKV[MetadataTypeSchema] = c.TypeSchema
 
 	return arrow.Field{
 		Name:     c.Name,
--- a/schema/column.go
+++ b/schema/column.go
@@ -47,6 +47,7 @@ type Column struct {
 	// PrimaryKeyComponent is a flag that indicates if the column is used as part of the input to calculate the value of `_cq_id`.
 	PrimaryKeyComponent bool `json:"primary_key_component"`
 
+	// If the column type is JSON, this field will have a JSON string that represents the schema of the JSON object.
 	TypeSchema string `json:"type_schema,omitempty"`
 }
 
--- a/schema/column.go
+++ b/schema/column.go
@@ -46,6 +46,8 @@ type Column struct {
 
 	// PrimaryKeyComponent is a flag that indicates if the column is used as part of the input to calculate the value of `_cq_id`.
 	PrimaryKeyComponent bool `json:"primary_key_component"`
+
+	TypeSchema string `json:"type_schema,omitempty"`
 }
 
 // NewColumnFromArrowField creates a new Column from an arrow.Field
@@ -70,6 +72,9 @@ func NewColumnFromArrowField(f arrow.Field) Column {
 	v, ok = f.Metadata.GetValue(MetadataPrimaryKeyComponent)
 	column.PrimaryKeyComponent = ok && v == MetadataTrue
 
+	v, _ = f.Metadata.GetValue(MetadataTypeSchema)
+	column.TypeSchema = v
+
 	return column
 }
 
@@ -92,6 +97,7 @@ func (c Column) ToArrowField() arrow.Field {
 	if c.PrimaryKeyComponent {
 		mdKV[MetadataPrimaryKeyComponent] = MetadataTrue
 	}
+	mdKV[MetadataTypeSchema] = c.TypeSchema
 
 	return arrow.Field{
 		Name:     c.Name,
@@ -111,6 +117,7 @@ func (c Column) MarshalJSON() ([]byte, error) {
 		Unique              bool   `json:"unique"`
 		IncrementalKey      bool   `json:"incremental_key"`
 		PrimaryKeyComponent bool   `json:"primary_key_component"`
+		TypeSchema          string `json:"type_schema,omitempty"`
 	}
 	var alias Alias
 	alias.Name = c.Name
@@ -121,6 +128,7 @@ func (c Column) MarshalJSON() ([]byte, error) {
 	alias.Unique = c.Unique
 	alias.IncrementalKey = c.IncrementalKey
 	alias.PrimaryKeyComponent = c.PrimaryKeyComponent
+	alias.TypeSchema = c.TypeSchema
 
 	return json.Marshal(alias)
 }
--- a/serve/package.go
+++ b/serve/package.go
@@ -75,7 +75,7 @@ func (s *PluginServe) writeTablesJSON(ctx context.Context, dir string) error {
 		}
 		columns := make([]cloudquery_api.PluginTableColumn, 0, len(table.Columns))
 		for _, column := range table.Columns {
-			columns = append(columns, cloudquery_api.PluginTableColumn{
+			c := cloudquery_api.PluginTableColumn{
 				Name:           column.Name,
 				Description:    column.Description,
 				Type:           column.Type.String(),
@@ -86,7 +86,12 @@ func (s *PluginServe) writeTablesJSON(ctx context.Context, dir string) error {
 				// 2. If the column is a `PrimaryKey` and both of the following are true column name is NOT `_cq_id`  and there are other columns that are a PrimaryKeyComponent
 				PrimaryKey: (column.PrimaryKey && !(column.Name == schema.CqIDColumn.Name && len(table.PrimaryKeyComponents()) > 0)) || column.PrimaryKeyComponent,
 				Unique:     column.Unique,
-			})
+			}
+			if column.TypeSchema != "" {
+				typeSchema := column.TypeSchema
+				c.TypeSchema = &typeSchema
+			}
+			columns = append(columns, c)
 		}
 		tablesToEncode = append(tablesToEncode, cloudquery_api.PluginTableCreate{
 			Description:   &table.Description,
--- a/transformers/struct.go
+++ b/transformers/struct.go
@@ -29,9 +29,6 @@ type structTransformer struct {
 	pkFieldsFound                 []string
 	pkComponentFields             []string
 	pkComponentFieldsFound        []string
-
-	currentJSONTypeSchemaDepth   int
-	useArrowNullForNilColumnType bool
 }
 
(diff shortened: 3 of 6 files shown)


serve/package.go
```diff
@@ -86,7 +86,12 @@ func (s *PluginServe) writeTablesJSON(ctx context.Context, dir string) error {
 				// 2. If the column is a `PrimaryKey` and both of the following are true column name is NOT `_cq_id`  and there are other columns that are a PrimaryKeyComponent
 				PrimaryKey: (column.PrimaryKey && !(column.Name == schema.CqIDColumn.Name && len(table.PrimaryKeyComponents()) > 0)) || column.PrimaryKeyComponent,
 				Unique:     column.Unique,
-			})
+			}
+			if column.TypeSchema != "" {
```
erezrokah, line 90: This is sent to our backend when uploading plugins versions

transformers/struct.go
```diff
@@ -106,22 +115,110 @@ func (t *structTransformer) ignoreField(field reflect.StructField) bool {
...
 	columnType, err := t.typeTransformer(field)
 	if err != nil {
-		return fmt.Errorf("failed to transform type for field %s: %w", field.Name, err)
+		return nil, fmt.Errorf("failed to transform type for field %s: %w", field.Name, err)
 	}
 
 	if columnType == nil {
 		columnType, err = DefaultTypeTransformer(field)
 		if err != nil {
-			return fmt.Errorf("failed to transform type for field %s: %w", field.Name, err)
```
erezrokah, line 122: Extracted this bit to `getColumnType`

transformers/struct.go
```diff
@@ -159,6 +256,11 @@ func (t *structTransformer) addColumnFromField(field reflect.StructField, parent
 		IgnoreInTests: t.ignoreInTestsTransformer(field),
 	}
 
+	// Avoid infinite recursion
+	if columnType == types.ExtensionTypes.JSON && t.jsonTypeSchemaDepth < maxJSONTypeSchemaDepth {
```
erezrokah, line 260: The diff is ugly but `addColumnFromField` didn't change much, only extracted `getColumnType` and added the condition with a call to `transformFieldToSchema` that handles getting the schema for the JSON type

schema/column.go
```diff
@@ -92,6 +97,7 @@ func (c Column) ToArrowField() arrow.Field {
 	if c.PrimaryKeyComponent {
 		mdKV[MetadataPrimaryKeyComponent] = MetadataTrue
 	}
+	mdKV[MetadataTypeSchema] = c.TypeSchema
```
erezrokah, line 100: We need to put the schema in the field metadata for the sake of the `cloudquery tables` command. Tables and fields get serialized over the wire (during the gRPC call), so if we want the `type_schema` to appear in the result of the tables command we need to send it via metadata

transformers/struct.go
```diff
@@ -106,24 +116,121 @@ func (t *structTransformer) ignoreField(field reflect.StructField) bool {
...
+	}
+	return reflect.New(field.Type)
+}
+
+func (t *structTransformer) transformFieldToSchema(field reflect.StructField) string {
+	transformInput := normalizePointer(field)
+	switch transformInput.Elem().Kind() {
+	case reflect.Struct:
+		table := &schema.Table{}
+		err := TransformWithStruct(
```
erezrokah, line 171: Tried to re-use our existing logic as much as possible

transformers/struct_test.go
```diff
@@ -480,3 +480,108 @@ func TestTableFromGoStruct(t *testing.T) {
...
+		testStruct any
+		want       map[string]string
+	}{
+		{
+			name: "simple map",
+			testStruct: struct {
+				Tags map[string]string `json:"tags"`
+			}{},
+			want: map[string]string{
+				"tags": "map<utf8, utf8, items_nullable>",
```
erezrokah, line 496: `items_nullable` might not be fully correct here, but I'm playing it safe

bbernays:
Adding @marianogappa as a reviewer to this because I don't have the bandwidth to dive into this and give it the proper level of attention

transformers/struct.go
```diff
@@ -106,24 +116,121 @@ func (t *structTransformer) ignoreField(field reflect.StructField) bool {
...
+			transformInput.Interface(),
+			WithNameTransformer(t.nameTransformer),
+			WithTypeTransformer(t.typeTransformer),
+			WithUnwrapAllEmbeddedStructs(),
+			withJSONTypeSchemaDepth(t.jsonTypeSchemaDepth+1),
```
marianogappa, line 176: There's no test that this is honoured.

Also, I find `withJSONTypeSchemaDepth` confusing. Since it's an option on a transformation function, it's unclear if it's documenting the current depth (I understand this is what it's doing) or setting the max allowed depth. The latter is what I'd expect for an option. I'd rename it to `withCurrentJSONTypeSchemaDepth`.

transformers/struct_test.go
```diff
@@ -480,3 +480,108 @@ func TestTableFromGoStruct(t *testing.T) {
...
+				Item struct {
+					Name   string `json:"name"`
+					Object any    `json:"object"`
+				} `json:"item"`
+			}{},
+			want: map[string]string{
+				"item": `{"name":"utf8","object":"any"}`,
+			},
+		},
+	}
```
marianogappa, line 566: I know I made a plugin where I use `TransformWithStruct` but I passthrough some json blobs on some fields. If I set the field as a `string` then I'll get `utf8` I guess, that's fair enough. But what if I set it to `map[string]interface{}`? There's no test for this case. It'd be good to show it doesn't blow up.

transformers/struct.go
erezrokah: Good idea I'll rename it and add a test for the depth

transformers/struct_test.go
erezrokah: I did add some code to handle `any/interface{}` and tests for it, but not specifically for a map with `any` values. I'll add it

schema/column.go
erezrokah: No reason, fixed in https://github.com/cloudquery/plugin-sdk/pull/1796/commits/d4044fc5ecc54531397808d87f1e3c13f19e3488

transformers/struct.go
erezrokah: Done in https://github.com/cloudquery/plugin-sdk/pull/1796/commits/d4044fc5ecc54531397808d87f1e3c13f19e3488

transformers/struct_test.go
erezrokah: Done in https://github.com/cloudquery/plugin-sdk/pull/1796/commits/d4044fc5ecc54531397808d87f1e3c13f19e3488

transformers/struct_test.go
```diff
@@ -480,3 +480,153 @@ func TestTableFromGoStruct(t *testing.T) {
...
+											Name string `json:"name"`
+										} `json:"level6"`
+									} `json:"level5"`
+								} `json:"level4"`
+							} `json:"level3"`
+						} `json:"level2"`
+					} `json:"level1"`
+				} `json:"level0"`
+			}{},
+			want: map[string]string{
```
erezrokah, line 603: Each level is a JSON encoded schema hence the quotes. Still looking if this can be improved

transformers/struct.go
```diff
@@ -111,17 +118,11 @@ func (t *structTransformer) addColumnFromField(field reflect.StructField, parent
 		return nil
 	}
 
-	columnType, err := t.typeTransformer(field)
+	columnType, err := t.getColumnType(field)
```
erezrokah, line 121: Extracted `getColumnType` to a dedicated function

erezrokah:
Sorry for the review noise, I basically re-did the PR without using `TransformWithStruct` as it didn't provide me the output I wanted from the schema. Please see updated tests that I believe look better now and always return a valid JSON

transformers/struct.go
```diff
@@ -159,6 +160,11 @@ func (t *structTransformer) addColumnFromField(field reflect.StructField, parent
 		IgnoreInTests: t.ignoreInTestsTransformer(field),
 	}
 
+	// Avoid infinite recursion
```
marianogappa, line 163: I think this comment made more sense before. Now it should be inside the `fieldToJSONSchema` function, if it's necessary.

transformers/struct.go
```diff
@@ -233,3 +239,99 @@ func TransformWithStruct(st any, opts ...StructTransformerOption) schema.Transfo
...
+			}
+			columnType, err := t.getColumnType(fieldType.Field(i))
+			if err != nil {
+				continue
+			}
+			if columnType == nil {
+				fieldsMap[name] = "any"
+				continue
+			}
+			if columnType == types.ExtensionTypes.JSON && depth < maxJSONTypeSchemaDepth {
```
marianogappa, line 292: I think the comment goes here

transformers/struct.go
```diff
@@ -233,3 +239,99 @@ func TransformWithStruct(st any, opts ...StructTransformerOption) schema.Transfo
...
+				fieldsMap[name] = t.fieldToJSONSchema(fieldType.Field(i), depth+1)
+				continue
+			}
+			if arrow.IsListLike(columnType.ID()) {
+				fieldsMap[name] = []any{columnType.(*arrow.ListType).Elem().String()}
```
marianogappa, line 297: Can this typecast ever be `!ok`? I guess I'm asking if a `columnType` that is list like cannot be cast to `*ListType`.

According to godoc:

>// IsListLike returns true for List, LargeList, FixedSizeList, and Map

I wonder if `Map` is an interesting edge case for `.Elem().String()`?

transformers/struct_test.go
```diff
@@ -480,3 +480,149 @@ func TestTableFromGoStruct(t *testing.T) {
...
+				"level0": `{"level1":{"level2":{"level3":{"level4":{"level5":{"level6":"json"}}}}}}`,
```
marianogappa, line 604: ✅

transformers/struct_test.go
```diff
@@ -480,3 +480,149 @@ func TestTableFromGoStruct(t *testing.T) {
...
+			},
+		},
+	}
```
marianogappa, line 607: How about a case for the arrow `ListLike` thing?

marianogappa (approved):
Some comments

transformers/struct.go
erezrokah: So we're the ones creating the list type here https://github.com/cloudquery/plugin-sdk/blob/7955c5358f1360bb0236fc223b490685c7ea5c31/transformers/type.go#L56 and we don't create other types.
However, someone might override the default transformer. I made the condition safer in https://github.com/cloudquery/plugin-sdk/pull/1796/commits/a59d1f7b87e9f7206eeabad80a4933c07a3ac24d

transformers/struct_test.go
erezrokah: It's covered by the array fields, for example `FlatItems` in https://github.com/cloudquery/plugin-sdk/pull/1796/files#diff-c5baf01f105731f0f9bff5c836860e3a0816cb9fbcedd3fd28e9d1cd608dd398R527

marianogappa (approved):

PR #1800, opened 2024-07-18
feat: Implement transformations logic.

marianogappa:
This is a first draft of sdk changes to add transformations support, so that transformer plugins can be created.

I don't want to merge it yet, because the channel logic is not well thought out. I'd rather want to do a test sync locally with some transformation.


internal/reversertransformer/reversertransformer_test.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+var mem = memory.NewGoAllocator()
+
+func TestReverserTransformer(t *testing.T) {
+	p := plugin.NewPlugin("test", "development", GetNewClient())
+	p.SetSkipTableValidation(true)
```
marianogappa, line 23: This is annoying 👀  a plugin doesn't know what it does, so it doesn't know that it has to automatically skip this validation if it's a transformer plugin. I don't like that whoever implements a Transformer plugin has to remember to skip this.

erezrokah (commented):
This looks good 🚀 Thank you for adding the tests 👍 

Will wait for an example plugin in our monorepo to approve

marianogappa:
I replaced this sdk onto the monorepo and added preliminary support for transformers in syncs, and I added a "reverser" transformer. Here's the result of hooking it up on an xkcd sync:

Screenshot 2024-07-19 at 12 59 06

erezrokah (approved):

PR #1821, opened 2024-07-25
feat: Add `zerolog.Logger` to `retryablehttp.LeveledLogger` adapter struct

erezrokah:
#### Summary

This allows plugins that use https://github.com/hashicorp/go-retryablehttp to easily log retries as plugins use `github.com/rs/zerolog`

---

- [ ] Update or add tests 🧪

### DIFF
--- a/helpers/retryablehttp/logger.go
+++ b/helpers/retryablehttp/logger.go
@@ -9,19 +9,19 @@ type leveledLogger struct {
 	zerolog.Logger
 }
 
-func (l *leveledLogger) Debug(msg string, keysAndValues ...interface{}) {
+func (l *leveledLogger) Debug(msg string, keysAndValues ...any) {
 	l.Logger.Debug().Fields(keysAndValues).Msg(msg)
 }
 
-func (l *leveledLogger) Error(msg string, keysAndValues ...interface{}) {
+func (l *leveledLogger) Error(msg string, keysAndValues ...any) {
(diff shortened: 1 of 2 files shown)


helpers/retryablehttp/logger.go
```diff
@@ -0,0 +1,30 @@
...
+
+func (l *leveledLogger) Info(msg string, keysAndValues ...any) {
+	l.Logger.Info().Fields(keysAndValues).Msg(msg)
+}
+
+func (l *leveledLogger) Warn(msg string, keysAndValues ...any) {
+	l.Logger.Warn().Fields(keysAndValues).Msg(msg)
+}
+
+func NewLeveledLogger(logger zerolog.Logger) retryablehttp.LeveledLogger {
```
marianogappa, line 28: You still have to instantiate this everywhere that you use the retryable, right?

marianogappa (approved):

helpers/retryablehttp/logger.go
erezrokah: Yes, I initially wrapped `retryablehttp` and did `NewRetryableClient` but then realized I would have to add all the different options as well (as some plugins modify them), so I think sharing only the logger is better

erezrokah: We have something similar in https://github.com/cloudquery/plugin-sdk/blob/5c77cee87d9fca292e9e81663c9ce3775962a623/helpers/grpczerolog/grpc.go#L12

PR #1828, opened 2024-07-29
feat: Add PluginKind "transformer".

marianogappa:
Add PluginKind's "transformer", which is necessary to enable the transformation's work.


erezrokah (approved):

erezrokah (approved):

erezrokah:
You'll need to run `go mod tidy` inside https://github.com/cloudquery/plugin-sdk/tree/main/examples/simple_plugin to fix the tests

PR #1835, opened 2024-07-30
fix: Handle commas in permissions array

erezrokah:
#### Summary

Follow up to https://github.com/cloudquery/plugin-sdk/pull/1827#discussion_r1696683944

---

- [ ] Update or add tests 🧪

### DIFF
--- a/schema/table.go
+++ b/schema/table.go
@@ -2,10 +2,10 @@ package schema
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"regexp"
 	"slices"
-	"strings"
 
 	"github.com/apache/arrow/go/v17/arrow"
 	"github.com/cloudquery/plugin-sdk/v4/glob"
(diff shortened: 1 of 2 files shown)


schema/table.go
```diff
@@ -182,14 +182,17 @@ func NewTableFromArrowSchema(sc *arrow.Schema) (*Table, error) {
 	for i, field := range fields {
 		columns[i] = NewColumnFromArrowField(field)
 	}
+
+	var permissionsNeededArr []string
+	_ = json.Unmarshal([]byte(permissionsNeeded), &permissionsNeededArr)
```
erezrokah, line 187: Don't think `Marshal`, `Unmarshal` can fail for `[]string` and if it does I don't think it should panic for a doc related feature

erezrokah: I mean it can fail if the string is not a JSON array but we're the ones setting it

marianogappa: Can it fail if it's not set?

marianogappa: Is that a good test case?

erezrokah: If it's not set `nil` array gets serialized to `null` I'll add a test for it

erezrokah: Added https://github.com/cloudquery/plugin-sdk/pull/1835/commits/baf5f303c65d2f9680c0641248581cacc78c3144

`nil` columns are always recreated as non `nil` due to https://github.com/cloudquery/plugin-sdk/blob/341c54641fc2de65c23287276989d9aebca2aecd/schema/table.go#L181

marianogappa (approved):

PR #1862, opened 2024-08-08
feat: Add basic fuzzing support

erezrokah:
#### Summary

A very crude and simple way to do https://github.com/cloudquery/cloudquery-issues/issues/1846.
Adds a new hidden fuzz scheduler that only multiplies the clients (at the moment).
The code is based on the shuffle scheduler then adds then duplicates client based on the multiplier.

See example:

```
cloudquery sync examples/pagerduty-postgres.yml

Loading spec(s) from examples/pagerduty-postgres.yml
Starting sync for: pagerduty (local@/Users/erezrokah/code/github/cloudquery/cloudquery-private/plugins/source/pagerduty/pagerduty) -> [postgresql (cloudquery/postgresql@v8.2.7)]
Sync completed successfully. Resources: 525, Errors: 0, Warnings: 0, Time: 7s
```

```
CQ_DEBUG_FUZZ_MULTIPLIER=50 cloudquery sync examples/pagerduty-postgres.yml

Loading spec(s) from examples/pagerduty-postgres.yml
Starting sync for: pagerduty (local@/Users/erezrokah/code/github/cloudquery/cloudquery-private/plugins/source/pagerduty/pagerduty) -> [postgresql (cloudquery/postgresql@v8.2.7)]
Sync completed successfully. Resources: 26385, Errors: 0, Warnings: 0, Time: 2m12s
```

This has a couple of downsides/tradeoffs
1. There will be clients with duplicate IDs which breaks the metrics counts https://github.com/cloudquery/plugin-sdk/blob/25ed3d25a529a22f351ab92e22fb03a19c9557d4/scheduler/metrics.go#L144
2. If a plugin uses the client ID to ensure uniqueness for state client keys, that logic will break too
3. If a table doesn't have any resources the impact of the multiplier will be lower

However I think this is still useful if we want to artificially make a sync large (e.g. simulate a sync on many AWS accounts)

---

- [ ] Update or add tests 🧪

### DIFF
--- a/scheduler/scheduler_fuzz.go
+++ b/scheduler/scheduler_fuzz.go
@@ -1,0 +1,94 @@
+package scheduler
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"strconv"
+	"sync"
+
+	"github.com/cloudquery/plugin-sdk/v4/schema"
+)
+
+const (
+	// This is an environment variable and not a spec option in each plugin to make it easier to enable it
+	cqDebugFuzzMultiplier = "CQ_DEBUG_FUZZ_MULTIPLIER"
+)
+
+func getFuzzMultiplier() (int, error) {
+	strValue, ok := os.LookupEnv(cqDebugFuzzMultiplier)
+	if ok {
+		intValue, err := strconv.Atoi(strValue)
+		if err != nil {
+			return 0, fmt.Errorf("failed to parse %s=%s as integer: %w", cqDebugFuzzMultiplier, strValue, err)
+		}
+		return intValue, nil
+	}
+	return 0, nil
+}
+
+func (s *syncClient) syncFuzz(ctx context.Context, fuzzMultiplier int, resolvedResources chan<- *schema.Resource) {
+	// we have this because plugins can return sometimes clients in a random way which will cause
+	// differences between this run and the next one.
+	preInitialisedClients := make([][]schema.ClientMeta, len(s.tables))
+	tableNames := make([]string, len(s.tables))
+	for i, table := range s.tables {
+		tableNames[i] = table.Name
+		clients := []schema.ClientMeta{s.client}
+		if table.Multiplex != nil {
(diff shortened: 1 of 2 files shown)


scheduler/scheduler_fuzz.go
```diff
@@ -0,0 +1,94 @@
...
+		seenClients := make(map[string]bool)
+		for _, c := range clients {
+			if _, ok := seenClients[c.ID()]; !ok {
+				seenClients[c.ID()] = true
+			} else {
+				s.logger.Warn().Str("client", c.ID()).Str("table", table.Name).Msg("multiplex returned duplicate client")
+			}
+		}
+		preInitialisedClients[i] = clients
+		// we do this here to avoid locks so we initial the metrics structure once in the main goroutines
```
marianogappa, line 51:
```suggestion
		// we do this here to avoid locks so we initialize the metrics structure once in the main goroutine
```

marianogappa (approved):
I don't think `scheduler_fuzz` is a good name for this. I did the exercise of looking at the code without looking at the issue first, and it made no sense to me what the point of this scheduler was 🤔 

It's more of a load test scheduler 🤔 It doesn't do fuzzing.

I think it would pay to add a small comment in the constructor of what problem it solves.

erezrokah:
> I don't think `scheduler_fuzz` is a good name for this. I did the exercise of looking at the code without looking at the issue first, and it made no sense to me what the point of this scheduler was 🤔
> 
> It's more of a load test scheduler 🤔 It doesn't do fuzzing.

💯 Happy to rename it. I don't like the fuzz name as well (though it might do some fuzzing in the future). I'll rename

erezrokah:
> 💯 Happy to rename it. I don't like the fuzz name as well (though it might do some fuzzing in the future). I'll rename

Done in https://github.com/cloudquery/plugin-sdk/pull/1862/commits/08bb4a04ba738cd8c866b2fa59d60be9b53056e7

PR #1866, opened 2024-08-11
feat: Add RemoteOAuth Token helper to refresh `access_token` from cloud environment

disq:
Implements https://github.com/cloudquery/cloudquery-issues/issues/1978 (internal issue)


auth/options.go
```diff
@@ -0,0 +1,87 @@
+package auth
```
disq, line 1: Not sure about the package name? `oauth`? `remoteoauth`?

auth/options.go
```diff
@@ -0,0 +1,87 @@
...
+
+func WithAccessToken(token, tokenType string, expiry time.Time) RemoteOAuthTokenOption {
+	return func(t *RemoteOAuthToken) {
+		t.AccessToken = token
+		t.TokenType = tokenType
+		t.Expiry = expiry
+	}
+}
+
+func WithCloudEnv() RemoteOAuthTokenOption {
```
disq, line 22: This can be the default/implicit as it's the only reason to use this package and won't be enabled unless `CQ_CLOUD` env var exists. If we did that some of the ID/UUID confusion and post-opt-validate code can be simplified.

disq: Done in https://github.com/cloudquery/plugin-sdk/pull/1866/commits/66d2cabdf8385d4e92b9e1de7f6f9281a12a3d70

auth/options.go
disq: Moved to `helpers/remoteoauth.Token`

helpers/remoteoauth/token.go
```diff
@@ -0,0 +1,165 @@
...
+func (t *Token) Valid() bool {
+	return t != nil && t.AccessToken != "" && !t.expired()
+}
+
+// timeNow is time.Now but pulled out as a variable for tests.
+var timeNow = time.Now
+
+// expired reports whether the token is expired.
+// t must be non-nil.
+func (t *Token) expired() bool {
```
marianogappa, line 111: Don't you think `isExpired` (i.e. using `is` or `are` for booleans) is more readable as a convention?

helpers/remoteoauth/token.go
```diff
@@ -0,0 +1,165 @@
...
+		}
+		oauthResp = resp.JSON200.Oauth
+	}
+
+	if oauthResp == nil {
+		return fmt.Errorf("missing oauth credentials in response")
+	}
+	t.remoteToken = true
+	t.AccessToken = oauthResp.AccessToken
+	if oauthResp.Expires == nil {
```
marianogappa, line 160:
```go
t.Expiry = time.Time{}
if oauthResp.Expires != nil {
    t.Expiry = *oauthResp.Expires
}
```

helpers/remoteoauth/token.go
```diff
@@ -0,0 +1,165 @@
...
+	"sync"
+	"time"
+
+	cloudquery_api "github.com/cloudquery/cloudquery-api-go"
+	"github.com/google/uuid"
+	"golang.org/x/oauth2"
+)
+
+type Token struct {
+	AccessToken string
```
marianogappa: These 3 first properties are logically grouped as a `oauth2.Token` and returned as such. Maybe it makes sense for the `Token` struct to just have a `oauth2.Token` inside, instead of the 3 properties.

helpers/remoteoauth/token.go
```diff
@@ -0,0 +1,165 @@
...
+
+	return &oauth2.Token{
+		AccessToken: t.AccessToken,
+		TokenType:   t.TokenType,
+		Expiry:      t.Expiry,
+	}, nil
+}
+
+// Valid reports whether t is non-nil, has an AccessToken, and is not expired.
+func (t *Token) Valid() bool {
```
marianogappa, line 98: Don't you think isValid (i.e. using is or are for booleans) is more readable as a convention?

If not ignore the other similar comment.

helpers/remoteoauth/token.go
```diff
@@ -0,0 +1,166 @@
...
+	"sync"
+	"time"
+
+	cloudquery_api "github.com/cloudquery/cloudquery-api-go"
+	"github.com/google/uuid"
+	"golang.org/x/oauth2"
+)
+
+type Token struct {
+	accessToken string
```
marianogappa, line 17: This 3 properties are logically grouped as an `oauth.Token`. Maybe `Token` should have an `oauth.Token` inside?

marianogappa (approved):
Just optional comments. Originally I had more to say but suddenly `TokenWithContext` got simplified, and the tests I was needed magically appeared 🤣

helpers/remoteoauth/token.go
disq: `hasExpired`? I'm guessing original Go authors also had the same conundrum.

helpers/remoteoauth/token.go
disq: This convention is coming from golang.org/x/oauth2/oauth2.go it's not part of the TokenSource interface it's still a public method.

helpers/remoteoauth/token.go
disq: Done in https://github.com/cloudquery/plugin-sdk/pull/1866/commits/75d68c7e2b44f1179d28514c6106bfd082415d3b

helpers/remoteoauth/token.go
disq: Fixed in https://github.com/cloudquery/plugin-sdk/pull/1866/commits/75d68c7e2b44f1179d28514c6106bfd082415d3b

helpers/remoteoauth/token.go
```diff
@@ -0,0 +1,170 @@
...
+// provided the current token is not acquired from a remote source.
+func (t *Token) Valid() bool {
+	t.tokenMu.RLock()
+	defer t.tokenMu.RUnlock()
+
+	if t.cloudEnabled && !t.remoteToken {
+		return false
+	}
+
+	return t != nil && t.currentToken.AccessToken != "" && !t.expired()
```
disq, line 104: Here we could've just done `return t.currentToken.Valid()` (and that would also remove `expired()`) but we need access to `timeNow` to easily test expiry.

helpers/remoteoauth/token.go
marianogappa: Either is fine

helpers/remoteoauth/token.go
marianogappa: TIL 👍

helpers/remoteoauth/tokenoptions.go
```diff
@@ -13,9 +13,9 @@ type TokenOption func(*Token)
 
 func WithAccessToken(token, tokenType string, expiry time.Time) TokenOption {
 	return func(t *Token) {
-		t.accessToken = token
-		t.tokenType = tokenType
-		t.expiry = expiry
+		t.currentToken.AccessToken = token
+		t.currentToken.TokenType = tokenType
+		t.currentToken.Expiry = expiry
```
marianogappa, line 18:
```go
t.currentToken = oauth.Token{
  AccessToken = token,
  TokenToke = tokenType,
  Expiry = expiry,
}
```

marianogappa (approved):

marianogappa (approved):

helpers/remoteoauth/tokenoptions.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+}
+
+func (t *Token) initCloudOpts() error {
+	_, t.cloudEnabled = os.LookupEnv("CQ_CLOUD")
+	if !t.cloudEnabled {
+		return nil
+	}
+
+	t.apiToken = os.Getenv("CLOUDQUERY_API_KEY")
```
erezrokah, line 31: Can we do `errors.join` and report all possible errors and not just the first one?

Maybe also put all the validation code in a dedicated function (non blocking)

helpers/remoteoauth/token.go
```diff
@@ -0,0 +1,170 @@
...
+	"net/http"
+	"sync"
+	"time"
+
+	cloudquery_api "github.com/cloudquery/cloudquery-api-go"
+	"github.com/google/uuid"
+	"golang.org/x/oauth2"
+)
+
+type Token struct {
```
erezrokah, line 16: `oauth2` has a `ReuseTokenSource` (or `ReuseTokenSourceWithExpiry`) where you can pass the initial token and a interface implementation of `Token()`.
That will save us from implementing all the locking involved, valid and expiry logic

helpers/remoteoauth/token.go
```diff
@@ -0,0 +1,170 @@
...
+			return nil, fmt.Errorf("failed to create api client: %w", err)
+		}
+	}
+
+	return t, nil
+}
+
+// Token returns the cached token if not expired, or a new token from the remote source.
+func (t *Token) Token() (*oauth2.Token, error) {
+	return t.TokenWithContext(context.TODO())
```
erezrokah, line 71: Is this something we should support? What happens if we want to cancel the context/request?

erezrokah (commented):
Nice helper to have, added a few comments

helpers/remoteoauth/token.go
disq: All token-refreshing methods in `oauth2` require client credentials to work and they use a `refresh_token` to do so. They are not usable for this use case on their own.

disq: As another point, we want to default to the client-provided token only if cloud env is not set (because plugins don't have the `expires_at` field in their spec as they assume the user-provided token would be valid for the lifetime of the plugin). That's why the `remoteToken` exists, if cloud env is set we override the spec-specified token with a cloud one and can have the actual expiry timestamp.

helpers/remoteoauth/token.go
disq: This is the interface of `oauth2.TokenSource`. If we don't support this, this can't be swapped in for plugins which use the normal/standard oauth2 package methods.

helpers/remoteoauth/tokenoptions.go
disq: Errors joined in https://github.com/cloudquery/plugin-sdk/pull/1866/commits/6613af3c564b6f1c38d90fed7a4b64af0bf7406d

`initCloudOpts` is already simple enough, I don't think it's beneficial to separate it even more.

helpers/remoteoauth/token.go
disq: ... I'll check if something can be done within the limitations of the things I've outlined in the above two comments.

disq: I don't think it saves us from anything, the code is simple and about the same as the official implementation. Wrapping it in another `ReuseTokenSource` will just make it complicated. Furthermore, removing `expired` will remove `timeNow` access and will make testing harder or slower.

erezrokah: > All token-refreshing methods in `oauth2` require client credentials to work and they use a `refresh_token` to do so. They are not usable for this use case on their own.

Looks like `ReuseTokenSource` doesn't use a refresh token it calls `.Token()` on the provided src interface https://github.com/golang/oauth2/blob/b52af7d5b4e39d5bb1ee067d8aa110fcce9e4cc7/oauth2.go#L310

You can even pass the initial token as `nil` if you'd like, e.g. `ReuseTokenSource(nil, customTokenSourceThatGetsNewTokenFromCloud)`

erezrokah: Refresh token logic is here https://github.com/golang/oauth2/blob/b52af7d5b4e39d5bb1ee067d8aa110fcce9e4cc7/oauth2.go#L250, it creates a `tokenRefresher` (which is an implementation of `TokenSource interface`). 
What I'm suggesting is that we do the same with our own `TokenSource interface` implementation

erezrokah: > if cloud env is not set (because plugins don't have the `expires_at` field in their spec as they assume the user-provided token would be valid for the lifetime of the plugin)

We can:
1. Set the expiry to a very long time in the future not in a cloud environment 
2. Have our own `TokenSource interface` always return the client provided token not in a cloud environment

disq: Done in https://github.com/cloudquery/plugin-sdk/pull/1866/commits/3e909a8e59c2937b26b230f83fd532cd9df538c1

helpers/remoteoauth/token.go
```diff
@@ -0,0 +1,186 @@
...
+)
+
+func NewTokenSource(opts ...tokenSourceOption) (oauth2.TokenSource, error) {
+	t := &tokenSource{}
+	for _, opt := range opts {
+		opt(t)
+	}
+
+	if _, cloudEnabled := os.LookupEnv("CQ_CLOUD"); !cloudEnabled {
+		return oauth2.StaticTokenSource(&t.currentToken), nil
```
erezrokah, line 23: Nice touch 🚀

helpers/remoteoauth/token.go
```diff
@@ -0,0 +1,186 @@
...
+	return t, nil
+}
+
+// Token returns the cached token if not expired, or a new token from the remote source.
+func (t *cloudTokenSource) Token() (*oauth2.Token, error) {
+	return t.TokenWithContext(context.TODO())
+}
+
+// TokenWithContext returns a new token from the remote source using the given context.
+func (t *cloudTokenSource) TokenWithContext(ctx context.Context) (*oauth2.Token, error) {
```
erezrokah, line 84: This is not part of the TokenSource interface I think, and you shouldn't need the locking here since `reuseTokenSource` already has it.

If you need to pass the `context` I think you can do it as an option for `NewTokenSource`

disq: done in https://github.com/cloudquery/plugin-sdk/pull/1866/commits/2600aeebdc288826d9c850ca97f38a3fcc2cba4f

PR #1869, opened 2024-08-14
feat: Add queue based scheduler

erezrokah:
#### Summary

Mostly an experiment to deal with https://github.com/cloudquery/cloudquery-issues/issues/2227 as I couldn't think of a nice way to make `singleNestedTableMaxConcurrency` dynamic without making the code super complex.

This PR adds a scheduler that uses a worker pool pattern on top of a priority queue. This should ensure that as long as there's work to be done, all Go routines will be occupied. Also the `concurrency` setting is not only for the top level tables, as it's the number of workers so it's a fixed limit and simpler.
The more table client pairs in the queue the less priority they'll have, this should prevent a specific table from occupying all the workers.

Opening as draft since:
1. I'm still testing this to see the impact
2. There's still a lot of code duplication with current code and refactoring needed to avoid it

---

- [ ] Update or add tests 🧪

### DIFF
--- a/scheduler/metrics/metrics.go
+++ b/scheduler/metrics/metrics.go
@@ -1,4 +1,4 @@
-package scheduler
+package metrics
 
 import (
 	"context"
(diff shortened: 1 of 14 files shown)


erezrokah:
Closing in favor of https://github.com/cloudquery/plugin-sdk/pull/1914

PR #1875, opened 2024-08-23
feat: Add remoteoauth helpers (`TokenAuthTransport` and `TokenAuthEditor`)

disq:
These helpers can be used to inject dynamic tokens into requests.

`TokenAuthTransport` is a `http.RoundTripper`, with optional custom rewriting capability
`TokenAuthEditor(oauth2.TokenSource)` returns a `func(context.Context, *http.Request) error` to be used as a `RequestEditorFn` in openapi clients.

### DIFF
--- a/helpers/remoteoauth/tokenautheditor.go
+++ b/helpers/remoteoauth/tokenautheditor.go
@@ -1,0 +1,26 @@
+package remoteoauth
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	"golang.org/x/oauth2"
+)
+
+// TokenAuthEditor returns a custom RequestEditorFn to inject the OAuth2 token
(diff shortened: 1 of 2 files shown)


helpers/remoteoauth/tokenautheditor.go
```diff
@@ -0,0 +1,26 @@
...
+	"context"
+	"fmt"
+	"net/http"
+
+	"golang.org/x/oauth2"
+)
+
+// TokenAuthEditor returns a custom RequestEditorFn to inject the OAuth2 token
+func TokenAuthEditor(tokenSource oauth2.TokenSource) func(context.Context, *http.Request) error {
+	return func(_ context.Context, req *http.Request) error {
```
disq, line 13: Unfortunately we can't use the `ctx` here for the `tokenSource.Token()` call.

erezrokah (approved):

PR #1889, opened 2024-09-04
feat: Support Contract Listing For AWS Marketplace

bbernays:
#### Summary

Implements https://catalog.workshops.aws/mpseller/en-US/container/integrate-contract


premium/offline.go
```diff
@@ -41,20 +48,83 @@ var publicKey string
...
+}
+
+func WithAWSLicenseManagerClient(awsLicenseManagerClient AWSLicenseManagerInterface) LicenseClientOptions {
+	return func(cl *CQLicenseClient) {
+		cl.awsLicenseManagerClient = awsLicenseManagerClient
+	}
+}
+
+func NewLicenseClient(ctx context.Context, logger zerolog.Logger, ops ...LicenseClientOptions) (CQLicenseClient, error) {
+	cl := &CQLicenseClient{
```
disq, line 84: Why create a ptr and then dereference it during return? do `cl := CQLicenseClient{...` and then `op(&cl)` instead?

bbernays: Implemented in [2d9b605](https://github.com/cloudquery/plugin-sdk/pull/1889/commits/2d9b6050b54c775e0137cb154701e2a8ea027a22)

erezrokah (commented):
Looks good, had one comment

premium/offline.go
```diff
@@ -80,6 +80,10 @@ func WithAWSLicenseManagerClient(awsLicenseManagerClient AWSLicenseManagerInterf
 	}
 }
 
+func isMarketplaceLicense() bool {
```
disq, line 83: since os.Getenv wouldn't return different things, I would add this as a`bool` in `CQLicenseClient` and initialize it on the same line:
```go
cl := CQLicenseClient{
		logger: logger,
		isMarketplaceLicense: os.Getenv("CQ_AWS_MARKETPLACE_LICENSE") == "true",
...
}
```

disq: You could also keep this as a local `bool` (and init the aws client on that), and depend on the fact that `cl. awsLicenseManagerClient` being `nil` or not inside `ValidateLicense()`.

premium/offline.go
```diff
@@ -41,20 +48,87 @@ var publicKey string
...
+		if err != nil {
+			return cl, fmt.Errorf("failed to load AWS config: %w", err)
+		}
+		cl.awsLicenseManagerClient = licensemanager.NewFromConfig(cfg)
+	}
+
+	return cl, nil
+}
+
+func (lc CQLicenseClient) ValidateLicense(ctx context.Context) error {
```
disq, line 105: Also this could be rewritten as a switch (or if elseif) statement and leave the `return ErrLicenseNotApplicable` as the last (or `default` in `switch`) option.

premium/offline.go
```diff
@@ -41,20 +48,87 @@ var publicKey string
...
+	// License can be provided via environment variable for AWS Marketplace or CLI flag
+	switch {
+	case lc.isMarketplaceLicense:
+		return lc.validateMarketplaceLicense(ctx)
+	case lc.licenseFileOrDirectory != "":
+		lc.validateCQLicense()
+	default:
+		return ErrLicenseNotApplicable
+	}
+
```
disq, line 115:
```suggestion
```

premium/offline.go
```diff
@@ -41,20 +48,86 @@ var publicKey string
...
+	return cl, nil
+}
+
+func (lc CQLicenseClient) ValidateLicense(ctx context.Context) error {
+	// License can be provided via environment variable for AWS Marketplace or CLI flag
+	switch {
+	case lc.isMarketplaceLicense:
+		return lc.validateMarketplaceLicense(ctx)
+	case lc.licenseFileOrDirectory != "":
+		lc.validateCQLicense()
```
disq, line 111:
```suggestion
		return lc.validateCQLicense()
```

premium/offline.go
```diff
@@ -158,3 +231,28 @@ func (l *License) IsValid(logger zerolog.Logger) error {
...
 	return nil
 }
+
+func (lc CQLicenseClient) validateMarketplaceLicense(ctx context.Context) error {
+	clientToken := uuid.New()
+
+	resp, err := lc.awsLicenseManagerClient.CheckoutLicense(ctx, &licensemanager.CheckoutLicenseInput{
+		CheckoutType: types.CheckoutTypeProvisional,
+		ClientToken:  aws.String(clientToken.String()),
+		ProductSKU:   aws.String("55ukc0d5qv3gebks148tjr62j"),
```
disq, line 241: maybe (fn-level or unexported) const, with a name?

premium/offline_test.go
```diff
@@ -153,12 +162,76 @@ func licenseTest(inputPath string, meta plugin.Meta, timeIs time.Time, expectErr
...
 }
+
+func TestValidateMarketplaceLicense(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	m := mocks.NewMockAWSLicenseManagerInterface(ctrl)
+	out := licensemanager.CheckoutLicenseOutput{}
+	in := licenseInput{
+		CheckoutLicenseInput: licensemanager.CheckoutLicenseInput{
+			CheckoutType: types.CheckoutTypeProvisional,
+			ProductSKU:   aws.String("55ukc0d5qv3gebks148tjr62j"),
```
disq, line 183: could've used the const here too.

bbernays: [f2d4386](https://github.com/cloudquery/plugin-sdk/pull/1889/commits/f2d4386e997ebd8dc4bfb67cbb2d37d18d13287e)

disq (approved):

erezrokah (approved):

PR #1891, opened 2024-09-11
feat: Add sync sharding

erezrokah:
#### Summary

Goes with https://github.com/cloudquery/plugin-pb-go/pull/401. ~~Still testing this so in draft~~

Part of https://github.com/cloudquery/cloudquery-issues/issues/2214 (internal issue)

---

- [ ] Update or add tests 🧪


examples/simple_plugin/go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/cloudquery/plugin-sdk/examples/simple_plugin
```
erezrokah, line 1: These changes can be reverted once https://github.com/cloudquery/plugin-pb-go/pull/401 is merged and released

scheduler/scheduler_dfs.go
```diff
@@ -40,27 +40,34 @@ func (s *syncClient) syncDfs(ctx context.Context, resolvedResources chan<- *sche
 		s.metrics.initWithClients(table, clients)
 	}
 
-	var wg sync.WaitGroup
+	tableClients := make([]tableClient, 0)
```
erezrokah, line 43: Historically the DFS scheduler didn't need to create table clients pairs, since we didn't do any sorting in the DFS scheduler. Because of the sharding support, we need to first the table client pairs, so we can shard them before the sync starts

scheduler/scheduler_shuffle.go
```diff
@@ -45,6 +45,7 @@ func (s *syncClient) syncShuffle(ctx context.Context, resolvedResources chan<- *
 	// so users have a little bit of control over the randomization.
 	seed := hashTableNames(tableNames)
 	shuffle(tableClients, seed)
+	tableClients = shardTableClients(tableClients, s.shard)
```
marianogappa, line 48: I do see that `shuffle` is deterministic (at the moment), but I still think it's a bad idea to shard after shuffling. I'd move it before the shuffle.

scheduler/scheduler.go
```diff
@@ -346,3 +360,23 @@ func maxDepth(tables schema.Tables) uint64 {
...
+	}
+	num := int(shard.num)
+	total := int(shard.total)
+	chunkSize := len(tableClients) / total
+	if chunkSize == 0 {
+		chunkSize = 1
+	}
+	chunks := lo.Chunk(tableClients, chunkSize)
+	if num > len(chunks) {
+		return nil
```
marianogappa, line 376: Shouldn't this be an error?

scheduler/scheduler.go
```diff
@@ -119,6 +126,11 @@ type Scheduler struct {
 	invocationID string
 }
 
+type shard struct {
+	num   int32
```
marianogappa, line 130: Why `int32`?

erezrokah: For context https://github.com/cloudquery/plugin-pb/pull/26#pullrequestreview-2310265498, `int32` comes from protobuf type support (no `int`)

scheduler/scheduler.go
erezrokah: I think an empty shard (e.g. 10 table clients, 100 shards) shouldn't error out, it simply shouldn't do anything

scheduler/scheduler_shuffle.go
erezrokah: OK let me try and switch the order and re-run the tests. We shuffle (this is the default in AWS) to avoid rate limits. Don't think sharding before shuffling will make a difference in that aspect but I'll re-test

erezrokah: > Don't think sharding before shuffling will make a difference in that aspect but I'll re-test

I think it will be fine since we round-robin before we shuffle anyways

erezrokah: Ok did a bit of testing and it looks good so we can shard before shuffle

marianogappa: Is there any case where collecting the tables could be non-deterministic? Normally the tables are hardcoded in a plugin, so it should not be the case. If there is a plugin where the tables are dynamic, and they could change between syncs (e.g. if they are discovered by an API which is non-deterministic), sharding would not work.

In either case, I think the deterministic requirement is worth a one-liner comment.

marianogappa (approved):

scheduler/scheduler_shuffle.go
erezrokah: > Is there any case where collecting the tables could be non-deterministic?

This is a good point, and definitely a limitation of this approach, see below ⬇️ 

- It can happen due to a bug in the plugin https://github.com/cloudquery/cloudquery-private/pull/4299.
- Plugins with dynamic tables don't use the scheduler, they do their own thing so they would need to implement sharding on the plugin's side (if needed. e.g. For Postgres source probably better to use a stronger machine instead of sharding)
- I can think of other cases, e.g. some creating an AWS account after shard 1/2 discovery and before shard 2/2 discovery. If we discover all accounts, that will mess up the sharding. A solution would be to hard code the accounts in the spec to avoid it.

PR #1900, opened 2024-09-19
feat: Add `opts.SchedulerOpts()` helper to convert `plugin.SyncOptions` for scheduler

disq:


plugin/plugin_source.go
```diff
@@ -6,6 +6,7 @@ import (
 
 	"github.com/cloudquery/plugin-sdk/v4/glob"
 	"github.com/cloudquery/plugin-sdk/v4/message"
+	"github.com/cloudquery/plugin-sdk/v4/scheduler"
```
disq, line 9: Unsure if it's worth the extra import. This is the only line in the `plugin` package to refer to `scheduler`.

erezrokah: We can put this here https://github.com/cloudquery/plugin-sdk/tree/main/helpers

disq: `scheduler` already imports `helpers` (DFS uses `helpers.InterfaceSlice()` so it has to be a separate package. `opts` package, `func SchedulerOpts()`?

erezrokah: Now I see 
image

disq: forgot to `git add`, sorry fixed in https://github.com/cloudquery/plugin-sdk/pull/1900/commits/4d699f5bab90c9770e3f1e2766c39fca2b8b62b3

erezrokah (approved):

PR #1905, opened 2024-09-23
feat: Add Time configtype

rhino1998:
#### Summary

This adds a new Time type to be used in plugin specs that can represent either fixed times or relative times. A future PR should extend the duration parsing to handle days, etc.

```yml
kind: source
spec:
  name: "orb"
  registry: "grpc"
  path: "localhost:7777"
  version: "v1.0.0"
  tables:
    ["*"]
  destinations:
    - "sqlite"
  spec:
    api_key: "${CQ_ORB_API_KEY}"
    timeframe_start: 10 days ago
    timeframe_end: "2024-09-25T03:52:40.14157935-04:00"
```

Here `timeframe_start` and `timeframe_end` are both configtype.Time types. Using mixed relative/fixed timestamps for timeframes isn't really useful here as the relative one will be relative to the run-time of the sync but in this example having
```yml
timeframe_start: 10 days ago
timeframe_end: now
```
or even omitting `timeframe_end` entirely is more useful
```yml
timeframe_start: 10 days ago
```

The plugin is free to define default like so
```go
type Spec struct {
	// TimeframeStart is the beginning of the timeframe in which to fetch cost data.
	TimeframeStart configtype.Time `json:"timeframe_start"`

	// TimeframeEnd is the beginning of the timeframe in which to fetch cost data.
	TimeframeEnd configtype.Time `json:"timeframe_end"`

	// ...
}


func (s *Spec) SetDefaults() {
	if s.TimeframeStart.IsZero() {
		s.TimeframeStart = configtype.NewRelativeTime(-1 * 365 * 24 * time.Hour)
	}

	if s.TimeframeEnd.IsZero() {
		s.TimeframeEnd = configtype.NewRelativeTime(0)
	}
}
```


erezrokah (commented):
Hi @rhino1998, nice 🚀 Can you give an example how this will be used in a plugin and in a spec file?

rhino1998:
Added some more context. Should `configtype.Time.Time` take a `time.Time` instead of a `func() time.Time`?

erezrokah (changes requested):
I like the idea of having this kind of type as an SDK helper.

A few comments:
1. If the purpose is to make it easier to configure relative times, I would use https://github.com/tj/go-naturaldate/tree/master instead of a wrapper over duration
2. I would hide the `IsRelative` and `NewRelativeTime` and do `NewTime(t string)` that accepts either an RFC3339 or a natural date string. I don't see a reason consumers need to know the internal representation, also I don't think we even need `typ timeType`, as you can have `NewTime(t string, base time.Time)` and convert the string on creation.

configtype/time.go
```diff
@@ -0,0 +1,141 @@
...
+
+func (t *Time) UnmarshalJSON(b []byte) error {
+	var s string
+	if err := json.Unmarshal(b, &s); err != nil {
+		return err
+	}
+
+	var err error
+	switch {
+	case timeRFC3339Regexp.MatchString(s):
```
erezrokah, line 69: Instead of using regular expressions can't we try to parse with each format, then return an error if all formats failed to parse?

hermanschaaf:
> If the purpose is to make it easier to configure relative times, I would use https://github.com/tj/go-naturaldate/tree/master instead of a wrapper over duration

I would push back on this; the library was last updated 4 years ago, and supports much more than we need to support. I think we would be better off supporting a small, well-defined number of formats.

erezrokah:
> the library was last updated 4 years ago

I agree it hasn't change in a long time, but it doesn't have any dependencies (other than the ones used for testing), and the author is https://github.com/tj (https://x.com/tjholowaychuk), which has a bunch of other popular repositories like https://github.com/tj/commander.js and https://github.com/koajs/koa

> well-defined number of formats.

Agree it does more than we need, but `timeframe_start: -1000h` seems more machine friendly than human friendly to me 

> Also I think the purpose here is not to make it easier, it's to make it possible at all (in Cloud), in a standardized way.

Can you add more context on the Cloud use case? We want the backend to specify a duration via spec generated via backend code in Go? If so is it possible to convert the duration to time before?

hermanschaaf:
> Can you add more context on the Cloud use case? We want the backend to specify a duration via spec generated via backend code in Go? If so is it possible to convert the duration to time before?

We want users to be able to specify a relative time, such as "30 days ago", which will be stored as part of the spec. Every time the sync runs, this should be converted to a timestamp relative to the current time. This way, we don't need to involve the backend to do string interpolation on the spec, and things remain relatively simple.

Right now, what happens in many plugins is they allow only a fixed timestamp to be specified, such as `2024-05-01T00:00:00Z`, or a default value, often relative (30 days ago or now). This doesn't lend us the flexibility we desire for scheduled syncs in Cloud

hermanschaaf:
> Agree it does more than we need, but timeframe_start: -1000h seems more machine friendly than human friendly to me

I agree with this.

I'm also okay with using that library, but if we do, let's not support everything that the library supports, but instead limit it to a defined list of supported formats that we can control. 

I would maybe say something like:
 - `now`
 - `x seconds [ago|from now]`
 - `x minutes [ago|from now]`
 - `x hours [ago|from now]`
 - `x days [ago|from now]`

would be a good enough start (we can always add more if requested)

erezrokah:
> I would maybe say something like:
> 
> * `now`
> * `x seconds [ago|from now]`
> * `x minutes [ago|from now]`
> * `x hours [ago|from now]`
> * `x days [ago|from now]`
> 
> would be a good enough start (we can always add more if requested)

Sounds like a plan to me

hermanschaaf:
I forgot that in addition to relative times, we'd also want to support fixed timestamps of course, probably using RFC 3339. And also some support for plain dates, e.g. `2024-05-10`, for plugins that need it.

configtype/time.go
rhino1998: Yeah, although we still need regexes for the jsonschema definitions

configtype/time_test.go
```diff
@@ -0,0 +1,180 @@
...
+		},
+		{
+			Name: "relative duration",
+			Err:  false,
+			Spec: `"10 months from now"`,
+		},
+		{
+			Name: "complex relative duration",
+			Err:  false,
+			Spec: `"10 months 3 days 4h20m from now"`,
```
hermanschaaf, line 133: I think if we're going to expose `MarshalJSON` then we should probably make this format `10 months 3 days 4 hours 20 minutes from now` for consistency (and so we don't need to make a breaking change later to fix it)

hermanschaaf: Alternatively we could remove the ability to specify an arbitrary duration in the constructor (just remove the constructor) and only allow this to be constructed using a simple string like `10 days ago`

configtype/duration.go
```diff
@@ -2,29 +2,161 @@ package configtype
...
+	"fmt"
+	"regexp"
+	"strconv"
+	"strings"
 	"time"
 
 	"github.com/invopop/jsonschema"
 )
 
+var (
```
erezrokah, line 14: I prefer if we don't change the semantics of the existing duration type, as that could breaking existing behavior for consumers of this type

configtype/time.go
```diff
@@ -0,0 +1,164 @@
...
+// Time is a wrapper around time.Time that should be used in config
+// when a time type is required. We wrap the time.Time type so that
+// the spec can be extended in the future to support other types of times
+type Time struct {
+	typ      timeType
+	time     time.Time
+	duration Duration
+}
+
+func NewTime(t time.Time) Time {
```
erezrokah, line 29: I would remove this in favor of `ParseTime` so there's only a single way to create this type

configtype/time.go
```diff
@@ -0,0 +1,164 @@
...
+	case timeTypeFixed:
+		return json.Marshal(t.time)
+	case timeTypeRelative:
+		return json.Marshal(t.duration)
+	default:
+		return json.Marshal(time.Time{})
+	}
+}
+
+func (t Time) Time(now time.Time) time.Time {
```
erezrokah, line 116:
```suggestion
func (t Time) AsTime(now time.Time) time.Time {
```

I think that's more common

configtype/time.go
```diff
@@ -0,0 +1,164 @@
...
+	case timeTypeRelative:
+		if t.duration.Duration() == 0 {
+			return "now"
+		}
+
+		return t.duration.String()
+	default:
+		return time.Time{}.String()
+	}
+}
```
erezrokah, line 164: Do we need to implement these? Where are they used?

configtype/time.go
```diff
@@ -0,0 +1,164 @@
...
+	if err != nil {
+		return err
+	}
+
+	*t = tim
+	return nil
+}
+
+func (t Time) MarshalJSON() ([]byte, error) {
+	switch t.typ {
```
erezrokah, line 106: I think we can keep the original string that we got in `ParseTime` and return that instead, WDYT?

configtype/time.go
```diff
@@ -0,0 +1,164 @@
...
+
+	"github.com/invopop/jsonschema"
+)
+
+type timeType int
+
+const (
+	timeTypeZero timeType = iota
+	timeTypeFixed
+	timeTypeRelative
```
erezrokah, line 17: Do wen need this enum? Can we check if `t.duration != nil` instead to know if it's a relative time?

erezrokah (changes requested):
Thanks for the following @rhino1998, added a few more comments

configtype/time_test.go
rhino1998: The reason for the weird hybrid format is for backwards compat for the Duration type. It ensures that anything constructed from a plain duration will emit the same json as before this PR. It may be reasonable to separate the json serialization format and the string parsed format and make the serialization format (but still unmarshalling the nicer human format) just `[-+](\d+[a-zA-Z])+` (basically time.ParseDuration format but with additional units). That frees us to have the string format be nicer and human readable

configtype/duration.go
rhino1998: All previously value values are still value for the duration type. This just adds more possible valid values. Additionally all previously valid values will serialize to the same string after this PR in its current state. (unless I missed something)

configtype/time.go
rhino1998: IsZero is used for optional spec fields. (although I guess a ptr to the type could work as well)
String is nice to have for debugging

I guess the rest could be removed

configtype/duration.go
erezrokah: > This just adds more possible valid values.

Exactly, so now the JSON validation is more relaxed, so something that would have failed before can pass now.
While this could be seen as a feature, it's better if plugins opt-in to it, then document the new behavior

configtype/time.go
erezrokah: > IsZero is used for optional spec fields

How is it used? Can we make these fields pointers instead and use nullability for optional fields?

erezrokah: > (although I guess a ptr to the type could work as well)

Ah yes, let's do that

configtype/time_test.go
erezrokah: Would https://github.com/cloudquery/plugin-sdk/pull/1905/files#r1775175473 solve this?

configtype/time.go
rhino1998: [263750b](https://github.com/cloudquery/plugin-sdk/pull/1905/commits/263750b142816f645bb9b548586602394e87cbc2), but unclear what the handling for an uninitialized Time or Duration should be. Currently it just marshals to `""`

rhino1998: Also, as a side-effect Time is no longer == comparable in a useful way

erezrokah: > [263750b](https://github.com/cloudquery/plugin-sdk/pull/1905/commits/263750b142816f645bb9b548586602394e87cbc2), but unclear what the handling for an uninitialized Time or Duration should be. Currently it just marshals to `""`

I think we need to decide that and pick whatever we want to happen, we can

1. Error out on uninitialized
2. Return `MarshalJSON` of the zero value of `time.Time`
3. Return `MarshalJSON` of the zero value of `time.Duration`
4. Return `null`
5. Maybe more options

Not sure what's the best option but it needs to be consistent with `UnmarshalJSON`.
We can see what happens for `time.Time` and `time.Duration` zero values

erezrokah: > Also, as a side-effect Time and Duration are no longer == comparable in a useful way

How could you compare Time and Duration before? Where do you need this comparison and how will it be used?

rhino1998: Duration was comparable with == (as it was just a struct) containing a time.Duration. I'm unsure if that was used anywhere but removing that is a breaking change. Adding days,months,etc to the struct maintained comparability for equal values but adding the `input` string field breaks that for anything besides values that had the same input string even if they're otherwise equivalent (including whitespace differences).

I guess Time wasn't usefully comparable with `==` as while time.Time is `==` comparable doing doesn't handle timezones and so is not recommended.

As for where it is needed, it isn't exactly for time, but the duration change is breaking change in the API (unclear if anything actually uses it though)

erezrokah: Ok so:
1. I don't think we should modify the existing duration config type. Relaxing the parsing is also a breaking change
2. Even if we did modify the existing type, we can used the parsed value of the duration for `Equals`

rhino1998: Regarding 2., that still will have broken `==` comparison, although again unclear if that is used externally or if that even counts as an explicit part of the API to be broken

rhino1998: Regarding default behavior, returning the zero value of time seems sane as it'll allow the current IsZero check to work. Although that is `"0001-01-01T00:00:00Z"`

https://go.dev/play/p/qI6SprLcw5p

erezrokah:
> @erezrokah @rhino1998 I lost track a bit, is there still anything unresolved on this PR?

Not sure https://github.com/cloudquery/plugin-sdk/pull/1905/files#r1774764467 is resolved, also https://github.com/cloudquery/plugin-sdk/pull/1905#discussion_r1775148934

I'd rather if we don't change the existing type behavior as that implicitly change spec parsing behavior for plugins. If plugins want to support the new human friend format mixed duration/time format they should do it explicitly I think

rhino1998:
> > @erezrokah @rhino1998 I lost track a bit, is there still anything unresolved on this PR?
> 
> Not sure https://github.com/cloudquery/plugin-sdk/pull/1905/files#r1774764467 is resolved, also [#1905 (comment)](https://github.com/cloudquery/plugin-sdk/pull/1905#discussion_r1775148934)
> 
> I'd rather if we don't change the existing type behavior as that implicitly change spec parsing behavior for plugins. If plugins want to support the new human friend format mixed duration/time format they should do it explicitly I think

The new mixed duration/time format is opt-in, but the new duration format does extend the old plain time.ParseDuration-based format.

How would you prefer the opt-in be done? A DurationV2 type? A separate package?

erezrokah:
> How would you prefer the opt-in be done? A DurationV2 type? A separate package?

I don't mind I mostly don't want to change the existing behavior of the current type. You can also put everything under the new `Time` type without another type

configtype/time_test.go
```diff
@@ -0,0 +1,146 @@
...
```
erezrokah, line 99: Should we change this to @hermanschaaf's suggestion of `10 months 3 days 4 hours 20 minutes from now`?

rhino1998: In its current state, both are handled and whatever was input will be emitted. I added some test cases to make this clearer. https://github.com/cloudquery/plugin-sdk/pull/1905/commits/1a5ae0a5ae982e11168e7ad49dcb613ca3ab105c

erezrokah (approved):

erezrokah:
Nothing blocking on my end, so you can merge at will. We might want to wait with releasing the new SDK until after today's plugins release cycle to avoid creating noise for the plugins teams

disq:
I think (sorry for the late comment) the `from now` syntax can be confusing, as people could expect `2 days from <some date here>` or worse, `2 days from new year's eve` to work.

rhino1998:
> I think (sorry for the late comment) the `from now` syntax can be confusing, as people could expect `2 days from <some date here>` or worse, `2 days from new year's eve` to work.

We could pretty easily add support for stuff like `2 days from <timestamp>`. That actually generalizes the input pretty nicely and will simplify/merge validation cases

rhino1998:
@disq  Do you consider the above blocking?

disq:
> @disq Do you consider the above blocking?

@rhino1998 No not at all.

erezrokah:
Not sure what's the status here, should we merge this or close it as stale?

PR #1913, opened 2024-10-02
fix: Error handling in StreamingBatchWriter

disq:
Actual clean up by @murarustefaan but I've done some minor updates to it and updated tests.

~Still need to test it E2E.~ Seems to work, handled the `panic: arrow/array: number of columns/fields mismatch [recovered]` error immediately (I tested with an old version of the S3 plugin, had to manually bump Arrow to v17)


writers/streamingbatchwriter/streamingbatchwriter_test.go
```diff
@@ -225,7 +235,7 @@ func TestStreamingBatchSizeRows(t *testing.T) {
 		t.Fatalf("expected 0 open tables, got %d", l)
 	}
 
-	if l := testClient.MessageLen(messageTypeInsert); l != 3 {
```
disq, line 228: Added two inserts above, but increased this from 3 to 4... This was incorrect to expect 3 because the flushing flow has been different for some time, and you don't need the third message to make it flush.

writers/streamingbatchwriter/streamingbatchwriter.go
```diff
@@ -441,6 +456,11 @@ func (s *streamingWorkerManager[T]) run(ctx context.Context, wg *sync.WaitGroup,
 				ticker.Reset(s.batchTimeout)
 			}
 			done <- true
+		case err := <-outputCh:
+			if err != nil {
+				s.errCh <- fmt.Errorf("handler failed on %s: %w", tableName, err)
+				return
```
disq, line 462: This `return` statement is my actual contribution apart from renaming variables and updating tests, oh and better handling of error type panics.

writers/streamingbatchwriter/streamingbatchwriter.go
```diff
@@ -182,26 +182,28 @@ func (w *StreamingBatchWriter) Write(ctx context.Context, msgs <-chan message.Wr
...
 
-	go func() {
-		for err := range errCh {
-			w.logger.Err(err).Msg("error from StreamingBatchWriter")
-		}
-	}()
+	for {
+		select {
+		case msg := <-msgs:
+			if msg == nil {
```
marianogappa, line 188: Wouldn't you want to `msg, ok :=` instead here?

writers/streamingbatchwriter/streamingbatchwriter.go
```diff
@@ -182,26 +182,28 @@ func (w *StreamingBatchWriter) Write(ctx context.Context, msgs <-chan message.Wr
...
+			}
+			w.lastMsgType = msgType
+			if err := w.startWorker(ctx, errCh, msg); err != nil {
 				return err
 			}
-		}
-		w.lastMsgType = msgType
-		if err := w.startWorker(ctx, errCh, msg); err != nil {
+
+		case err := <-errCh:
```
marianogappa, line 203: The previous logic wasn't shutting down on error; is this correct?

marianogappa:
@erezrokah adding you as well; this is a little tricky with not that much experience with the writer

writers/streamingbatchwriter/streamingbatchwriter.go
disq: Yeah that would be clearer, fixed in https://github.com/cloudquery/plugin-sdk/pull/1913/commits/512d6c0af379dd8b31b50a47c2071933f6946731

writers/streamingbatchwriter/streamingbatchwriter.go
disq: Previous logic would do a `w.logger.Err(err).Msg("error from StreamingBatchWriter")` and continue working.

erezrokah (approved):

disq:
Merge at will, I'm not merging right now to prevent any potential noise. (and the unreleased SDK already has some interesting changes)

erezrokah:
Looks like this broke something https://github.com/cloudquery/cloudquery/pull/19312

PR #1914, opened 2024-10-02
feat: Implement RandomQueue scheduler strategy

marianogappa:
This PR implements a new Scheduler Strategy based on a _Concurrent Random Queue_. It is based on @erezrokah 's Priority Queue Scheduler Strategy.

## How does it work

This is hopefully a much simpler scheduling strategy. It doesn't have any semaphores; it just uses the existing concurrency setting.

Table resolvers (and their relations) get `Push`ed into a work queue, and `concurrency` workers `Pull` from this queue, but they pull a random element from it.

## Why it should work better

**The key benefit of this strategy is this:**
- Assumption 1: most slow syncs are actually slow because of rate limits, not because of I/O limits or too much data.
- Assumption 2: the meaty part of the sync is syncing relations, because each child table has a resolver per parent.
- Benefit: because the likelihood of picking up a child resolver of a given table is uniformly distributed across the `int32` range, all relation API calls are evenly spread throughout the sync, thus optimally minimising rate limits!

## Does it work better?

Still working on results. Notably AWS & Azure yield mixed results; still have to look into why.

### GCP

**Before**

```
$ cli sync .
Loading spec(s) from .
Starting sync for: gcp (grpc@localhost:7777) -> [postgresql (cloudquery/postgresql@v8.0.7)]
Sync completed successfully. Resources: 25799, Errors: 0, Warnings: 0, Time: 2m23s
```

UPDATE: GCP is moving to Round Robin strategy, and it's much faster with this strategy:

```
$ cli sync .
Loading spec(s) from .
Starting sync for: gcp (grpc@localhost:7777) -> [postgresql (cloudquery/postgresql@v8.0.7)]
Sync completed successfully. Resources: 26355, Errors: 0, Warnings: 0, Time: 40s
```

**After**

```
$ cli sync .
Loading spec(s) from .
Starting sync for: gcp (grpc@localhost:7777) -> [postgresql (cloudquery/postgresql@v8.0.7)]
Sync completed successfully. Resources: 26186, Errors: 0, Warnings: 0, Time: 34s
```

**Result:  76.22% reduction in time, or 3.21 times faster.**
**Result against Round Robin: 15% reduction in time, or 0.18 times faster (probably within margin of error)**

### BigQuery

**Before**

```
$ cli sync bigquery_to_postgresql.yaml
Loading spec(s) from bigquery_to_postgresql.yaml
Starting sync for: bigquery (cloudquery/bigquery@v1.7.0) -> [postgresql (cloudquery/postgresql@v8.6.0)]
Sync completed successfully. Resources: 26139, Errors: 0, Warnings: 0, Time: 2m7s
```

**After**

```
$ cli sync bigquery_to_postgresql.yaml
Loading spec(s) from bigquery_to_postgresql.yaml
Starting sync for: bigquery (cloudquery/bigquery@v1.7.0) -> [postgresql (cloudquery/postgresql@v8.6.0)]
Sync completed successfully. Resources: 26139, Errors: 0, Warnings: 0, Time: 1m26s
```

**Result: 32.28% reduction in time, or 0.48 times faster**

### SentinelOne

**Before** (it was already quite fast due to previous merged improvement)

```
$ cli sync .
Loading spec(s) from .
Starting sync for: sentinelone (grpc@localhost:7777) -> [postgresql (cloudquery/postgresql@v8.5.5)]
Sync completed successfully. Resources: 1295, Errors: 0, Warnings: 0, Time: 15s
```

**After**

```
$ cli sync .
Loading spec(s) from .
Starting sync for: sentinelone (grpc@localhost:7777) -> [postgresql (cloudquery/postgresql@v8.5.5)]
Sync completed successfully. Resources: 1295, Errors: 0, Warnings: 0, Time: 8s
```

**Result: 46.67% reduction in time, or 0.875 times faster**

## How to test

- Add a `go.mod` replace for sdk: `replace github.com/cloudquery/plugin-sdk/v4 => github.com/cloudquery/plugin-sdk/v4 v4.63.1-0.20241002131015-243705c940c6` (check last commit on this PR)
- Run source plugin via grpc locally; make sure to configure the scheduler strategy to `scheduler.StrategyRandomQueue`.

## How scary is it to merge

- This scheduler strategy is not used by any plugins by default, so in principle this should be safe to merge.


scheduler/queue/concurrent_random_queue.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+func (q *ConcurrentRandomQueue[T]) Pop() *T {
+	q.mu.Lock()
+	defer q.mu.Unlock()
+
+	if len(q.queue) == 0 {
+		return nil
+	}
+	idx := rand.Intn(len(q.queue))
```
erezrokah, line 33: I like the randomness idea 💯 

Maybe we can use a seed like we do here https://github.com/cloudquery/plugin-sdk/blob/00b9d9a6c93bad24b0ea7d489ef13d65cead2023/scheduler/scheduler_shuffle.go#L46

so the order is somewhat the same if your table list doesn't change (e.g. you re-run the same sync)

scheduler/scheduler_random_queue.go
```diff
@@ -0,0 +1,47 @@
...
+		}
+		preInitialisedClients[i] = clients
+		// we do this here to avoid locks so we initial the metrics structure once in the main goroutines
+		// and then we can just read from it in the other goroutines concurrently given we are not writing to it.
+		s.metrics.InitWithClients(table, clients)
+	}
+
+	tableClients := roundRobinInterleave(s.tables, preInitialisedClients)
+	seed := hashTableNames(tableNames)
+	shuffle(tableClients, seed)
```
erezrokah, line 29: We should use `shardTableClients` before passing the `tabeClients` to the scheduler

marianogappa: Added the sharding.

scheduler/queue/concurrent_random_queue.go
marianogappa: I passed the seed to the Scheduler's constructor, and from it to the Random Queue.

scheduler/queue/active_work_signal.go
```diff
@@ -0,0 +1,67 @@
...
+// if it's empty after that worker finishes. That's why we need this.
+//
+// Use it like this:
+//
+// - When a worker picks up a task, call `Add()` (like a WaitGroup)
+// - When a worker finishes a task, call `Done()` (like a WaitGroup)
+//
+// - If the queue is empty, check `IsIdle()` to check if no workers are active.
+// - If workers are still active, call `Wait()` to block until state changes.
+type activeWorkSignal struct {
```
erezrokah, line 23: Nice 🚀 Much better than what I did before 🙈

erezrokah (approved):
Code looks great, I'm going to test it a bit with a few plugins, also with OTEL enabled

scheduler/strategy.go
```diff
@@ -70,11 +70,12 @@ func (Strategy) JSONSchema() *jsonschema.Schema {
...
-var AllStrategies = Strategies{StrategyDFS, StrategyRoundRobin, StrategyShuffle}
+var AllStrategies = Strategies{StrategyDFS, StrategyRoundRobin, StrategyShuffle, StrategyRandomQueue}
 var AllStrategyNames = [...]string{
-	StrategyDFS:        "dfs",
-	StrategyRoundRobin: "round-robin",
-	StrategyShuffle:    "shuffle",
+	StrategyDFS:         "dfs",
+	StrategyRoundRobin:  "round-robin",
+	StrategyShuffle:     "shuffle",
+	StrategyRandomQueue: "random-queue",
```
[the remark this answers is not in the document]
marianogappa: Renamed to `shuffle-queue`.

scheduler/queue/concurrent_random_queue.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+func (q *ConcurrentRandomQueue[T]) Pop() *T {
+	q.mu.Lock()
+	defer q.mu.Unlock()
+
+	if len(q.queue) == 0 {
+		return nil
+	}
+	idx := q.random.Intn(len(q.queue))
```
erezrokah, line 34: OK so one thing with this. It also randomizes the top level tables.

Maybe we should consider giving priority to those. So first we at least start all the top level tables, and only then start doing relations.

erezrokah: Resolving this as non blocking. We can test my suggestion later

scheduler/queue/worker.go
```diff
@@ -0,0 +1,170 @@
...
+	for r := range res {
+		w.resolveResource(ctx, table, client, parent, r)
+	}
+
+	endTime := time.Now()
+	duration := endTime.Sub(startTime)
+	tableMetrics.Duration.Store(&duration)
+	tableMetrics.OtelEndTime(ctx, endTime)
+	if parent == nil {
+		logger.Info().Uint64("resources", tableMetrics.Resources).Uint64("errors", tableMetrics.Errors).Dur("duration_ms", duration).Msg("table sync finished")
```
erezrokah, line 113: OK now I remember what I removed here. We used to log a message here with the metrics of all relations after a parent finished.

Not we don't do that since reaching this code doesn't mean all relations finished syncing, it only means the parent finished pushing all the new work units into the queue.

See https://github.com/cloudquery/plugin-sdk/blob/2977ae3d373042219e514f744d59cf07af5faea3/scheduler/scheduler_dfs.go#L125

I don't think it's a blocker and we can add logs alter on. But still something to consider

PR #1921, opened 2024-10-03
fix: Error handling in StreamingBatchWriter

disq:
original PR: https://github.com/cloudquery/plugin-sdk/pull/1913

Reverts the revertion, plus more improvements:
- Client's handlers are now allowed to return an immediate error without draining the channel. The channel will be automatically drained for the error condition, and subsequent writes for that table won't get sent to the client.
- Some potential race conditions are fixed (`ensureOpened()` inline-func now got refactored into `s.send()` and handles the sending as well. The spawned goroutine doesn't refer to `inputCh` directly so that it can be replaced in `closeFlush()`.
- Shutdown logic is handled better: even if `client.Write()` returns an error after `msgs` is closed, it's still logged and returned.


writers/streamingbatchwriter/streamingbatchwriter.go
```diff
@@ -178,30 +179,36 @@ func (w *StreamingBatchWriter) Close(context.Context) error {
...
 	errCh := make(chan error)
 	defer close(errCh)
-
-	go func() {
-		for err := range errCh {
-			w.logger.Err(err).Msg("error from StreamingBatchWriter")
-		}
+	defer func() {
+		err := w.Close(ctx)
+		retErr = errors.Join(retErr, err)
```
disq, line 187: New improvement: always do w.Close even we returned an error, and join errors.

erezrokah (approved):

marianogappa (approved):

disq:
Should be merged after merging:
- https://github.com/cloudquery/cloudquery/pull/19317
- https://github.com/cloudquery/cloudquery/pull/19318
- https://github.com/cloudquery/cloudquery/pull/19319
(S3 PR is already merged)

murarustefaan (approved):

PR #1941, opened 2024-10-22
fix: Ensure module field exists in all log messages

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery-issues/issues/2627

The wrong value in the `module` field is because of the bug fixed in https://github.com/cloudquery/plugin-pb-go/pull/421

---

- [ ] Update or add tests 🧪

### DIFF
--- a/serve/plugin.go
+++ b/serve/plugin.go
@@ -134,6 +134,8 @@ func (s *PluginServe) newCmdPluginServe() *cobra.Command {
 				logger = log.Output(zerolog.ConsoleWriter{Out: os.Stdout}).Level(zerologLevel)
 			}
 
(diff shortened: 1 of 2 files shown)


serve/plugin.go
```diff
@@ -134,6 +134,8 @@ func (s *PluginServe) newCmdPluginServe() *cobra.Command {
 				logger = log.Output(zerolog.ConsoleWriter{Out: os.Stdout}).Level(zerologLevel)
 			}
 
+			logger = logger.With().Str("module", s.plugin.Name()+"-"+string(s.plugin.Kind())).Logger()
```
erezrokah, line 137: Not show in the diff but we do `s.plugin.SetLogger(logger)` a bit below

marianogappa (approved):

PR #1947, opened 2024-10-28
fix: Flaky usage tests

dcelasun:
The problem is that there is no way to know if all rows have already been processed by the time `*BatchUpdater.Close()` is called. If they have been, then `numberOfUpdates()` will return 1 less than expected, causing the flakiness.

There are two ways to fix this:
- Make sure enough time passes between calls to `Increase()` and `Close()`. A call to `runtime.Gosched()` in the test before calling close could also help, but there are no guarantees. Timing-based tests are also generally icky.
- Keep track of whether `Close()` actually resulted in any processed rows, and adjust the expected number of updates if so. This has to advantage of working 100% reliably, eliminating the flakiness, at the cost of introducing a field only used for tests.

This commit implements the second approach.

This fixes https://github.com/cloudquery/cloudquery-issues/issues/2367

### DIFF
--- a/premium/usage_test.go
+++ b/premium/usage_test.go
@@ -287,15 +287,20 @@ func TestUsageService_Increase_WithMinimumUpdateDuration(t *testing.T) {
 
 	usageClient := newClient(t, apiClient, WithBatchLimit(0), WithMinTimeBetweenFlushes(30*time.Second))
 
-	for i := 0; i < 10000; i++ {
+	rows := 10000
+	for i := 0; i < rows; i++ {
 		err = usageClient.Increase(1)
 		require.NoError(t, err)
 	}
 	err = usageClient.Close()
 	require.NoError(t, err)
 
-	assert.Equal(t, 10000, s.sumOfUpdates(), "total should equal number of updated rows")
-	assert.Equal(t, 2, s.numberOfUpdates(), "should only update first time and on close if minimum update duration is set")
+	assert.Equal(t, rows, s.sumOfUpdates(), "total should equal number of updated rows")
+	if usageClient.dataOnClose {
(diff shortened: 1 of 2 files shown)


premium/usage_test.go
```diff
@@ -287,15 +287,20 @@ func TestUsageService_Increase_WithMinimumUpdateDuration(t *testing.T) {
...
+	for i := 0; i < rows; i++ {
 		err = usageClient.Increase(1)
 		require.NoError(t, err)
 	}
 	err = usageClient.Close()
 	require.NoError(t, err)
 
-	assert.Equal(t, 10000, s.sumOfUpdates(), "total should equal number of updated rows")
-	assert.Equal(t, 2, s.numberOfUpdates(), "should only update first time and on close if minimum update duration is set")
+	assert.Equal(t, rows, s.sumOfUpdates(), "total should equal number of updated rows")
```
erezrokah, line 298: I was thinking we can change the test to something like this:

```go
// Number of updates should be between 1 and 2 depending if when close is called there are remaining rows to update
assert.GreaterOrEqual(t, 1, s.numberOfUpdates())
assert.LessOrEqual(t, 2, s.numberOfUpdates())
```

I think it's ok to test the side effect of the code, doesn't really matter why it's 1 or 2. Now the test logic depends on the correctness of `dataOnClose`, not a big issue though. So you can take it or leave it

erezrokah (approved):

erezrokah:
Woops didn't notice the automerge. Anyways I changed the PR title so this change doesn't show up in the change log as it's a non user visible change

premium/usage_test.go
dcelasun: Will take a note for next time :)

PR #1950, opened 2024-10-29
fix: Clean up usage retry logic

disq:
I may have used the term "clean up" loosely.

### DIFF
--- a/premium/usage.go
+++ b/premium/usage.go
@@ -581,6 +581,7 @@ func (u *BatchUpdater) updateMarketplaceUsage(ctx context.Context, rows uint32) 
 		if !errors.As(lastErr, &de) {
 			return fmt.Errorf("failed to update usage: %w", lastErr)
 		}
+		u.logger.Debug().Err(lastErr).Int("try", retry).Uint32("rows", rows).Msg("usage update failed due to duplicate request")
 
 		jitter := time.Duration(rand.Intn(1000)) * time.Millisecond
 		time.Sleep(marketplaceDuplicateWaitTime + jitter)
--- a/premium/usage.go
+++ b/premium/usage.go
@@ -12,6 +12,7 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
+	awshttp "github.com/aws/aws-sdk-go-v2/aws/transport/http"
 	awsConfig "github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/service/marketplacemetering"
 	"github.com/aws/aws-sdk-go-v2/service/marketplacemetering/types"
@@ -31,6 +32,9 @@ const (
 	defaultMaxWaitTime           = 60 * time.Second
 	defaultMinTimeBetweenFlushes = 10 * time.Second
 	defaultMaxTimeBetweenFlushes = 30 * time.Second
+
+	marketplaceDuplicateWaitTime = 1 * time.Second
+	marketplaceMinRetries        = 20
 )
 
 const (
@@ -99,7 +103,9 @@ func WithMinTimeBetweenFlushes(minTimeBetweenFlushes time.Duration) UsageClientO
 // WithMaxRetries sets the maximum number of retries to update the usage in case of an API error
 func WithMaxRetries(maxRetries int) UsageClientOptions {
 	return func(updater *BatchUpdater) {
-		updater.maxRetries = maxRetries
+		if maxRetries > 0 {
+			updater.maxRetries = maxRetries
+		}
 	}
 }
 
@@ -278,6 +284,9 @@ func (u *BatchUpdater) setupAWSMarketplace() error {
 	u.batchLimit = 1000000000
 
 	u.minTimeBetweenFlushes = 1 * time.Minute
+	if u.maxRetries < marketplaceMinRetries {
+		u.maxRetries = marketplaceMinRetries
+	}
 	u.backgroundUpdater()
 	return nil
 }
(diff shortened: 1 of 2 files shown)


premium/usage.go
```diff
@@ -242,18 +242,26 @@ func NewUsageClient(meta plugin.Meta, ops ...UsageClientOptions) (UsageClient, e
...
+		u.apiClient, err = cqapi.NewClientWithResponses(u.url,
+			cqapi.WithRequestEditorFn(func(_ context.Context, req *http.Request) error {
+				token, err := u.tokenClient.GetToken()
+				if err != nil {
+					return fmt.Errorf("failed to get token: %w", err)
+				}
+				req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", token))
+				return nil
+			}),
+			cqapi.WithHTTPClient(retryClient.StandardClient()),
```
disq, line 260: This is what the default `cqapi.NewClientWithResponses` does anyway, we just set the retry params beforehand.

premium/usage.go
```diff
@@ -465,7 +473,7 @@ func (u *BatchUpdater) backgroundUpdater() {
 				}
 
 				if err := u.updateUsageWithRetryAndBackoff(ctx, totals, tables); err != nil {
-					log.Warn().Err(err).Msg("failed to update usage")
+					u.logger.Warn().Err(err).Msg("failed to update usage")
```
disq, line 476: Cleaning up some "untethered" logger usage.

premium/usage.go
```diff
@@ -543,54 +560,79 @@ func (u *BatchUpdater) reportUsageToAWSMarketplace(ctx context.Context, rows uin
...
+		var (
+			statusCode = -1
+			rerr       *awshttp.ResponseError
+		)
+		if errors.As(lastErr, &rerr) {
+			statusCode = rerr.HTTPStatusCode()
 		}
 
-		retryDuration, err := u.calculateRetryDuration(resp.StatusCode(), resp.HTTPResponse.Header, queryStartTime, retry)
+		retryDuration, err := u.calculateMarketplaceRetryDuration(statusCode, queryStartTime, retry)
```
disq, line 595: AWS client should have its own retry mechanism inside anyway - which won't help us as we want the usage timestamp to point to the next second and not stay in the past.

premium/usage.go
```diff
@@ -278,6 +292,9 @@ func (u *BatchUpdater) setupAWSMarketplace() error {
 	u.batchLimit = 1000000000
 
 	u.minTimeBetweenFlushes = 1 * time.Minute
+	if u.maxRetries < marketplaceMinRetries {
```
marianogappa, line 295: This is:

```go
u.maxRetries = max(u.maxRetries, marketplaceMinRetries)
```

premium/usage.go
```diff
@@ -543,54 +560,79 @@ func (u *BatchUpdater) reportUsageToAWSMarketplace(ctx context.Context, rows uin
...
+		if lastErr == nil {
+			u.logger.Debug().Int("try", retry).Uint32("rows", rows).Msg("usage updated")
+			return nil
 		}
 
-		if len(tables) > 0 {
-			payload.Tables = &tables
+		var de *types.DuplicateRequestException
+		if errors.As(lastErr, &de) {
+			jitter := time.Duration(rand.Intn(1000)) * time.Millisecond
```
marianogappa, line 582: I've seen this already used in this code. Wouldn't it make more sense for the jitter max to be proportional to the backoff, say 5% of it, instead of hardcoded to 1 second?

marianogappa (commented):
Can't find anything wrong with the PR so I put some boring nits there. I think it boils down to testing it in practice and rubber stamping it with an approval if it works. Would you mind putting some evidence syncs in the description?

mnorbury (commented):
Still looking good 🚀

premium/usage.go
disq: Here the backoff is constant at 1 sec, due to how we want the data to be registered to the next second (to prevent `DuplicateRequestException`, which happens because presumably we're running multiple instances and everybody's reporting their usage at the same second)

Because of this, the jitter max value can be increased to 2sec or more (as we only care about 1-second blocks...), but having it lower also makes it sense that it would finish quicker.

disq: The other part where this code lives is for the "regular" retries (not DuplicateRequestException) which is implemented on top of the (presumed already existing) AWS SDK backoff mechanisms. We could remove that cases and set up the AWS SDK backoff params instead, but I'm not an expert on that. This also protects against similar errors (if any that we've potentially missed, or is undocumented) that is not covered by the `errors.As`.

premium/usage.go
```diff
@@ -99,7 +102,9 @@ func WithMinTimeBetweenFlushes(minTimeBetweenFlushes time.Duration) UsageClientO
 // WithMaxRetries sets the maximum number of retries to update the usage in case of an API error
 func WithMaxRetries(maxRetries int) UsageClientOptions {
 	return func(updater *BatchUpdater) {
-		updater.maxRetries = maxRetries
+		if maxRetries > 0 {
```
bbernays, line 105: This means that retries can never be turned off?

premium/usage.go
```diff
@@ -537,60 +555,69 @@ func (u *BatchUpdater) reportUsageToAWSMarketplace(ctx context.Context, rows uin
...
 	return nil
 }
 
-func (u *BatchUpdater) updateUsageWithRetryAndBackoff(ctx context.Context, rows uint32, tables []cqapi.UsageIncreaseTablesInner) error {
+func (u *BatchUpdater) updateMarketplaceUsage(ctx context.Context, rows uint32) error {
+	var lastErr error
 	for retry := 0; retry < u.maxRetries; retry++ {
-		u.logger.Debug().Str("url", u.url).Int("try", retry).Int("max_retries", u.maxRetries).Uint32("rows", rows).Msg("updating usage")
-		queryStartTime := time.Now()
+		u.logger.Debug().Int("try", retry).Int("max_retries", u.maxRetries).Uint32("rows", rows).Msg("updating usage")
```
bbernays, line 572: Since we know we have an issue with duplicate values, can we also include the `timestamp` value in this debug line?

premium/usage.go
```diff
@@ -537,60 +555,69 @@ func (u *BatchUpdater) reportUsageToAWSMarketplace(ctx context.Context, rows uin
...
-		}
-		if resp.StatusCode() >= 200 && resp.StatusCode() < 300 {
-			u.logger.Debug().Str("url", u.url).Int("try", retry).Int("status_code", resp.StatusCode()).Uint32("rows", rows).Msg("usage updated")
-			u.lastUpdateTime = time.Now().UTC()
-			if resp.HTTPResponse != nil {
-				u.updateConfigurationFromHeaders(resp.HTTPResponse.Header)
-			}
-			return nil
-		}
+		jitter := time.Duration(rand.Intn(1000)) * time.Millisecond
```
bbernays, line 585: Do we want to log the error (maybe as a `debug`)?

premium/usage.go
disq: maxRetries = 1 would be "single request, no retries" but I didn't want to be pedantic about the loop and the descriptions.

premium/usage.go
disq: It's not easy to do the way we've set this up, and it's the timestamp from the log line anyway.

premium/usage.go
disq: Added in https://github.com/cloudquery/plugin-sdk/pull/1950/commits/25c1593f5346a35813a750bfcadc2e508021088b

marianogappa (approved):

PR #1972, opened 2024-11-19
feat: Expose invocation ID in OpenTelemetry metrics

disq:
I'm not sure if invocation ID was left out on purpose, as this would increase cardinality of metrics over time? Feel free to close without merging.


disq:
The reason being: metrics are more lightweight/straightforward to parse and make sense of than spans. I will change my approach to use traces.

disq:
Maybe we can enable this as an option, otherwise there isn't a good way to track in-progress tables and whether they're completed or not. I'm currently investigating if this is feasible before diving into how to make it an option.

disq:
This still seems unnecessary as each Metric also contains at least one Exemplar entry for a trace and a span, so it's possible to correlate with some minor work.

PR #1973, opened 2024-11-20
feat: Expose resource stats in traces as separate event

disq:
Still to be tested


erezrokah (approved):

scheduler/queue/worker.go
```diff
@@ -80,6 +80,13 @@ func (w *worker) resolveTable(ctx context.Context, table *schema.Table, client s
 		logger.Info().Msg("top level table resolver started")
 	}
 	tableMetrics := w.metrics.TableClient[table.Name][clientName]
+	defer func() {
+		span.AddEvent("sync.finish.stats", trace.WithAttributes(
```
disq, line 84: Turns out you can also do `span.SetAttributes(...)` but having a separate "finish event" is better.

erezrokah (approved):

PR #2011, opened 2024-12-18
fix: Use field name in json type schema if json tag is missing

erezrokah:
#### Summary

I'm working on validating some data (internal issue https://github.com/cloudquery/cloudquery-issues/issues/2971) using the json type schema and noticed a bug (noticeable on AWS where structs don't have json tags).
We should not use the name transformer to get the name as it defaults to `ToSnake` when a json tag is missing.
Instead we should mimic the JSON marshaling default behavior to use the field name as is.

A better approach can be to create an instance of the type, marshal it to JSON, then use that for the names. Open to ideas how to do that using reflection, with a caveat that we might need to initialize it using non zero values otherwise those can be omitted

---

- [ ] Update or add tests 🧪


marianogappa:
I think this impacts the behaviour of the `jsonflattener` transformer plugin. It might be that it's relying on this syntax. Could you run an aws sync with the flattener and see if it stops working first?

erezrokah:
> I think this impacts the behaviour of the `jsonflattener` transformer plugin. It might be that it's relying on this syntax. Could you run an aws sync with the flattener and see if it stops working first?

I'll do that. Also added an option so plugins can override this. It's required for Azure that doesn't use `json` tags and implements custom marshaling for all models, example in https://github.com/Azure/azure-sdk-for-go/blob/d9d44c0d1d484769706a833157a34975f9623d17/sdk/resourcemanager/compute/armcompute/models_serde.go#L124

image

erezrokah:
Also opened https://github.com/Azure/azure-sdk-for-go/issues/23882

dcelasun (approved):

erezrokah:
I'll just verify the json flatter then we can release this

erezrokah:
OK so it changes the nested columns names in AWS with the JSON flattener 
image which are now a more accurate representation of the JSON structure:
image

We could make it so the `jsonflattener` always `toSnake` the flattened columns (we don't say anything about the casing of the columns in https://www.cloudquery.io/blog/introducing-the-json-flattener-transformer-integration).

Regardless don't think this should block this PR

PR #2077, opened 2025-02-11
feat: Add way to get Configtype time value for hashing

blesniewski:
#### Summary

First option was using exported fields, but even that would require additional logic- e.g. what should happen, if the user provided input is `50 minutes ago` and the sync is run twice at different times?

I think this approach makes more sense, but let me know if there are any other ideas or comments to consider.

---
closes https://github.com/cloudquery/cloudquery-private/issues/6478

### DIFF
--- a/configtype/time.go
+++ b/configtype/time.go
@@ -15,9 +15,10 @@ import (
 // when a time type is required. We wrap the time.Time type so that
 // the spec can be extended in the future to support other types of times
 type Time struct {
-	input    string
-	time     time.Time
-	duration *timeDuration
+	input       string
+	time        time.Time
+	duration    *timeDuration
+	hashNowFunc func() time.Time
 }
 
 func ParseTime(s string) (Time, error) {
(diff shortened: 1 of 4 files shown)


maaarcelino (approved):

disq (approved):

configtype/time.go
```diff
@@ -130,6 +130,11 @@ func (t Time) String() string {
 	return t.input
 }
 
+func (t Time) Hash() (uint64, error) {
+	at := t.AsTime(time.Date(2000, 1, 1, 0, 0, 0, 0, time.UTC))
+	return uint64(at.UnixNano()), nil
```
disq, line 135: How do we know if the hash belongs to a duration or a static time? We don't need to, but what if we did? Maybe we can say, `if hash < unixnano(2020-01-01)` then it's a duration (a 20-year duration wouldn't be useful to anyone)

Also, `.UnixNano`? why not just `.Unix` and take the integer part?

blesniewski: >How do we know if the hash belongs to a duration or a static time? We don't need to, but what if we did?

That's why I used the `AsTime` method which pretty conveniently does most of the thinking and results in a static time being returned for both possible inputs.
There's also a question of possibly including `t.input` into the hash - WDYT?

>.UnixNano? why not just .Unix and use seconds?

Using nano because AFAIK we can possibly support such timestamps as input (looking at the existing test cases), so why not include it in the hash?

murarustefaan (approved):

configtype/time.go
blesniewski: Now that I think about it, I'm no longer sure this is enough.
But including the input wouldn't fix what I'm worried about):
- lets say a user provides a table option saying `5 days ago` and this is the only table option provided
- then runs the sync again after a day
- hash matches with the previous incremental state key and we get the last state (which is incorrect), because it should run from scratch again.

I think having a static time set in the `Hash` function for use in `AsTime` is wrong as it makes durations match when they shouldn't.

blesniewski: Modified the function a bit, to allow simulating it being run at the exactly same time for the tests, should also fix the problems with durations as we're calculating them from the current timestamp in normal operations.

PR #2091, opened 2025-03-07
fix: Ignore context cancelled errors.

marianogappa:
fixes https://github.com/cloudquery/cloudquery-issues/issues/3451

The CLI closes the context that the Transformer plugin uses upon completion of the sync. This is not an error, so the SDK should not raise it in the `Transform` function.

This error is benign in that the sync still works as expected, but it looks bad in the logs, which end up showing up in `cloudquery.log`, and it bubbles up to other products like Asset Inventory which happens to show errors in bright red colours.

Here's a typical current run

Screenshot 2025-03-07 at 10 57 36

Here's a fixed run

Screenshot 2025-03-07 at 10 54 49

### DIFF
--- a/internal/servers/plugin/v3/plugin.go
+++ b/internal/servers/plugin/v3/plugin.go
@@ -456,6 +456,10 @@ func (s *Server) Transform(stream pb.Plugin_TransformServer) error {
 			}
 			if err != nil {
 				close(recvRecords)
+				if errors.Is(err, context.Canceled) {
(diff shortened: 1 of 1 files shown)


hermanschaaf (approved):

murarustefaan (approved):
Yeah, makes sense.

disq (approved):

erezrokah (approved):

PR #2106, opened 2025-03-31
feat: Pass installation ID from env to usage report

disq:
Goes with https://github.com/cloudquery/cloudquery/pull/20455

### DIFF
--- a/premium/usage.go
+++ b/premium/usage.go
@@ -180,8 +180,9 @@ type BatchUpdater struct {
 	awsMarketplaceClient AWSMarketplaceClientInterface
 
 	// Plugin details
-	teamName   cqapi.TeamName
-	pluginMeta plugin.Meta
+	teamName       cqapi.TeamName
+	pluginMeta     plugin.Meta
+	installationID string
 
 	// Configuration
 	batchLimit            uint32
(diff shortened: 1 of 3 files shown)


premium/usage.go
```diff
@@ -724,6 +728,15 @@ func (u *BatchUpdater) getTeamNameByTokenType(tokenType auth.TokenType) (string,
...
 }
 
+func maybeGetInstallationID(tokenType auth.TokenType) string {
+	switch tokenType {
+	case auth.SyncRunAPIKey, auth.SyncTestConnectionAPIKey:
+		return os.Getenv("_CQ_INSTALLATION_ID")
+	default:
+		return ""
+	}
+}
```
murarustefaan, line 738: Could we not have a function with a name starting with maybe? :)) 
Since this is used in a single place either way, let's skip extracting it as a separate function; or just rename it :)).

murarustefaan (commented):
One comment, ok otherwise.

premium/usage.go
disq: is `determineInstallationID()` ok? https://github.com/cloudquery/plugin-sdk/pull/2106/commits/771f100f69391e53d9d50e933c4c1066d17453c6

PR #2113, opened 2025-04-01
fix: Error if both PKs and PK components are set

erezrokah:
#### Summary

I spent too much time debugging a bug I had when I (meaning Cursor AI) auto complete `PrimaryKeyComponent: true` instead of `PrimaryKey: true` for me. Which means I had both PKs and PK components on the same table, causing `_cq_id` to only be calculated from the PK component, losing data when `pk_mode: cq-id-only` was enabled.

See https://github.com/cloudquery/cloudquery-private/pull/6902/commits/ef8762eb1f1a6624913e5ee9ff7a88e9e0944537

This PR adds a validation to ensure only either ones are defined, or not defined at all. This has to happen after the `TransformTables` call since columns can be configured manually as well

---

- [ ] Update or add tests 🧪

### DIFF
--- a/transformers/tables.go
+++ b/transformers/tables.go
@@ -26,6 +26,9 @@ func TransformTables(tables schema.Tables) error {
 		if err := TransformTables(table.Relations); err != nil {
 			return err
 		}
+		if len(table.PrimaryKeys()) > 0 && len(table.PrimaryKeyComponents()) > 0 {
+			return fmt.Errorf("primary keys and primary key components cannot both be set for table %q", table.Name)
+		}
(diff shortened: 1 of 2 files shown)


disq:
Would this be a `feat`, though?

erezrokah:
> Would this be a `feat`, though?

It's a new validation so yes I think. Could be a `fix` too if you consider it a bug

disq:
I think `fix` is better because it's preventing a bug to happen by adding a validation to check against.

przste-go (approved):

PR #2124, opened 2025-04-21
feat: Show plugin version in plugin server logs

disq:
This could make certain things easier to debug.


plugin/plugin.go
```diff
@@ -166,6 +166,10 @@ func (p *Plugin) Meta() Meta {
 	}
 }
 
+func (p *Plugin) PackageAndVersion() string {
```
disq, line 169: I didn't want to do a Stringer without any input, but not opposed to it.

erezrokah (approved):

savme (approved):

PR #2129, opened 2025-04-24
feat: Remove reflect methods

bbernays:
#### Summary

---

- [ ] Update or add tests 🧪


savme (approved):

serve/package.go
```diff
@@ -132,6 +132,9 @@ func (s *PluginServe) build(pluginDirectory string, target plugin.BuildTarget, d
 	args := []string{"build", "-o", pluginPath}
 	args = append(args, "-buildmode=exe")
 	args = append(args, "-ldflags", ldFlags)
+	// This disables gRPC tracing, which introduces a dependency that blocks dead code elimination\
```
erezrokah, line 135: This conflicts with the FIPS support I believe as it will override https://github.com/cloudquery/cloudquery-private/pull/7320/files#diff-e6ce7bc316a2e750a956a86aa73a9d0e59192339395eb4c0c1a70d5ff19a6df1R93

cc @przste-go.

Maybe we need to add an option to the SDK to pass build tags

erezrokah (commented):
Maybe we can bring back https://github.com/cloudquery/plugin-sdk/pull/1142? That should resolve https://github.com/cloudquery/cloudquery-issues/issues/2500

serve/package.go
erezrokah: Would be good to do the build flags as an option so we can gradually rollout the `grpcnotrace` change as well

bbernays:
> Maybe we can bring back #1142? That should resolve [cloudquery/cloudquery-issues#2500](https://github.com/cloudquery/cloudquery-issues/issues/2500)

I have no problem with this

serve/package.go
bbernays: [07a9db1](https://github.com/cloudquery/plugin-sdk/pull/2129/commits/07a9db138b42d8579e910d2eb474602dc7a6aef1)

serve/package.go
```diff
@@ -455,6 +462,7 @@ func (s *PluginServe) newCmdPluginPackage() *cobra.Command {
 	cmd.Flags().StringP("dist-dir", "D", "", "dist directory to output the built plugin. (default: <plugin_directory>/dist)")
 	cmd.Flags().StringP("docs-dir", "", "", "docs directory containing markdown files to copy to the dist directory. (default: <plugin_directory>/docs)")
 	cmd.Flags().StringP("message", "m", "", "message that summarizes what is new or changed in this version. Use @<file> to read from file. Supports markdown.")
+	cmd.Flags().StringP("build-tags", "", "", "build tags to use when building the plugin. (default: grpcnotrace)")
```
erezrokah, line 465: This should not be a CLI flag, it should be a field in `BuildTarget` https://github.com/cloudquery/plugin-sdk/blob/b27291d68e3dca2019a36278f704c1a1f9263cd8/plugin/plugin_package.go#L39 controlled via `WithBuildTargets` https://github.com/cloudquery/plugin-sdk/blob/b27291d68e3dca2019a36278f704c1a1f9263cd8/plugin/options.go#L20

We can change the default in https://github.com/cloudquery/plugin-sdk/blob/b27291d68e3dca2019a36278f704c1a1f9263cd8/plugin/plugin_package.go#L80 to include `grpcnotrace`

erezrokah (changes requested):

serve/package.go
bbernays: If we do it there then as soon as we use a go tag (eg `-tags=fipsEnabled`) then it will be overridden.

erezrokah: Can't the plugin set multiple tags? i.e. use the default build targets, then append to the default tags the `fipsEnabled` one?

erezrokah: Using a CLI flag is problematic since it's not backwards compatible, i.e. older plugins will fail is you pass them `--build-tags`

bbernays: Multiple tags is difficult because go requires a single `-tags` and doesn't combine between `GOFLAGS` and `-tags` it will just take the last one.

This is already a breaking change...

erezrokah: This PR also overrides the tags from the flag so won't the plugin will need to be called with `--buildTags=grpcnotrace,<additional-tags>` by whoever is invoking the package command?

I think the cleaner approach would be for the plugin to decide on the build tags, not the caller of the package command.

For FIPS plugins we would create a `plugin_fips.go` similar to https://github.com/cloudquery/cloudquery/blob/bfe4714dbac5feeffe47ca8cc7ebcf3bf0d6d008/plugins/source/test/resources/plugin/plugin_fips.go#L26

where we append the `fipsEnabled` tag when configuring the build targets, similar to other build options

bbernays: Implemented it in [c83e36a](https://github.com/cloudquery/plugin-sdk/pull/2129/commits/c83e36ab792d44f5b82b2ded0460d457b77318f6)

bbernays:
> We'll need to do a follow up in the FIPS plugins but I think that's fine. One thing I'm not sure about is classifying this as a breaking change, that would require a bit of work for our plugins to do the major version bump. The potential breaking change here is the `grpcnotrace` flag, which plugins authors can override now

I am fine making it a non-breaking change, but the reason why it is breaking change is because it removes the ability for the plugin to generate markdown docs...

erezrokah:
> but the reason why it is breaking change is because it removes the ability for the plugin to generate markdown docs...

Ah yeah that hasn't been used for a long time I believe and we were the only users

bbernays:
So just to be clear you are good with me marking this as a non breaking change?

cc @murarustefaan

erezrokah:
BTW https://github.com/cloudquery/plugin-sdk/pull/1142 removes docs generation altogether, do we need JSON docs? Can be done in another PR

murarustefaan:
> So just to be clear you are good with me marking this as a non breaking change?

Yes, I do not think the potential impact this change has warrants a major SDK version bump.

> ... do we need JSON docs? ...

Nope. But since the changeset for removing doc generation completely is bigger, we'd be better off having that in a separate PR.

murarustefaan (approved):

PR #2130, opened 2025-04-25
fix: Prevent deadlock in trasnformer

przste-go:
#### Summary

Currently sync method has 3 gorutines running in single errorGroup let's call them G1,G2 and G3
 first gorutine
 ```
 	eg.Go(func() error {
		if err := s.Plugin.Transform(gctx, recvRecords, sendRecords); err != nil {
			return status.Error(codes.Internal, err.Error())
		}
		return nil
	})
 ```
runs the transformer logic which consumes records from recvRecords and pushes them to sendRecords both of which are unbuffered channels. Gorutine g2 consumes record produced by G1
```
record := range sendRecords
(..)
```
when gorutine g2 returns an error, code executing in G1 may hang indefinietly trying to write into channel that has no associated error. 

---

- [ ] Update or add tests 🧪


murarustefaan (approved):
Good job on catching this!

erezrokah (approved):

PR #2131, opened 2025-04-28
chore: Double the limit of grpc messages

przste-go:
#### Summary

Following up on the transformer hanging issue reported by a customer and addressed in PR [#2130](https://github.com/cloudquery/plugin-sdk/pull/2130):
Currently, the transformer plugin can increase the message size beyond the limit enforced in the sync method of plugin-sdk [1](https://github.com/cloudquery/plugin-sdk/blob/bc50fd3d2be414edba8f8ad5bb7739a012840bf1/internal/servers/plugin/v3/plugin.go#L267).
To accommodate scenarios where transformers or destination plugins need to handle larger messages than the source plugin, we should consider increasing this limit to provide more flexibility.

---

- [ ] Update or add tests 🧪

### DIFF
--- a/serve/constants.go
+++ b/serve/constants.go
@@ -6,7 +6,7 @@ import (
 
 const (
 	// bufSize used for unit testing grpc server and client
-	testBufSize  = 1024 * 1024
-	flushTimeout = 5 * time.Second
-	MaxMsgSize   = 100 * 1024 * 1024 // 100 MiB
+	testBufSize    = 1024 * 1024
+	flushTimeout   = 5 * time.Second
+	MaxGrpcMsgSize = 200 * 1024 * 1024 // 100 MiB
 )
--- a/serve/constants.go
+++ b/serve/constants.go
@@ -8,5 +8,5 @@ const (
 	// bufSize used for unit testing grpc server and client
 	testBufSize    = 1024 * 1024
 	flushTimeout   = 5 * time.Second
-	MaxGrpcMsgSize = 200 * 1024 * 1024 // 100 MiB
+	MaxGrpcMsgSize = 200 * 1024 * 1024 // 200 MiB
 )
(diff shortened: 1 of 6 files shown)


murarustefaan:
I'm wondering if we need to change it here as well; probably not, since `serve` wrap around the plugin, but worth checking.
https://github.com/cloudquery/plugin-pb-go/blob/main/managedplugin/plugin.go#L39

serve/constants.go
```diff
@@ -6,7 +6,7 @@ import (
 
 const (
 	// bufSize used for unit testing grpc server and client
-	testBufSize  = 1024 * 1024
-	flushTimeout = 5 * time.Second
-	MaxMsgSize   = 100 * 1024 * 1024 // 100 MiB
+	testBufSize    = 1024 * 1024
+	flushTimeout   = 5 * time.Second
+	MaxGrpcMsgSize = 200 * 1024 * 1024 // 100 MiB
```
blesniewski, line 11:
```suggestion
	MaxGrpcMsgSize = 200 * 1024 * 1024 // 200 MiB
```
comment update

przste-go:
> I'm wondering if we need to change it here as well; probably not required, since `serve` wraps around the plugin either way. Worth checking, we could change it for consistency reasons. https://github.com/cloudquery/plugin-pb-go/blob/main/managedplugin/plugin.go#L39

@murarustefaan Not sure here, I have to find out how that code is actually used.

przste-go:
@murarustefaan I think you are right, this could be also needed due to destination client size limit in CLI.
I opened another PR
https://github.com/cloudquery/plugin-pb-go/pull/494

murarustefaan:
Nice! Then let's merge that one, get it out and upgrade the dependency version in the SDK here and only then get this one merged.

przste-go:
@murarustefaan Released and updated

murarustefaan (approved):

PR #2140, opened 2025-05-02
fix: Pass installation ID from env for usage report

disq:
Follow-up to https://github.com/cloudquery/plugin-sdk/pull/2106

Plugins use regular team API keys for usage reports, not platform-specific local API keys.

### DIFF
--- a/premium/usage.go
+++ b/premium/usage.go
@@ -285,7 +285,7 @@ func NewUsageClient(meta plugin.Meta, ops ...UsageClientOptions) (UsageClient, e
 		}
 		u.teamName = teamName
 	}
-	u.installationID = determineInstallationID(u.tokenClient.GetTokenType())
+	u.installationID = determineInstallationID()
 
 	u.backgroundUpdater()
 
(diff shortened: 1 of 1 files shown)


premium/usage.go
```diff
@@ -730,7 +730,7 @@ func (u *BatchUpdater) getTeamNameByTokenType(tokenType auth.TokenType) (string,
 
 func determineInstallationID(tokenType auth.TokenType) string {
 	switch tokenType {
-	case auth.SyncRunAPIKey, auth.SyncTestConnectionAPIKey:
+	case auth.APIKey:
 		return os.Getenv("_CQ_INSTALLATION_ID")
```
hermanschaaf, line 734: Why switch on token type at all and not always return `os.Getenv("_CQ_INSTALLATION_ID")`?

disq: We can also do that, yes

disq: Just did that in https://github.com/cloudquery/plugin-sdk/pull/2140/commits/63fb2dcfa2a18c09b891c37442c9d3a588762200

hermanschaaf (approved):

PR #2153, opened 2025-05-15
fix: Change logic for batch writing to write when batch size is reached, not exceeded

maaarcelino:
The current logic for batch writing prevents us from cleanly testing the `DeleteRecord` handling in the plugin. Currently we have to set `BatchSize: 1` and send two `DeleteRecord` messages in the test, which is not the cleanest logic. This PR changes the logic so the batch is flushed when the batch size is reached, not exceeded (so for `BatchSize: 1` it will flush after one `DeleteRecord` is received).

This also matches the logic we have in https://github.com/cloudquery/plugin-sdk/blob/main/internal/batch/cap.go#L7.

### DIFF
--- a/writers/batchwriter/batchwriter.go
+++ b/writers/batchwriter/batchwriter.go
@@ -278,7 +278,7 @@ func (w *BatchWriter) Write(ctx context.Context, msgs <-chan message.WriteMessag
 			w.deleteStaleMessages = append(w.deleteStaleMessages, m)
 			l := int64(len(w.deleteStaleMessages))
 			w.deleteStaleLock.Unlock()
-			if w.batchSize > 0 && l >= w.batchSize {
+			if w.isLimitReached(l) {
 				if err := w.flushDeleteStaleTables(ctx); err != nil {
 					return err
 				}
@@ -298,7 +298,7 @@ func (w *BatchWriter) Write(ctx context.Context, msgs <-chan message.WriteMessag
 			w.deleteRecordMessages = append(w.deleteRecordMessages, m)
 			l := int64(len(w.deleteRecordMessages))
 			w.deleteRecordLock.Unlock()
-			if w.batchSize > 0 && l >= w.batchSize {
+			if w.isLimitReached(l) {
 				if err := w.flushDeleteRecordTables(ctx); err != nil {
 					return err
 				}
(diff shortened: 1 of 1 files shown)


murarustefaan (commented):
Can't the described behavior be fixed by running tests in destinations with lower batch sizes / timeouts, rather than configuring the same delete message to be sent more times than it's needed?

maaarcelino:
@murarustefaan Yeah I guess that's a fair point. I updated the PR to change the logic for batch writers to use `>=` instead of `>` for comparisons against the batch size. This is to enable running the `WriterTestSuiteTests` test suite which sends a single `DeleteRecord` message to the plugin. We need to set the batch size to `1` in the test and to test that we have to allow for using `>=` in the logic. Does that make sense?

murarustefaan (approved):
Yeah I don't think this will impact anything, it will flush at 1 instead of 2 for `batch_size: 1`and at 5000 instead of 5001 at `batch_size: 5000`

murarustefaan:
Could you change the title and description of this PR, @maaarcelino ?

writers/batchwriter/batchwriter.go
```diff
@@ -298,7 +298,7 @@ func (w *BatchWriter) Write(ctx context.Context, msgs <-chan message.WriteMessag
 			w.deleteRecordMessages = append(w.deleteRecordMessages, m)
 			l := int64(len(w.deleteRecordMessages))
 			w.deleteRecordLock.Unlock()
-			if w.batchSize > 0 && l > w.batchSize {
+			if w.batchSize > 0 && l >= w.batchSize {
```
erezrokah, line 301:
```suggestion
			limit := batch.CappedAt(0, w.batchSize)
			limit.AddRows(l)
			if limit.ReachedLimit() {
```

I think this is more aligned with the rest of the codebase e.g. https://github.com/cloudquery/plugin-sdk/blob/c82d006af1b3fbc52f6324486df2bfd7efb927dd/writers/streamingbatchwriter/streamingbatchwriter.go#L406

same comment for other places where we check the limit

erezrokah (commented):
Agree with the change, seems like a left over from when we refactored the code to support multi row batches.

I would first try to update a few destinations (e.g. Postgres, S3 and MySQL) to this branch sha to see if this triggers any test failures, as existing tests might rely on this behavior

writers/batchwriter/batchwriter.go
maaarcelino: Done in [05b4212](https://github.com/cloudquery/plugin-sdk/pull/2153/commits/05b421261b2d926c13e61cfe9b70e97e8a70b6a5). Tested and it works correctly

PR #2162, opened 2025-05-19
fix: Ensure no table clients are ever dropped

bbernays:
#### Summary

Prior to this PR, under certain circumstances (5 clients and 3 shards) clients could be dropped because the logic was that chunks were defined as 5/3 = 1 and so each shard would have 1 client and the last client would have 2 clients. After this fix, the last shard would have 3 clients

### DIFF
--- a/scheduler/scheduler_test.go
+++ b/scheduler/scheduler_test.go
@@ -577,6 +577,51 @@ func Test_shardTableClients(t *testing.T) {
 				{client: &testExecutionClient{}, table: &schema.Table{Name: "table_5"}},
 			},
 		},
+
+		{
+			name: "uneven split 1 of 3",
+			tableClients: []tableClient{
+				{client: &testExecutionClient{}, table: &schema.Table{Name: "table_1"}},
+				{client: &testExecutionClient{}, table: &schema.Table{Name: "table_2"}},
+				{client: &testExecutionClient{}, table: &schema.Table{Name: "table_3"}},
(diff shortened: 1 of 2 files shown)


scheduler/scheduler_test.go
```diff
@@ -577,6 +577,51 @@ func Test_shardTableClients(t *testing.T) {
...
+				{client: &testExecutionClient{}, table: &schema.Table{Name: "table_1"}},
+				{client: &testExecutionClient{}, table: &schema.Table{Name: "table_2"}},
+				{client: &testExecutionClient{}, table: &schema.Table{Name: "table_3"}},
+				{client: &testExecutionClient{}, table: &schema.Table{Name: "table_4"}},
+				{client: &testExecutionClient{}, table: &schema.Table{Name: "table_5"}},
+			},
+			shard: &shard{num: 3, total: 3},
+			expected: []tableClient{
+				{client: &testExecutionClient{}, table: &schema.Table{Name: "table_3"}},
+				{client: &testExecutionClient{}, table: &schema.Table{Name: "table_4"}},
```
bbernays, line 621: Prior to this PR, only `table_3` and `table_4` would have been synced in shard 3/3 and table_5 would never have been synced

murarustefaan: I proposed an alternative in https://github.com/cloudquery/plugin-sdk/pull/2163 where the clients are distributed evenly across shards, while also preventing any from being dropped.

murarustefaan:
Closing in favor of https://github.com/cloudquery/plugin-sdk/pull/2163.

PR #2163, opened 2025-05-20
feat: Improve shard chunking

murarustefaan:
Proposed as an alternative to https://github.com/cloudquery/plugin-sdk/pull/2162/files, it simplifies the logic and ensures clients are distributed evenly across shards.


erezrokah (approved):

savme (approved):
Nice! Looks cleaner than adding extra cases

PR #2179, opened 2025-05-30
feat: Make SDK FIPS-compliant by using internal SHA1 module

maaarcelino:
The Go standard library has removed support for using SHA1 when in FIPS mode in https://github.com/golang/go/commit/54693a81fd605a9c1abbee83da072c61e38d3ebf. However the [NIST FIPS spec](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-131Ar1.pdf) allows for using SHA1 in certain cases:

> SHA-1 for non-digital signature applications:
> For all other hash function applications, the use of SHA-1 is **acceptable**. The other
> applications include HMAC, Key Derivation Functions (KDFs), Random Bit Generation, and
> hash-only applications (e.g., hashing passwords and using SHA-1 to compute a checksum,
> such as the approved integrity technique specified in Section 4.6.1 of [FIPS 140]).

The above removal from the standard lib is too restrictive, as we should be able to use this algo in our SDK for checksums. I've borrowed the implementation of packages `sha1` and `byteorder` and added them as internal modules.


schema/arrow_test.go
```diff
@@ -162,3 +163,12 @@ func TestAddInternalColumnsToRecord(t *testing.T) {
...
 	}
 }
+
+func TestCQIDHashingConsistency(t *testing.T) {
+	record := buildTestRecord("")
+	got, err := AddInternalColumnsToRecord(record, "")
+	require.NoError(t, err)
+	cqIDFields := got.Schema().FieldIndices(CqIDColumn.Name)
+	require.Len(t, cqIDFields, 1)
+	assert.Equal(t, "d8f3b1de-8c63-5a0e-a1aa-19e9b5311c24", got.Column(cqIDFields[0]).ValueStr(0))
```
erezrokah, line 173: Was `d8f3b1de-8c63-5a0e-a1aa-19e9b5311c24` calculated using the official `Sha1` package? Maybe we can have a test that creates a UUID using our internal package, then using the official one, then compares the values are the same?

maaarcelino: Yes `d8f3b1de-8c63-5a0e-a1aa-19e9b5311c24` is the value from the official package. This is the same value that is on main currently. This test does what you are describing, but without using the official package to calculate the hash. Instead it is just hardcoded as a string. Does that work, wdyt?

erezrokah: Yeah it works but hard to reason about when reading the test (and also hard to update it), maybe add a comment on how that value was created?

maaarcelino: Done

erezrokah (approved):

PR #2182, opened 2025-05-30
fix: Correctly validate backend_options configuration

murarustefaan:


state/client.go
```diff
@@ -55,7 +55,7 @@ func NewConnectedClient(ctx context.Context, backendOpts *plugin.BackendOptions)
 // The state client is guaranteed to be non-nil (it defaults to the NoOpClient).
 // You must call Close() on the returned Client object.
 func NewConnectedClientWithOptions(ctx context.Context, backendOpts *plugin.BackendOptions, connOpts ConnectionOptions, clOpts ClientOptions) (Client, error) {
-	if backendOpts == nil {
+	if backendOpts == nil || backendOpts.Connection == "" || backendOpts.TableName == "" {
```
bbernays, line 58: Maybe if only one of those options is blank we should log a warning?

bbernays: If both are empty just return `NoOP` if only one is empty return an error? Because at that point they tried to configure something but messed up

murarustefaan: Yeah, I think that makes sense.

murarustefaan: Done in [`5ec0193` (#2182)](https://github.com/cloudquery/plugin-sdk/pull/2182/commits/5ec0193ec12b1b872de7c5a226d98c0dacc5f9f5)

murarustefaan: Considered moving this logic on a BackendOptions received, but only complicates it's uses.

PR #2197, opened 2025-06-27
feat: Make more data dynamic

bbernays:
#### Summary

This PR makes more fields dynamic, based on the random number seed. The data stays deterministic, but enables each row to have a different value


erezrokah (commented):
I would put this behind an option if possible, since it's likely we have existing tests that will break after this change.

Or maybe release it and see what happens when we update to the SDK 🙈

murarustefaan (approved):

PR #2517, opened 2026-06-11
fix: Don't exit early if a single PreResource Resolver fails

bbernays:
#### Summary

---

- [ ] Update or add tests 🧪

### DIFF
--- a/scheduler/resolvers/resolvers.go
+++ b/scheduler/resolvers/resolvers.go
@@ -9,9 +9,7 @@ import (
 	"github.com/cloudquery/plugin-sdk/v4/caser"
 	"github.com/cloudquery/plugin-sdk/v4/scheduler/metrics"
 	"github.com/cloudquery/plugin-sdk/v4/schema"
-	"github.com/getsentry/sentry-go"
 	"github.com/rs/zerolog"
-	"github.com/thoas/go-funk"
 )
 
 func resolveColumn(ctx context.Context, logger zerolog.Logger, m *metrics.Metrics, selector metrics.Selector, client schema.ClientMeta, resource *schema.Resource, column schema.Column, c *caser.Caser) {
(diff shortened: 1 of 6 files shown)


murarustefaan:
I think this is a good catch. 

It aligns PreResourceResolver with the rest of the resolver pipeline, column resolvers and PostResourceResolver already log-count-and-continue per resource. PreResourceResolver was the only per-resource hook with chunk-level failure semantics.

scheduler/resolvers/resolvers.go
```diff
@@ -79,13 +79,16 @@ func ResolveResourcesChunk(ctx context.Context, logger zerolog.Logger, m *metric
 	}
 
 	if table.PreResourceResolver != nil {
+		filtered := make([]*schema.Resource, 0, len(resources))
 		for _, resource := range resources {
 			if err := table.PreResourceResolver(ctx, client, resource); err != nil {
 				tableLogger.Error().Err(err).Msg("pre resource resolver failed")
```
murarustefaan, line 90:
```suggestion
				if ctx.Err() != nil {
				    return nil // context cancelled, remaining resolvers would all fail the same way
				}
				tableLogger.Error().Err(err).Msg("pre resource resolver failed")
```

Maybe add a guard for context cancellation here as well.

Before, a context cancel would stop the following resolver calls, but with continue we'd have a row of errors here since all would go trough the err flow.

erezrokah (approved):

erezrokah (commented):
Can we somehow test this on a real full sync? Maybe extend the test plugin https://github.com/cloudquery/cloudquery/tree/main/plugins/source/test/resources/services to have a PreResourceResolver that fails only for some of the resources? It would be good to test this with
1. Parent -> Child 1 when some Parent resources fail to resolve
2. Parent -> Child 1 -> Child 2 when some Child 1 resources fail to resolve

scheduler/scheduler_preresource_partial_failure_test.go
```diff
@@ -0,0 +1,120 @@
...
+				collect(child1),
+				"children of surviving parents should sync, except the failing child; no children of the dropped parent")
+
+			require.ElementsMatch(t,
+				[]string{"p0/c0/g0", "p0/c1/g0", "p1/c1/g0", "p3/c0/g0", "p3/c1/g0", "p4/c0/g0", "p4/c1/g0"},
+				collect(child2),
+				"grandchildren should only sync under surviving child1 rows")
+		})
+	}
+}
```
murarustefaan, line 120: @erezrokah this tests the 2 cases you mentioned. Managed to also find a race condition when passing work items from the scheduler to workers.

scheduler/resolvers/resolvers.go
```diff
@@ -79,13 +79,16 @@ func ResolveResourcesChunk(ctx context.Context, logger zerolog.Logger, m *metric
...
 				m.AddErrors(ctx, 1, selector)
```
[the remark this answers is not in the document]
murarustefaan: Logging the index would not help at all.

erezrokah (approved):

PR #2554, opened 2026-08-03
feat(testdata): Let destinations bound generated 64-bit integers

murarustefaan:
## Current

arrow-go v18.7.0 fixed large-integer precision in JSON decoding ([apache/arrow-go#816](https://github.com/apache/arrow-go/pull/816)), and `TestDataGenerator` builds records via `builder.UnmarshalJSON` — so until now every int64/uint64 it produced was silently rounded to a float64-representable value. They are now exact, which turned 16 destination plugins red because their storage or wire format round-trips integers through float64.

## Expected

Destinations declare the integer width they can represent exactly via `MaxIntegerBits` (`schema.Float64SafeIntegerBits` for the float64 case). Unset keeps the full 64-bit range, so consumers that do not opt in are unaffected, and `TableDiff` now names the option when a differing column matches after a float64 round-trip.

## Why / Impact

Test-only API, additive, no behaviour change without opt-in; `timestamp_ns` already handles this same root cause unconditionally, and keeping this opt-in preserves real 64-bit coverage for destinations that are correct. Verified end-to-end: the `file` destination's csv failure at arrow-go v18.7.0 passes with `MaxIntegerBits: 53`, and removing any plumbing line now fails `TestPluginFloat64Ints`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/memdb/memdb.go
+++ b/internal/memdb/memdb.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/apache/arrow-go/v18/arrow"
 	"github.com/apache/arrow-go/v18/arrow/array"
+	"github.com/apache/arrow-go/v18/arrow/memory"
 	"github.com/cloudquery/plugin-sdk/v4/message"
 	"github.com/cloudquery/plugin-sdk/v4/plugin"
 	"github.com/cloudquery/plugin-sdk/v4/schema"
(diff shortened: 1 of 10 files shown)


erezrokah (commented):
There's some overlap with https://github.com/cloudquery/plugin-sdk/pull/2553 I think, @disq fixed this already in https://github.com/cloudquery/filetypes/pull/757/changes/b25d89f9d3c67bcbe0ad76fef0083e64209d88a0

PR #2593, opened 2026-09-09
feat: Removed Column Index lookup

bbernays:
#### Summary

Simplified implementation of https://github.com/cloudquery/plugin-sdk/pull/2592 that handles the main call site. In the future we can look at implementing a more fully implemented cache


erezrokah:
My suggestion in #2592 was wrong. `TransformWithStruct` defaults every column to `schema.PathResolver`, which calls `r.Set(c.Name, ...)`, so the `Resolver == nil` branch here is almost never taken and the linear lookup stays.

To cover it, make `Set` itself fast: store the current column index on the resource before calling the resolver, and have `Set` check it first.

```go
// schema/resource.go
type Resource struct {
    // ...
    columnIndexHint int
}

func (r *Resource) Set(columnName string, value any) error {
    index := r.columnIndexHint
    if index < 0 || index >= len(r.data) || r.Table.Columns[index].Name != columnName {
        index = r.Table.Columns.Index(columnName)
        if index == -1 {
            panic(columnName + " column not found")
        }
    }
    // existing r.data[index].Set(value) + panic on error
}

func (r *Resource) setColumnIndexHint(i int) { r.columnIndexHint = i }
```

`resolveColumn` calls `resource.setColumnIndexHint(colIndex)` before `column.Resolver(...)` or `Set`. Since `scheduler/resolvers` is a separate package, the setter must be exported (or the loop moves into `schema`). This removes the need for `SetWithIndex`.

PR #2595, opened 2026-09-22
fix: Split oversized sync messages instead of dropping them

erezrokah:
Oversized sync messages are now split into several messages instead of being silently dropped, a single row that still exceeds the limit fails the sync with the table name, and the scheduler batcher gained a byte cap so oversized batches are rarely built in the first place.

Fixes https://linear.app/env-zero/issue/ATH-807/sdk-split-oversized-sync-messages-instead-of-dropping-them


cloudquerydaniel:
## Summary

Oversized gRPC sync messages used to be logged and silently dropped, losing rows. This PR splits an oversized insert into several smaller messages, fails the sync with the table name when a single row still does not fit, and adds a byte cap to the scheduler's batcher so oversized batches are built less often.

**Key changes:**
- `internal/batch.SplitRecord` — a new helper that turns one record into a list of records under a row or byte cap.
- `internal/batch/slice.go` — `getToFlush` now clears the record after the one-row-per-slice path, and `bytesPerRow` / `Bytes` are clamped so a slice's byte count cannot go negative.
- `plugin/v3.Server.sendInsert` — splits and retries until every message fits, or errors naming the table.
- `scheduler` — new `MaxSizeBytes` batch setting, defaulted to 50 MiB.

## Diagram

```mermaid
%%{init: {'theme':'neutral','flowchart':{'nodeSpacing':16,'rankSpacing':26,'padding':3},'themeVariables':{'fontSize':'11px'}}}%%
flowchart LR
  subgraph New["New (this PR)"]
    B[scheduler worker.send] -->|"batch > MaxSizeBytes"| SP[batch.SplitRecord]
    SI[Server.sendInsert] -->|"proto.Size > MaxMsgSize"| MR[maxRowsPerMessage]
    MR -->|"row cap"| SP
    SI -->|"1 row still too big"| ER[["error: fails the sync"]]
  end
  subgraph Reused["Reused / shared"]
    SR[batch.SliceRecord]:::reuse
    CAP[batch.CappedAt]:::reuse
  end
  SP --> SR
  SP --> CAP
  B -->|"SyncInsert msgs"| SI
  SI -->|"stream.Send"| CLI[(CLI / destination)]
  classDef reuse stroke:#3fb950,stroke-width:2px;
```

<sub>Click the diagram to zoom / pan.</sub>

## Review

### 🟡 MERGE WITH CARE

The core splitting logic is correct and the arithmetic holds up: the size comparison is `size <= maxMsgSize` (send) / `> maxMsgSize` (split), which matches grpc-go's own `len(payload) > maxSendMessageSize` check, so there is no off-by-one; `maxRowsPerMessage` targets 90% of the limit and provably returns a value strictly below the current row count, so the retry loop always makes progress and terminates; `sendInsert` prepends split chunks to the front of `pending`, so row order on the stream is preserved, and `stream.Send` still supplies flow-control backpressure per message. The single remaining concern is that the scheduler's new byte cap does not bound the *first* batch a worker builds.

The behaviour change is worth calling out in the release notes: a single row that exceeds 100 MiB now fails the whole sync (`plugin.go:325`) where it used to be dropped, and an oversized non-insert message now does the same (`plugin.go:282`). That is the point of the fix, but it turns a silent data-loss path into a hard failure for anyone relying on it today.

**Testing:** well covered for a change of this shape — unit tests for `SplitRecord` (including the no-single-row-fits and empty-record edges), for `sendInsert` (split, single-message, single-row-too-large), for the batcher's byte cap, and a regression assertion on row counts in `batchwriter_test.go` that pins the `getToFlush` duplication fix. The one gap is the changed non-insert error path at `plugin.go:282`.

### Potential Issues

<details>
<summary>🟡 <b>Medium</b> · <code>real</code> · <code>correctness</code> · <code>scheduler/batch.go:165</code> — the byte cap never bounds the first batch, so with <code>MaxRows</code> disabled the worker buffers without limit</summary>

`w.bytesPerRow` is only assigned inside `send()` (`scheduler/batch.go:149`), so it is `0` until the worker has flushed at least once. `reachedSizeLimit()` guards on `w.bytesPerRow > 0`, so the size check at `scheduler/batch.go:187` is inert for the first batch. With the default `MaxRows = 50` the damage is capped at 50 rows, but `WithBatchMaxRows(0)` disables the row cap entirely, and then nothing bounds the batch until the timeout fires.

Verified with a probe against this branch: `BatchSettings{MaxRows: 0, MaxSizeBytes: 4096}` fed 2000 rows of 1 KiB buffered all 2000 rows (~2 MB) in the builder before any flush, then emitted 2000 messages from the single `send()`. No rows are lost — `SplitRecord` in `send()` still keeps every emitted message under the cap — so this is a memory-footprint issue, not a correctness one, but it means the PR's stated goal (oversized batches are rarely built in the first place) is not met on that path.

**Suggested fix:**

```go
// seed bytesPerRow from the schema so the very first batch is bounded too,
// or refuse to leave both caps unbounded:
func (w *worker) reachedSizeLimit() bool {
	if w.maxSizeBytes <= 0 {
		return false
	}
	if w.bytesPerRow <= 0 {
		// no measurement yet: fall back to measuring the builder periodically
		return w.maxRows <= 0 && int64(w.curRows) >= defaultUnmeasuredRowCap
	}
	return int64(w.curRows)*w.bytesPerRow >= w.maxSizeBytes
}
```

</details>

<details>
<summary>🔵 <b>Low</b> · <code>real</code> · <code>tests</code> · <code>internal/servers/plugin/v3/plugin.go:282</code> — the non-insert oversized path changed from drop-and-continue to failing the sync, with no test</summary>

The post-switch check used to log and `continue`; it now returns `codes.Internal`. That covers `SyncMigrateTable`, `SyncDeleteRecord` and `SyncError`. `plugin_msgsize_test.go` exercises only the insert path via `sendInsert`, so nothing pins the new behaviour of this branch — a future refactor could quietly restore the drop. The repo publishes no testing contract, so this is judged from the change: an error path on a non-Tier-1 message type, unlikely to trigger in practice (a 100 MiB table schema), which is why it is 🔵 rather than 🟠.

**Suggested fix:**

```go
// internal/servers/plugin/v3/plugin_msgsize_test.go
// drive Server.Sync with a plugin emitting an oversized *message.SyncMigrateTable
// against a Server{MaxMsgSize: <small>} and assert the returned status is Internal
// and mentions the message type.
```

</details>

### Code quality

The change is tight and reuses `batch.SliceRecord` / `batch.CappedAt` rather than writing a second splitter, which is the right call. Two small items, both cosmetic.

- **One unreachable guard** in `maxRowsPerMessage`.
- **A third test file** in `internal/batch`, where the two existing ones already cover the same type.

<details>
<summary>🔵 <b>Low</b> · <code>dead-code</code> · <code>internal/servers/plugin/v3/plugin.go:343</code> — the <code>perMessage >= rows</code> guard cannot fire · <b>saves 3 lines</b></summary>

`maxRowsPerMessage` is called from exactly one place, `plugin.go:328`, and only after `size > maxMsgSize` has been established at `plugin.go:322`. With `size > maxMsgSize`, `rows * maxMsgSize * 9 / (size * 10) < rows * 9 / 10 < rows`, so `perMessage >= rows` is never true. Confirmed with a probe over `rows ∈ {2, 3, 7, 50, 1000, 1<<20}` and `size ∈ {max+1, max+7, 2·max, 100·max}` — no hit. `TestMaxRowsPerMessage`'s `require.Less(t, got, tc.rows)` passes without the branch. The `max(perMessage, 1)` floor is the guard that actually does the work.

**Suggested fix:**

```go
func maxRowsPerMessage(size int, rows int64, maxMsgSize int) int64 {
	// callers only reach here when size > maxMsgSize, so this is always < rows
	return max(rows*int64(maxMsgSize)*9/(int64(size)*10), 1)
}
```

</details>

<details>
<summary>🔵 <b>Low</b> · <code>duplication</code> · <code>internal/batch/negative_bytes_test.go:1</code> — a third test file for one test, with a helper the sibling file already has a form of · <b>saves 1 file, ~12 lines</b></summary>

`internal/batch` already has `slice_test.go`, and this PR adds both `split_test.go` and `negative_bytes_test.go`. `boolRecord` (`negative_bytes_test.go:14`) does the same job as `testRecord` (`split_test.go:16`) with a different schema, and `TestSlicedRecordBytesNeverNegative` is a case of the same behaviour `split_test.go` covers. Folding it in as a subtest of `TestSplitRecord` (or a second `func` in `split_test.go`) drops a file and the duplicate imports without changing what is asserted.

</details>

<details>
<summary>🔵 <b>Low</b> · <code>readability</code> · <code>scheduler/batch.go:22</code> — the 50 MiB batch cap and the 100 MiB message cap are silently coupled · <b>no line change — stops the two constants drifting</b></summary>

`DefaultBatchMaxSizeBytes = 50 * 1024 * 1024` is half of `MaxMsgSize = 100 * 1024 * 1024` (`internal/servers/plugin/v3/plugin.go:25`), which reads deliberate — it leaves headroom so the scheduler's batches normally fit a single gRPC message. Nothing says so, and the two constants sit in different packages, so raising one without the other would quietly push every batch back through the splitting path. One line naming the relationship is enough; `MaxMsgSize` sits in an `internal/` package so it cannot be referenced directly.

**Suggested fix:**

```go
	// half of internal/servers/plugin/v3.MaxMsgSize, so a full batch normally
	// fits into one gRPC message without splitting
	DefaultBatchMaxSizeBytes = 50 * 1024 * 1024
```

</details>

<details>
<summary>Files reviewed (11)</summary>

| File | Outcome |
|---|---|
| `internal/servers/plugin/v3/plugin.go` | 🔵 untested non-insert error path · 🔵 quality: unreachable guard in `maxRowsPerMessage` |
| `internal/batch/slice.go` | reviewed, nothing found — the `getToFlush` nil-out fixes a real pre-existing duplication (`rest` used to return the whole record alongside the per-row slices), and the `max(…, 0)` / `max(…, 1)` clamps hold for records with more rows than bytes |
| `scheduler/batch.go` | 🟡 byte cap inert on the first batch · 🔵 quality: 50 MiB / 100 MiB coupling undocumented |
| `scheduler/scheduler.go` | reviewed, nothing found — defaults `MaxSizeBytes`; `BatchSettings` built elsewhere still gets `0` (disabled), which is consistent |
| `internal/batch/split_test.go` | reviewed, nothing found |
| `internal/batch/negative_bytes_test.go` | 🔵 quality: fold into `split_test.go` |
| `internal/servers/plugin/v3/plugin_msgsize_test.go` | reviewed, nothing found |
| `scheduler/batch_size_test.go` | reviewed, nothing found |
| `writers/batchwriter/batchwriter_test.go` | reviewed, nothing found — the `InsertedRows` assertion is the regression test for the `getToFlush` fix |
| `examples/simple_plugin/go.mod` | reviewed, nothing found — unrelated CI fix bundled in; adds `github.com/stoewer/go-strcase v1.3.1` as indirect, already in the root module |
| `examples/simple_plugin/go.sum` | skipped — follows its manifest |

</details>

<sub>Stack: Go (go.mod, arrow-go, grpc) · kinds reviewed: application code, tests, dependency manifest · repo guidance: none found (no AGENTS.md/CLAUDE.md, no .claude/skills; CONTRIBUTING.md read — it covers build/lint only)</sub>
<sub><sub>Last reviewed commit: [9872505](https://github.com/cloudquery/plugin-sdk/commit/98725054ef0abc83ca221c6c1c5defe997fa372a) · reviewed 2026-09-22 16:56 UTC</sub></sub>

cloudquerydaniel (approved):

